PR #140, opened 2019-03-19
Add release, CI status, and godoc badges

MikaelSmith:


MikaelSmith:
Do we want to tag a release so the releases badge works?

PR #141, opened 2019-03-19
Add deployment stage to build and deploy wash releases

MikaelSmith:
When puppetlabs/wash is tagged, will build `wash` binaries for macOS and Linux and include them with the release.


MikaelSmith:
Didn't work right.

PR #142, opened 2019-03-19
Centralize the cache ID in CacheConfig

ekinanp:
This way, CachedList, CachedOpen, and CachedMetadata do not require the
caller to pass-in a cache ID. Thus, plugin authors are free to use them.
CachedMetadata is especially useful for a container/ec2 instance's metadata.json
file.

This commit also paves the way for creating a more general CachedOp
method that plugin authors can use if they need more fine-grained
caching.


plugin/cache.go
```diff
@@ -134,7 +149,11 @@ func ClearCacheFor(path string) ([]string, error) {
 	return cache.Delete(rx), nil
 }
 
-func cachedOpHelper(op cachedOp, entry Entry, id string, generateValue func() (interface{}, error)) (interface{}, error) {
+func panicNilCacheConfig() {
```
ekinanp: I wonder if we can just embed `CacheConfig` in the `Entry` interface (i.e. have its methods be part of the `Entry` interface). We can force plugin authors to use EntryBase everywhere (and use `plugin.NewEntry` to create these objects). Embedding `CacheConfig` in `Entry` would avoid the nil cache config case, and it makes it easy to reason about stuff (all entries are cacheable and have cache IDs, but you can choose to turn off caching if you don't need the behavior)

ekinanp:
Exposing `CachedOp` would also solve https://github.com/puppetlabs/wash/blob/master/volume/dir.go#L13 -- you'd be able to create a `CachedDirMap` method and pass-around the main volume object across `VolumeDir` and `VolumeFile`

plugin/cache.go
ekinanp: Actually, we could just have `GetTTLOf`, and `ID` added to Entry while the setters could be restricted to EntryBase.

I don't see a good reason why plugin authors shouldn't use EntryBase all the time.

plugin/types.go
```diff
@@ -125,8 +120,43 @@ type Attributes struct {
 // SizeUnknown can be used to denote that the size is unknown and should be queried from content.
 const SizeUnknown = ^uint64(0)
 
-// The Registry contains the core filesystem data.
-// Plugins: maps plugin mount points to their implementations.
+// Registry represents the plugin registry. It is also Wash's root.
```
MikaelSmith: I'd put this n a separate file now. It's something plugin writers don't need to know, and I've been trying to keep types.go constrained to things you use when writing plugins.

plugin/cache.go
MikaelSmith: That seems ok.

ekinanp: I'll do that in this PR as a separate commit

PR #145, opened 2019-03-19
Embed CacheConfig in the Entry interface

ekinanp:
This commit removes CacheConfig by embedding its methods in the Entry
interface. Doing so makes the code easier to read and maintain because
it avoids passing around a CacheConfig pointer, and it structurally
captures the fact that all entries have cacheable operations.

Additionally, this commit makes plugin.Entry a sealed interface,
which forces internal plugin authors to use plugin.NewEntry when
creating their objects.

### DIFF
--- a/plugin/entryBase.go
+++ b/plugin/entryBase.go
@@ -1,0 +1,87 @@
+package plugin
+
+import (
+	"flag"
+	"time"
+)
+
+type cacheableOp int8
+
+const (
+	// List represents Group#List
+	List cacheableOp = iota
+	// Open represents Readable#Open
+	Open
+	// Metadata represents Resource#Metadata
+	Metadata
+)
+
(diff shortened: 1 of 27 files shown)


plugin/entryBase.go
```diff
@@ -0,0 +1,87 @@
...
+// You should use plugin.NewEntry to create new EntryBase objects.
+type EntryBase struct {
+	name string
+	// id represents the entry's wash ID. It is set in CachedList.
+	id  string
+	ttl [3]time.Duration
+}
+
+// NewEntry creates a new entry
+func NewEntry(name string) EntryBase {
```
ekinanp: I could move `NewEntry` back to `types.go` and put the tests in a `types_test.go` file if that's preferred.

MikaelSmith: This seems ok. The documentation will all be rolled up, and it's enough extra stuff that it would obfuscate the other things in types.go.

PR #147, opened 2019-03-19
Add version command

### LINKED ISSUE #144: Add a `wash version` command
Add a `wash version` command that people can use to check if their installation succeeded and if they have the right version of Wash.

Set version in builds according to https://www.reddit.com/r/golang/comments/4cpi2y/question_where_to_keep_the_version_number_of_a_go/d1kbap7.

MikaelSmith:
Lists the current version. Version is set with `go build -ldflags="-X
github.com/puppetlabs/wash/cmd.version=${VERSION}"` as part of tagged
builds.

Resolves #144.

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -1,0 +1,27 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+// Version is set with `go build -ldflags="-X github.com/puppetlabs/wash/cmd.version=${VERSION}"`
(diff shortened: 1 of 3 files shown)


cmd/version.go
```diff
@@ -0,0 +1,25 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var version = "unknown"
```
ekinanp: Commenting how this works in the code would be more useful than leaving it in the commit message.

MikaelSmith: Added.

PR #149, opened 2019-03-20
Better docs

### LINKED ISSUE #148: Useful godoc output
Provide more useful godoc output.

MikaelSmith:
Improves the landing page for external plugins docs.

Adds package-level documentation to provide useful context for starting to improve documentation in individual packages.

Resolves #148.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,3 +1,4 @@
+// Package cmd implements Wash's CLI using https://github.com/spf13/cobra.
 package cmd
 
 import (
--- a/journal/core.go
+++ b/journal/core.go
@@ -1,3 +1,11 @@
+// Package journal provides tools for recording wash operations to journals stored
+// in the user's cache directory. The cache directory is created at 'wash/activity'
+// in the directory found via https://golang.org/pkg/os/#UserCacheDir. Journals are
+// separated by Journal ID.
+//
+// Wash plugins should use
+//	journal.Record(ctx context.Context, msg string, a ...interface{})
+// to record entries. The context contains the Journal ID.
 package journal
 
 import (
--- a/plugin/types.go
+++ b/plugin/types.go
@@ -1,4 +1,31 @@
+/*
+Package plugin defines a set of interfaces that plugins must implement to enable wash
+functonality.
+
+All resources must implement the Entry interface. To do so they should include the EntryBase
+type, and initialize it via NewEntry. For example
+	type myResource struct {
+		plugin.EntryBase
+	}
+	...
+	rsc := myResource{plugin.NewEntry("a resource")}
+EntryBase gives the resource a name - which is how it will be displayed in the filesystem
+or referenced via the API - and tools for controlling how its data is cached.
+
+The Group interface identifies the resource as a container for other things. Implementing it
+enables displaying it as a directory in the filesystem. Anything that does not implement
+Group will be displayed as a file.
+
+The File interface allows control over its filesystem attributes.
+
+The Readable interface gives a file its contents when read via the filesystem.
+
+All of the above, as well as other types - Resource, Execable, Pipe - provide
+additional functionality via the HTTP API.
+*/
 package plugin
+
+// This file should be reserved for types that plugin authors need to understand.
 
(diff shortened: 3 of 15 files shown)


cmd/root.go
```diff
@@ -1,3 +1,4 @@
+// Package cmd implements a CLI using https://github.com/spf13/cobra.
```
ekinanp: Nit: `implements a CLI` => `implements Wash's CLI`

journal/core.go
```diff
@@ -1,3 +1,11 @@
+// Package journal provides tools for recording wash operations to journals stored
+// in the user's cache directory. The cache directory is created at 'wash/activity'
+// in the directory found via https://golang.org/pkg/os/#UserCacheDir. Journals are
+// separated by Journal ID.
+//
+// Wash plugins should use
+//	journal.Record(ctx context.Context, msg string, a ...interface{})
+// to record entries. In wash, the context will be configured with a particular Journal ID.
```
ekinanp: `In wash, the context ...` => `The context contains the journal ID.`

plugin/action.go
```diff
@@ -44,10 +44,8 @@ var ExecAction = newAction("exec", "Execable")
 // entry's supported actions.
 func SupportedActionsOf(entry Entry) []string {
 	switch t := entry.(type) {
-	case *ExternalPluginRoot:
-		return t.supportedActions
-	case *ExternalPluginEntry:
-		return t.supportedActions
+	case DynamicEntry:
```
ekinanp: How about `ExternalPluginEntry` ? `DynamicEntry`'s a bit weird.

plugin/entryBase.go
```diff
@@ -72,6 +72,16 @@ func (e *EntryBase) setID(id string) {
 	e.id = id
 }
 
+// GetTestTTLOf gets the TTL for an operation for testing.
```
ekinanp: Nit: `for an operation` => `of an operation`

plugin/types.go
```diff
@@ -1,14 +1,39 @@
...
+
+All resources must implement the Entry interface. To do so they should include the EntryBase
+type, and initialize it via NewEntry. For example
+	type myResource struct {
+		plugin.EntryBase
+	}
+	...
+	rsc := myResource{plugin.NewEntry("a resource")}
+EntryBase gives the resource a name - which is how it will be displayed in the filesystem
+or referenced via the API - and tools for controlling how its data are cached.
```
ekinanp: `are cached` => `is cached`

plugin/external/pluginRoot.go
```diff
@@ -0,0 +1,75 @@
+// Package external defines types to represent the structure of plugin implemented
```
ekinanp: Now that I think about it, I kind of like the old structure better, where external plugin stuff was in the `plugin` package (even though I accidentally ended up using that structure due to a dependency cycle that this PR solves). The reason is b/c an external plugin's implementation is internal to Wash and is not something that people should touch. In fact, the only relevant bit that should be exported (and documented) is `ExternalPluginSpec`.

Right now, `NewExternalPluginRoot` needs to be exported, but I think we can clean that up by adding an extra method to the plugin registry (`RegisterExternalPlugin(ExternalPluginSpec)`). That can be a separate PR though.

plugin/types.go
ekinanp: It might be better if this was in a `docs/core_plugins/` section so we can add more detail and maybe even a walkthrough.

MikaelSmith: I thought it likely people would end up at https://godoc.org/github.com/puppetlabs/wash/plugin while implementing a plugin, and wanted to have some sort of roadmap. We'll also want something more complete walking through how to create a plugin.

plugin/external/pluginRoot.go
MikaelSmith: I'll explore making more of them private types.

MikaelSmith: Removed.

MikaelSmith:
I've addressed the comments.

PR #151, opened 2019-03-20
Make external plugin implementation private

MikaelSmith:
The external plugin implementation is an implementation detail in wash,
it should not be exported from the package. Make most of the
implementation private. Leaves RegisterExternalPlugin (and
ExternalPluginSpec) public to allow the CLI to augment the registry.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -171,11 +171,9 @@ func loadLogger(levelStr string, logfile string) (*os.File, error) {
 
 func loadPlugin(registry *plugin.Registry, name string, root plugin.Root) {
 	log.Infof("Loading %v", name)
-	if err := root.Init(); err != nil {
+	if err := registry.RegisterPlugin(root); err != nil {
 		// %+v is a convention used by some errors to print additional context such as a stack trace
 		log.Warnf("%v failed to load: %+v", name, err)
-	} else {
-		registry.RegisterPlugin(name, root)
 	}
 }
 
@@ -213,8 +211,11 @@ func loadExternalPlugins(registry *plugin.Registry, externalPluginsPath string) 
 	}
 
 	for _, p := range externalPlugins {
-		root := plugin.NewExternalPluginRoot(p)
-		loadPlugin(registry, p.Name, root)
+		log.Infof("Loading %v", p.Script)
+		if err := registry.RegisterExternalPlugin(p); err != nil {
+			// %+v is a convention used by some errors to print additional context such as a stack trace
+			log.Warnf("%v failed to load: %+v", p.Script, err)
+		}
 	}
 
 	log.Infof("Finished loading external plugins")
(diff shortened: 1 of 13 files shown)


plugin/registry.go
```diff
@@ -34,10 +34,29 @@ func (r *Registry) Plugins() map[string]Root {
...
+
+// ExternalPluginSpec represents an external plugin's specification.
+type ExternalPluginSpec struct {
+	Name   string
+	Script string
+}
+
+// RegisterExternalPlugin initializes an external plugin from its spec and
+// passes it to RegisterPlugin.
+func (r *Registry) RegisterExternalPlugin(spec ExternalPluginSpec) error {
```
ekinanp: Would be useful to check that `spec.Script` exists

MikaelSmith: I did that and added some testing.

cmd/server.go
```diff
@@ -213,8 +211,11 @@ func loadExternalPlugins(registry *plugin.Registry, externalPluginsPath string)
 	}
 
 	for _, p := range externalPlugins {
-		root := plugin.NewExternalPluginRoot(p)
-		loadPlugin(registry, p.Name, root)
```
MikaelSmith: Is it worth keeping Name in the YAML config just for this?

docs/external_plugins/introduction.md
```diff
@@ -8,14 +8,11 @@ External plugins let Wash talk to other things outside of the built-in plugins.
...
-3. Add the plugin to the (configurable) `plugins.yaml` file by specifying its name and path to the plugin script. An example `plugins.yaml` file is shown below:
+3. Add the plugin to the (configurable) `plugins.yaml` file by specifying a path to the plugin script. The name will be determined by invoking the script with the `init` action. An example `plugins.yaml` file is shown below:
 
     ```
-    - name: 'external-aws'
-      script: '/Users/enis.inan/wash/external-plugins/external-aws.rb'
-  
-    - name: 'network'
-      script: '/Users/enis.inan/wash/external-plugins/network.sh'
+    - script: '/Users/enis.inan/wash/external-plugins/external-aws.rb'
```
MikaelSmith: Should we keep Name just for logging output? Should we keep this an object with a single field, or accept a list of scripts?

plugin/registry_test.go
```diff
@@ -0,0 +1,90 @@
...
+	args := m.Called()
+	return args.Error(0)
+}
+
+func (m *mockRoot) List(ctx context.Context) ([]Entry, error) {
+	args := m.Called(ctx)
+	return args.Get(0).([]Entry), args.Error(1)
+}
+
+func (suite *RegistryTestSuite) TestRegisterPlugin() {
```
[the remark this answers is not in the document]
MikaelSmith: I find it easier to reason about tests when they're small. So I'd prefer to keep them separate when possible.

Issue #156, opened 2019-03-20
Allow plugins to specify / in their entry names


ekinanp:
Currently, entries cannot have `/`es in their names, since `/` is the path separator. This is restricting since some APIs could return entry names with `/`es in them or have a hierarchical way of representing paths like `//foo`. For example in S3, `//foo` is modeled as `/:/:foo` (where `:` is the path separator).

Both the API and FUSE do not allow paths like `//foo`. However, we can fix the API to recognize these paths. For FUSE, we can indicate in a `KNOWN ISSUES` section that paths like `//foo` won't work, and that people should use the Wash commands `ls`, `find` when writing their scripts.

MikaelSmith:
We can live without FUSE, but I'd like to understand more how API calls would refer to a path, and how that would affect plugin development.

ekinanp:
Yeah, I'll work on exploring this next once the caching stuff's done + the core plugin docs.

ekinanp:
@MikaelSmith I'm starting to agree with your earlier suggestion on addressing this if it becomes annoying for plugin authors. However I do think we should remove storing the API key in xattr and just compute it ourselves. This way, we don't have to depend on FUSE to retrieve that information.

MikaelSmith:
Not relying on FUSE for the API key is probably a good move. Some optimizations will be hard if we first have to query the filesystem.

ekinanp:
I think there's a pretty easy way to solve this. If we require plugin names to conform to some regex, e.g. like `[A-Za-z_]+`, then in [findEntry](https://github.com/puppetlabs/wash/blob/master/api/helpers.go#L11), if we strip the `/<plugin_name>/` part of the path (which we can do since our Regex guarantees that the plugin name won't contain a `/`), then notice that our path will be of the form (using psuedo-regex notation):
```
<entry_name>(<sep><entry_name>)*
```

where `<sep>` is some arbitrary path separator. Using this, if we change the matching algorithm to something like
```
curGroup = pluginRoot
curEntry = pluginRoot
while path is not empty
  if curGroup is not a Group
    error
  end

  entries = curGroup.List
  for entry in entries do
    if entry.Name().IsPrefixOf(path)
      curGroup = entry
      curEntry = entry
      path = path.StripPrefix(entry.Name()).StripSingleChar
      continue
    end
  end

  return error
end

return curEntry
```

ekinanp:
[...]

PR #158, opened 2019-03-21
Attr() cleanup; Shrink Entry interface; use plugin.CachedOp + plugin.CachedMetadata

ekinanp:


fuse/core.go
```diff
@@ -51,6 +52,14 @@ func getIDs() (uint32, uint32) {
 
 var uid, gid = getIDs()
 
+func joinPath(parentPath string, e plugin.Entry) string {
```
ekinanp: Sorry to re-introduce this again. Figured it's better long-term, since it separates the FUSE concept of a path (which is filesystemy) from Wash's concept (which right now is filesystemy, but may not be in the future depending on how annoying the issue described in https://github.com/puppetlabs/wash/issues/156 is)

plugin/aws/ec2InstanceConsoleOutput.go
```diff
@@ -29,30 +26,29 @@ func newEC2InstanceConsoleOutput(inst *ec2Instance, latest bool) *ec2InstanceCon
...
+	output, err := cl.inst.cachedConsoleOutput(ctx, cl.latest)
 	if err != nil {
-		return nil, err
+		return plugin.Attributes{}, err
 	}
 
-	content, err := base64.StdEncoding.DecodeString(awsSDK.StringValue(resp.Output))
+	cl.Ctime = output.mtime
+	attr, _ := cl.EntryBase.Attr(ctx)
+	attr.Size = uint64(len(output.content))
```
ekinanp: Do we even need `Size` ? Could this be calculated from `SizedReader` instead (and have EntryBase mark the Size attribute as unknown by default)? Am leaning towards that. This way, it's clear that an S3 object's Ctime is its Mtime (which is precisely how S3 models its objects -- they're atomic blobs).

plugin/aws/ec2InstanceMetadataJSON.go
```diff
@@ -23,7 +23,7 @@ func newEC2InstanceMetadataJSON(inst *ec2Instance) *ec2InstanceMetadataJSON {
 }
 
 func (im *ec2InstanceMetadataJSON) Open(ctx context.Context) (plugin.SizedReader, error) {
```
ekinanp: Could we turn off caching for Open here? How much overhead would prettifying the metadata impose on performance?

plugin/entryBase.go
```diff
@@ -67,9 +69,19 @@ func (e *EntryBase) ID() string {
...
+// are set to this value.
+//
+// Entries can override Attr if they'd like to return a different
+// set of attributes.
 func (e *EntryBase) Attr(ctx context.Context) (Attributes, error) {
-	return e.attr, nil
+	return Attributes{
+		Ctime: e.Ctime,
+		Mtime: e.Ctime,
+		Atime: e.Ctime,
```
ekinanp: Should plugin authors worry about `Atime` ?  It seems like if you do anything with any entry in Wash, you're "accessing" it so `Atime` should always be `time.Now()`. Hm, maybe for sshfs-type files it could be useful.

plugin/types.go
```diff
@@ -39,10 +39,10 @@ import (
 // creating your plugin objects.
 type Entry interface {
 	Name() string
-	ID() string
 	Attr(ctx context.Context) (Attributes, error)
-	getTTLOf(op actionOpCode) time.Duration
+	id() string
```
ekinanp: Unexporting the id also prevents plugin authors from overwriting it.

plugin/entryBase.go
MikaelSmith: `Atime` generally means when it was "read". For actual files it makes sense. For other things I'm not sure.

ekinanp: Oh ok. Might as well keep it then.

plugin/aws/ec2Instance.go
```diff
@@ -64,16 +44,15 @@ const (
...
 	ec2Instance := &ec2Instance{
 		EntryBase:        plugin.NewEntry(ID),
 		session:          session,
 		client:           client,
 		ssmClient:        ssm.New(session),
 		cloudwatchClient: cloudwatchlogs.New(session),
-		attr:             attr,
 	}
-	ec2Instance.TurnOffCachingFor(plugin.List)
+	ec2Instance.TurnOffCaching()
```
MikaelSmith: I'd definitely like a comment here. `TurnOffCaching` doesn't really describe what's happening anymore either; it's more like `DisableDefaultCaching` (because it only affects the ops that are native to the plugin package).

plugin/aws/ec2InstanceMetadataJSON.go
MikaelSmith: I think we could, it should be very minimal overhead.

plugin/aws/s3Object.go
```diff
@@ -36,49 +36,62 @@ func newS3Object(bucket string, key string, client *s3Client.S3) *s3Object {
...
 }
 
 func (o *s3Object) Attr(ctx context.Context) (plugin.Attributes, error) {
-	resp, err := o.cachedHeadObject(ctx)
+	result, err := o.cachedHeadObject(ctx)
 	if err != nil {
 		return plugin.Attributes{}, nil
 	}
 
-	size := awsSDK.Int64Value(resp.ContentLength)
```
MikaelSmith: Why did you move this within the cached operation?

plugin/aws/ec2Instance.go
ekinanp: Agreed. I think `TurnOffCachingFor` => `DisableCachingFor`, `TurnOffCaching` => `DisableDefaultCaching` makes more sense.

plugin/entryBase.go
```diff
@@ -24,11 +24,13 @@ var actionOpCodeToNameMap = [3]string{"List", "Open", "Metadata"}
 // EntryBase implements Entry, making it easy to create new entries.
 // You should use plugin.NewEntry to create new EntryBase objects.
 type EntryBase struct {
-	name string
+	// Ctime specifies the entry's creation time. See Attr for more
+	// details on how this is used.
+	Ctime time.Time
```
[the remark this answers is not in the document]
ekinanp: Yeah, I think you're right. I'll revert it

plugin/aws/s3Object.go
ekinanp: So there wouldn't be a race condition on `Ctime` (since I'm setting it then using EntryBase's `Attr`). But that indicates `Ctime` is a bad idea, so I'll go ahead and remove it then take this out of the cached operation.

plugin/entryBase.go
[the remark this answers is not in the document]
ekinanp: It feels weird to me b/c it's early taxonomy that doesn't give a lot of value and is an extra thing plugin authors have to keep in mind. It's much simpler to just say "Implement Attr if your entry has attributes".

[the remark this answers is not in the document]
MikaelSmith: It does make some plugins cleaner. And it should never change. It feels like a very pragmatic improvement.

plugin/aws/s3Object.go
MikaelSmith: Ah, you also want to be careful calling CachedOp from within another CachedOp. They could end up using the same mutex and deadlock.

plugin/entryBase.go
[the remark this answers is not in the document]
ekinanp: Hmm, I think this could be less weird if we reinforced the `it should never change` fact. I'll explore that.

ekinanp:
Accidentally pushed up the caching updates. Still addressing the remaining feedback in the PR.

plugin/cache.go
```diff
@@ -87,6 +88,10 @@ type opFunc func() (interface{}, error)
 // For example, CachedOp could be useful to cache an API request whose response
 // lets you implement Attr() and Metadata() for the given entry.
 func CachedOp(opName string, entry Entry, ttl time.Duration, op opFunc) (interface{}, error) {
+	if !opNameRegex.MatchString(opName) {
```
MikaelSmith: This isn't quite right. `opNameRegex` allows anything to follow after the regex, it needs to end with `$`.

plugin/cache.go
```diff
@@ -49,8 +49,8 @@ func UnsetTestCache() {
 	cache = nil
 }
 
-var opNameRegex = regexp.MustCompile("^[a-zA-Z]+")
-var opQualifier = opNameRegex.String() + "::"
+var opNameRegex = regexp.MustCompile("^[a-zA-Z]+$")
+var opQualifier = "^[a-zA-Z]+::"
```
MikaelSmith: This could be a `const`.

MikaelSmith:
```suggestion
const opQualifier = "^[a-zA-Z]+::"
```

PR #166, opened 2019-03-25
 Allow entry names to have '/'es by introducing cnames

ekinanp:


plugin/helpers.go
```diff
@@ -18,6 +18,36 @@ import (
 // DefaultTimeout is the default timeout for prefetching
 var DefaultTimeout = 10 * time.Second
 
+/*
+CName returns the entry's canonical name, which is what Wash uses
+to construct the entry's path (see plugin.Path). The entry's cname
+is e.Name(), but with all '/' characters replaced by a '#' character.
```
ekinanp: I decided to go ahead and do the `#` replacement because it simplifies the CName semantics.

plugin/registry.go
```diff
@@ -31,13 +32,25 @@ func (r *Registry) Plugins() map[string]Root {
 	return r.plugins
 }
 
+var pluginNameRegex = regexp.MustCompile("^[0-9a-zA-Z_-]+$")
```
ekinanp: Figured we should probably just enforce this.

plugin/helpers.go
```diff
@@ -18,6 +18,36 @@ import (
...
+CNames are unique. CName uniqueness is checked in plugin.CachedList.
+
+NOTE: The '#' character was chosen because it is unlikely to appear in
+a meaningful entry's name. If, however, there's a good chance that an
+entry's name can contain the '#' character, and that two entries can
+have the same cname (e.g. 'foo/bar', 'foo#bar'), then you can use
+e.SetSlashReplacementChar(<char>) to change the default slash replacement
+character from a '#' to <char>.
+*/
+func CName(e Entry) string {
```
ekinanp: Turns out that making the path a query parameter was a good idea -- the `#` broke the existing URL format because it'd omit the latter part of the path (e.g. `/foo/bar#baz` was interpreted as (`/foo/bar`).

ekinanp:
Blocked on https://github.com/puppetlabs/wash/pull/164

plugin/helpers.go
MikaelSmith: Yeah, I figured there were some common names that would break it that we just hadn't tested.

plugin/entryBase.go
```diff
@@ -74,6 +76,19 @@ func (e *EntryBase) Attr(ctx context.Context) (Attributes, error) {
 	}, nil
 }
 
+func (e *EntryBase) cname() string {
+	return strings.Replace(
```
MikaelSmith: This could also be `ReplaceAll`. That seems more clear.

ekinanp: That's in Go 1.12 unfortunately

plugin/helpers.go
MikaelSmith: Why is this a separate function rather than directly exporting the `CName` method on `EntryBase`?

MikaelSmith: Oh, is it because you don't want someone to be able to override it? Can you add a comment to that effect on `cname`?

plugin/aws/s3Bucket.go
```diff
@@ -46,23 +47,27 @@ func listObjects(ctx context.Context, client *s3Client.S3, bucket string, prefix
 	)
 
 	for i, p := range resp.CommonPrefixes {
-		prefix := awsSDK.StringValue(p.Prefix)
-		// Skip the top-level '/' prefix, it would be redundant to list it.
```
MikaelSmith: Is it not really redundant? Is there going to be a directory named `#` that includes the same stuff as we list at top-level?

plugin/aws/s3Bucket.go
```diff
@@ -46,23 +47,27 @@ func listObjects(ctx context.Context, client *s3Client.S3, bucket string, prefix
...
+		if name == "" {
+			// Key has a trailing "/", so skip it
+			continue
+		}
+		entries[i] = newS3Object(name, bucket, key, client)
+		i++
 	}
 
-	return entries, nil
+	return entries[0:i], nil
```
MikaelSmith: This could be simpler by creating `entries = make([]plugin.Entry, 0, numPrefixes+numObjects)` and using `entries = append(entries, ...)` everywhere. Then you don't have to track index.

plugin/aws/s3Bucket.go
ekinanp: Figured if the AWS console supports it, we might as well support it too if it's easy to do so. Also, you could (somehow) upload a `/` folder (e.g. something like `/<foo>/<bar>/<baz>`).

MikaelSmith: I'm not quite sure what all's happening with CommonPrefixes and Contents. Maybe an example comment would help.

plugin/aws/s3Dir.go
```diff
@@ -38,10 +38,10 @@ func (s *s3Dir) List(ctx context.Context) ([]plugin.Entry, error) {
 
 	buckets := make([]plugin.Entry, len(resp.Buckets))
 	for i, bucket := range resp.Buckets {
-		name := bucket.Name
+		name := awsSDK.StringValue(bucket.Name)
 
 		locRequest := &s3Client.GetBucketLocationInput{
-			Bucket: name,
+			Bucket: awsSDK.String(name),
```
MikaelSmith: How about use `bucket.Name` directly here so you don't have to cast twice?

ekinanp: I remember running into some weird things when I did that for other values, but since this code worked OK before, I can go back to re-using `bucket.Name`.

plugin/aws/s3Bucket.go
```diff
@@ -45,12 +114,14 @@ func listObjects(ctx context.Context, client *s3Client.S3, bucket string, prefix
 		numObjects,
 	)
 
+	// resp.CommonPrefixes represents all of the object keys
 	for _, p := range resp.CommonPrefixes {
 		commonPrefix := awsSDK.StringValue(p.Prefix)
 		name := strings.TrimPrefix(commonPrefix, prefix)
 		if name != "/" {
 			name = strings.TrimSuffix(name, "/")
```
MikaelSmith: This will probably do bad things for your example `//`.

plugin/aws/s3Bucket.go
```diff
@@ -45,21 +114,25 @@ func listObjects(ctx context.Context, client *s3Client.S3, bucket string, prefix
...
-		entries[numPrefixes+i] = newS3Object(
-			bucket,
-			awsSDK.StringValue(o.Key),
-			client,
-		)
+	for _, o := range resp.Contents {
+		key := awsSDK.StringValue(o.Key)
+		name := strings.TrimPrefix(key, prefix)
+		if name == "" {
+			// Key has a trailing "/", so skip it
```
MikaelSmith: I'm still not sure how this comment connects to testing `name == ""`.

plugin/aws/s3Bucket.go
ekinanp: `//` => `#/#` in Wash. That's b/c the prefix there will be `/`, so name will == `/`.

PR #170, opened 2019-03-26
Delay getting region for bucket

MikaelSmith:
Querying the region for every bucket when listing available buckets
results in many queries to list them. We only need the region when we're
listing the contents of the bucket. Delay getting the region name until
we need it.

The region result will be cached indefinitely. If there's an error,
delete the cache entry for the bucket to force a retry.

This saves ~4 seconds for listing an s3 account with ~180 buckets. It
now takes ~4 seconds to do the initial listing, then ~100ms for each
subsequent listing when the cache expires.

### DIFF
--- a/plugin/aws/s3Bucket.go
+++ b/plugin/aws/s3Bucket.go
@@ -2,14 +2,18 @@ package aws
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
+	"github.com/ekinanp/go-cache"
 	"github.com/puppetlabs/wash/journal"
 	"github.com/puppetlabs/wash/plugin"
 
+	"github.com/aws/aws-sdk-go/aws"
 	awsSDK "github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/session"
 	s3Client "github.com/aws/aws-sdk-go/service/s3"
 )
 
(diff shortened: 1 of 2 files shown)


plugin/aws/s3Bucket.go
```diff
@@ -198,7 +207,32 @@ func (b *s3Bucket) Metadata(ctx context.Context) (plugin.MetadataMap, error) {
...
+		return nil, err
+	}
+	metadata["region"] = region
 
 	return metadata, nil
 }
+
+func (b *s3Bucket) getRegion(ctx context.Context) (string, error) {
+	var err error
+	b.regioner.Do(func() {
```
ekinanp: I'd have this try again if getting the region returns an error (the retry would happen when it's invoked). You can do that by resetting the regioner in the `Do` function (I do something similar in https://github.com/puppetlabs/wash/blob/master/plugin/aws/ec2Instance.go#L217), and storing the region error in a `regionErr` variable.

MikaelSmith: Updated.

plugin/aws/s3Bucket.go
```diff
@@ -198,7 +207,33 @@ func (b *s3Bucket) Metadata(ctx context.Context) (plugin.MetadataMap, error) {
...
+			// TODO: Should we log a warning and continue instead of returning an error?
+			err = fmt.Errorf("could not get the region of bucket %v: %v", b.Name(), reqErr)
+		}
+
+		// The response will be empty if the bucket is in Amazon's default region (us-east-1)
+		b.region = s3Client.NormalizeBucketLocation(awsSDK.StringValue(resp.LocationConstraint))
+		// Update client to be region-specific
+		b.client = s3Client.New(b.session, aws.NewConfig().WithRegion(b.region))
+	})
+	if err != nil {
```
ekinanp: Oh yeah, this code will only get executed by a single thread b/c err's set in Once which also gets executed by that same thread. A comment describing that would help because it requires a bit of thinking.

MikaelSmith:
The once stuff was getting hard to reason about, went with using caching instead.

PR #174, opened 2019-03-27
Add support for IAM Roles

### LINKED ISSUE #168: Introduce capability to assume IAM roles to the AWS plugin
This work requires duplicating the functionality specified in https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html.

MikaelSmith:
Adds support for IAM Roles when configured according to https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html.

Includes file caching of credentials requested with MFA, and uses that to prompt for credentials on startup and cache them for later sessions.

Credentials retrieved currently last 15 minutes. See aws/aws-sdk-go#2532 for discussion of why we haven't modified that yet.

Known issue: `region` is not correctly inherited from the `source_profile`.

Resolves #168.

### DIFF
--- a/plugin/aws/profile.go
+++ b/plugin/aws/profile.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"context"
 
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/puppetlabs/wash/journal"
 	"github.com/puppetlabs/wash/plugin"
@@ -23,8 +24,9 @@ func newProfile(ctx context.Context, name string) (*profile, error) {
 	// Create the session. SharedConfigEnable tells AWS to load the profile
 	// config from the ~/.aws/credentials and ~/.aws/config files
 	session, err := session.NewSessionWithOptions(session.Options{
-		Profile:           profile.Name(),
-		SharedConfigState: session.SharedConfigEnable,
+		Profile:                 profile.Name(),
+		AssumeRoleTokenProvider: stscreds.StdinTokenProvider,
+		SharedConfigState:       session.SharedConfigEnable,
 	})
 	if err != nil {
 		return nil, err
--- a/plugin/aws/profile.go
+++ b/plugin/aws/profile.go
@@ -2,7 +2,9 @@ package aws
 
 import (
 	"context"
+	"fmt"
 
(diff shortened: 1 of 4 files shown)


plugin/aws/profile.go
```diff
@@ -23,8 +24,9 @@ func newProfile(ctx context.Context, name string) (*profile, error) {
 	// Create the session. SharedConfigEnable tells AWS to load the profile
 	// config from the ~/.aws/credentials and ~/.aws/config files
 	session, err := session.NewSessionWithOptions(session.Options{
-		Profile:           profile.Name(),
-		SharedConfigState: session.SharedConfigEnable,
+		Profile:                 profile.Name(),
+		AssumeRoleTokenProvider: stscreds.StdinTokenProvider,
```
ekinanp: How long does the session created w/ `AssumeRoleTokenProvider` last? One thing I'm concerned about is that you can't use Wash's API with this unless you explicitly provide the MFA token via Stdin, so if you do something like `wash ls` inside the root, wouldn't Wash hang?

ekinanp: Nice to see that it's very easy to add the MFA stuff though

MikaelSmith: I'm not sure how long the session lasts. For the CLI it'll re-use a cached one for 6 hours. I'll see if I can find a documented answer, it wouldn't be great to have wash stop working after a couple of hours and have to restart it.

The experience of using this is that the wash server starts up normally, but the first time you list the `aws` plugin the wash server will prompt for an MFA token. My preference would be to force that to happen during server startup.

MikaelSmith: It looks like the duration of the MFA authorization is specified as part of IAM policy. So we'll need to iterate on how re-authorization should work.

ekinanp: Hmm, why not have this code be part of the `Init` method then? Seems like the user should be prompted for the MFA token when loading the AWS plugin.

We could handle re-authorization later as long as the session lasts long enough for us to do anything meaningful. That could be a known issue until that point, since it looks like we're going to be needing the MFA stuff in order to do anything useful with the AWS plugin.

MikaelSmith: Ok, authorization is actually only requested when we need it, like when making a query to list S3 resources. So have to go pretty deep to trigger it.

PR #176, opened 2019-03-29
Generalize the concept of attributes and metadata

ekinanp:


plugin/cache.go
```diff
@@ -221,9 +221,13 @@ func cachedOp(opName string, entry Entry, ttl time.Duration, op opFunc) (interfa
...
-		// The Registry's ID is set to "/", while all other entries' IDs are
-		// set in CachedList. Thus, we should never hit this code-path.
-		panic("entry.id() returned an empty ID")
+		// This can happen if the entry's just been created, and it calls one of the Cached*
+		// methods to query its Metadata so that it can set its initial attributes. In that case,
+		// just invoke op. Ideally we should panic here, but panicking makes it significantly more
+		// annoying for plugin authors to write their code. Since the plugin package manages the
+		// entry's ID, and the entry's ID is set in CachedList, an empty ID is such a rare edge case
+		// that it's better to not annoy people.
+		return op()
```
ekinanp: The panic behavior led me to either moving the cached op to the parent, or creating separate uncached/cached methods, where I'd use the former to compute the initial attributes in a parent's List call, and the latter for subsequent operations.

plugin/entryAttributes.go
```diff
@@ -0,0 +1,181 @@
...
+	var meta EntryMetadata
+	// Internal error if not a JSON object
+	errz.Fatal(json.Unmarshal(inrec, &meta))
+	return meta
+}
+
+// EntryAttributes represents an entry's attributes. We use a struct
+// instead of a map for efficient memory allocation/deallocation,
+// which is needed to make Group#List fast.
+type EntryAttributes struct {
```
ekinanp: I wonder if we should include the entry's supported actions here via an `actions` attribute. Could fill this info. in `CachedList`.

plugin/entryBase.go
```diff
@@ -87,13 +91,42 @@ func (e *EntryBase) setID(id string) {
...
+
+// Sync syncs the attribute with the given key's value in the entry's
+// metadata whenever the latter's refreshed via a call to plugin.CachedMetadata.
+// Use Sync if you expect the attribute to change.
+//
+// TODO: Find a way to let plugin authors specify a munging function. We'll
+// need this for the State attribute, and in case the metadata returns some of
+// the other attributes in a different format (e.g. like an arbitrarily formatted
+// string for one of the time fields)
+func (e *EntryBase) Sync(attr SyncableAttribute, key string) {
```
ekinanp: Was thinking of making this something like
```
Sync(SyncableAttribute, key, mungeFunc...)
```

(using Go's variadic args to simulate optional input).

plugin/entryBase.go
```diff
@@ -87,13 +91,42 @@ func (e *EntryBase) setID(id string) {
...
+	// TODO: Check if a hash can be used here.
```
ekinanp: I didn't go w/ a hash because even an empty map can take up 40+ bytes of memory. That can add-up when you have hundreds of thousands of entries (+ it could risk slowing list down). If neither of those things are an issue, I'm happy to use a hash.

plugin/externalPluginEntry.go
```diff
@@ -15,6 +15,9 @@ import (
 	"github.com/puppetlabs/wash/journal"
 )
 
+/*
+TODO: Come back and rethink attributes once all the attributes/metadata work is finished.
```
ekinanp: I'll probably do this much later, after we've got some use-cases addressed. The main fixes would likely be to beef-up the attributes key w/ more attributes + add a separate `synced_attributes` key.

plugin/helpers.go
```diff
@@ -163,49 +153,27 @@ func ToFileMode(mode interface{}) (os.FileMode, error) {
...
-			}
-
-			attr.Size = uint64(size)
-		}
+// RefreshAttributes refreshes the entry's attributes. It is primarily
+// needed by FUSE because FUSE nodes have a very long lifetime.
+func RefreshAttributes(ctx context.Context, e Entry) error {
+	journal.Record(ctx, "Refreshing the attributes of %v", Path(e))
+
+	meta, err := CachedMetadata(ctx, e)
```
ekinanp: Can't put this logic in EntryBase b/c it will call EntryBase's Metadata method.

plugin/syncableAttribute.go
```diff
@@ -0,0 +1,170 @@
...
+		if err != nil {
+			return err
+		}
+
+		entry.attr.SetSize(size)
+		return nil
+	},
+}
+
+func mungeToTimeVal(v interface{}) (time.Time, error) {
```
ekinanp: May be worth having a separate `munge` package, or at least a `plugin/munge.go` file.

plugin/syncableAttribute.go
```diff
@@ -0,0 +1,170 @@
...
+	switch t := v.(type) {
+	case time.Time:
+		return t, nil
+	case string:
+		// TODO: The layout should be specified by the plugin author.
+		// We could likely have helpers that generate a munging function
+		// given some parameters. For now, this is enough.
+		layout := "2006-01-02T15:04:05Z"
```
ekinanp: This was to get S3 object syncs working, where the last modified time is in this format.

volume/helpers_test.go
```diff
@@ -29,27 +29,27 @@ const fixture = `
...
 	assert.Equal(t, "mnt/path", path)
-	assert.Equal(t, plugin.Attributes{
-		Atime: time.Unix(1550611510, 0),
-		Mtime: time.Unix(1550611448, 0),
-		Ctime: time.Unix(1550611448, 0),
-		Mode:  0755 | os.ModeDir,
-		Size:  96,
-	}, attr)
+	expectedAttr := plugin.EntryAttributes{}
+	expectedAttr.SetAtime(time.Unix(1550611510, 0))
```
ekinanp: This is a bit verbose. I was thinking of using the builder pattern, but it's kind of weird in Go. Something like
```
expectedAttr.
  SetAtime(<time>).
  SetCtime(<time>).
  SetMtime(<time>)
```

Dunno if that reads well. I don't personally mind it.

plugin/entryBase.go
```diff
@@ -87,13 +91,42 @@ func (e *EntryBase) setID(id string) {
...
+// Name returns the entry's name as it was passed into
+// plugin.NewEntry. You should use e.Name() when making
+// the appropriate API calls within your plugin.
+func (e *EntryBase) Name() string {
+	return e.name()
+}
+
+// SetInitialAttributes sets the entry's initial attributes. Use it
+// after creating the entry via a call to NewEntry.
+func (e *EntryBase) SetInitialAttributes(attr EntryAttributes) {
```
ekinanp: I chose not to cache `meta` b/c I'm thinking that hundreds of thousands of lock acquisitions in `List` could slow things down. Happy to cache if that isn't an issue, although that could introduce other problems (e.g. how do you know this `meta`'s the freshest [if we just override it]?).

cmd/ls.go
```diff
@@ -36,10 +36,7 @@ func formatListEntries(apiPath string, ls []apitypes.ListEntry) string {
 	table := make([][]string, len(ls))
 	for i, entry := range ls {
 		var ctimeStr string
-		if err, ok := entry.Errors["attributes"]; ok {
-			cmdutil.ErrPrintf("errored on %v: could not retrieve the attributes: %v\n", entry.Name, err)
-			ctimeStr = "<unknown>"
-		} else if entry.Attributes.Ctime.IsZero() {
+		if entry.Attributes.Ctime == nil {
```
ekinanp: `wash ls` is pretty slow compared with `ls` (~4s to list S3 buckets using Michael's account). I think the latency's because of FUSE having to do a find to get the xattrs. If so, it's probably good to remove the xattr dependency soon.

plugin/kubernetes/pod.go
```diff
@@ -19,56 +19,98 @@ import (
...
+	meta := plugin.ToMeta(pdInfo.pd)
+	meta["LogSize"] = len(pdInfo.logContent)
+	return meta
+}
+
+func (p *pod) cachedPodInfo(ctx context.Context, queryPodMeta bool) (podInfoResult, error) {
+	cachedPdInfo, err := plugin.CachedOp("PodInfo", p, 15*time.Second, func() (interface{}, error) {
+		result := podInfoResult{}
+
+		if queryPodMeta {
```
ekinanp: This is here to avoid an unnecessary query to `Pods` when the current pod's being created. Note that the result here won't get cached b/c the pod has no id, so this code's safe.

May be worth rethinking this at some point.

plugin/aws/ec2InstanceConsoleOutput.go
```diff
@@ -28,25 +32,82 @@ func newEC2InstanceConsoleOutput(inst *ec2Instance, latest bool) *ec2InstanceCon
...
+
+	attr := plugin.EntryAttributes{}
+	attr.SetCtime(output.mtime)
+	attr.SetMtime(output.mtime)
+	attr.SetAtime(output.mtime)
+	attr.SetSize(uint64(len(output.content)))
+	attr.SetMeta(output.toMeta())
+	cl.SetInitialAttributes(attr)
+
+	cl.Sync(plugin.CtimeAttr(), "LastModified")
```
ekinanp: I was also toying with something like `Sync(key, ...SyncableAttrField)` but these seem like special enough cases that it isn't worth it (especially since it makes specifying a munge function difficult).

plugin/aws/ec2InstanceMetadataJSON.go
```diff
@@ -15,26 +15,57 @@ type ec2InstanceMetadataJSON struct {
 	inst *ec2Instance
 }
 
-func newEC2InstanceMetadataJSON(inst *ec2Instance) *ec2InstanceMetadataJSON {
+func newEC2InstanceMetadataJSON(ctx context.Context, inst *ec2Instance) (*ec2InstanceMetadataJSON, error) {
```
ekinanp: One interesting thing about these read-only files is that we can track an `mtime` for them. Probably not worth it for the metadata.json files, since we can just track the specific VM/container/pod's mtime.

fuse/core.go
```diff
@@ -51,70 +52,90 @@ func getIDs() (uint32, uint32) {
...
 
 // Applies attributes where non-default, and sets defaults otherwise.
-func applyAttr(a *fuse.Attr, attr *plugin.Attributes) {
-	a.Valid = 1 * time.Second
-	if attr.Valid != 0 {
-		a.Valid = attr.Valid
-	}
+func (f *fuseNode) applyAttr(a *fuse.Attr, attr *plugin.EntryAttributes) {
+	// This doesn't quite work for some reason.
+	a.Valid = attrRefreshInterval
```
ekinanp: @MikaelSmith You were right, this doesn't work that well. It's inconsistent.

ekinanp:
I did some basic tests with docker containers and AWS S3/EC2 instances to make sure everything's working (especially the refreshing). It all seems to work well, and `ls` is reasonably fast (aside from the simultaneous refresh case). Could use some help in testing docker volumes + the kube plugin, since I'm not familiar with them.

cmd/ls.go
MikaelSmith: It's a theory worth testing. I wouldn't have expected it to be slower because the query to get xattrs should result in later activity being cached.

plugin/aws/ec2InstanceConsoleOutput.go
MikaelSmith: Yeah, doesn't seem worth it.

plugin/aws/ec2InstanceMetadataJSON.go
MikaelSmith: An mtime as in keep track of when we get new data that differs from the previous data? I think that would be interesting idea for later.

MikaelSmith: History of metadata is another interesting idea in that vein.

plugin/aws/s3Object.go
```diff
@@ -23,14 +22,29 @@ type s3Object struct {
...
 		bucket:    bucket,
 		key:       key,
 		client:    client,
 	}
-	o.DisableCachingFor(plugin.Metadata)
+	o.DisableCachingFor(plugin.MetadataOp)
+
+	attr := plugin.EntryAttributes{}
+	mtime := awsSDK.TimeValue(rawObj.LastModified)
+	attr.SetCtime(mtime)
```
MikaelSmith: We have no way to get creation time? Or will it never differ in S3? Could use a comment clarifying why it's LastModified given that these are very file-like.

plugin/aws/s3Object.go
```diff
@@ -23,14 +22,29 @@ type s3Object struct {
...
+	attr.SetMtime(mtime)
+	attr.SetAtime(mtime)
+	// TODO: Export a mungeSize helper to abstract away the common
+	// logic of validating a negative size
+	attr.SetSize(uint64(awsSDK.Int64Value(rawObj.Size)))
+	o.SetInitialAttributes(attr)
+
+	o.Sync(plugin.CtimeAttr(), "LastModified")
```
MikaelSmith: Every plugin has some unenforced tie between Sync name and the result of Metadata. I'd like to come up with a way to simply validate those are consistent, but not seeing anything besides using it against a real (or mocked) API.

plugin/cache.go
MikaelSmith: Ah right, because they're using their own caching now. This seems like a reasonable compromise.

plugin/docker/container-log.go
```diff
@@ -28,28 +29,56 @@ func (clf *containerLogFile) isTty(ctx context.Context) bool {
...
-		}
-	} else {
-		// Write stdout and stderr to the same buffer.
-		if n, err = stdcopy.StdCopy(&buf, &buf, rdr); err != nil {
-			return nil, err
-		}
+	return cachedContent.([]byte), nil
+}
+
+func (clf *containerLogFile) Metadata(ctx context.Context) (plugin.EntryMetadata, error) {
```
MikaelSmith: So this essentially gets rid of the default behavior of using `Open` to determine the size implicitly?

MikaelSmith: Also, why add a separate method rather than using CachedOpen directly?

plugin/docker/container.go
```diff
@@ -15,35 +14,43 @@ import (
...
-	return plugin.Attributes{
-		Ctime: c.startTime,
-		Mtime: c.startTime,
-		Atime: c.startTime,
-	}, nil
+	return plugin.ToMeta(raw), nil
 }
 
 func (c *container) List(ctx context.Context) ([]plugin.Entry, error) {
+	// TODO: May be worth creating a helper that makes it easy to create
```
MikaelSmith: Yes, I think it'd make it more readable. The different behavior for `meta["Size"]` is kind of annoying though. How does the `Sync` behavior handle it?

plugin/entryAttributes.go
MikaelSmith: What benefit would that bring? Where do we currently do it?

volume/helpers_test.go
MikaelSmith: I don't mind it either.

plugin/entryBase.go
```diff
@@ -87,13 +91,42 @@ func (e *EntryBase) setID(id string) {
...
 
-func (e *EntryBase) getTTLOf(op actionOpCode) time.Duration {
+func (e *EntryBase) getTTLOf(op defaultOpCode) time.Duration {
 	return e.ttl[op]
 }
 
 // OTHER METHODS USED TO FACILITATE PLUGIN DEVELOPMENT
 // AND TESTING
 
+// Name returns the entry's name as it was passed into
```
MikaelSmith: I'm still unclear why both `Name` and `plugin.Name` exist.

plugin/entryBase.go
MikaelSmith: Or just another function like `SyncMunged`.

plugin/entryBase.go
MikaelSmith: Array definitely makes the most sense here. I'd change the TODO to a comment like the one you posted here.

plugin/externalPluginEntry.go
```diff
@@ -33,6 +36,7 @@ func unixSecondsToTimeAttr(seconds int64) time.Time {
 	return time.Unix(seconds, 0)
 }
 
+/*
```
MikaelSmith: Why are you keeping this around?

plugin/helpers.go
```diff
@@ -12,17 +11,27 @@ import (
...
 	log "github.com/sirupsen/logrus"
 )
 
 // DefaultTimeout is the default timeout for prefetching
 var DefaultTimeout = 10 * time.Second
 
+/*
+Name returns the entry's name as it was passed into
+plugin.NewEntry. It is meant to be called by other
+Wash packages. Plugin authors should use EntryBase#Name
```
MikaelSmith: I'm still not sure why both of these are needed.

plugin/helpers.go
MikaelSmith: Seems like a useful code comment.

plugin/syncableAttribute.go
```diff
@@ -0,0 +1,170 @@
+package plugin
+
+import (
+	"fmt"
+	"time"
+)
+
+// AtimeAttr represents the Atime attribute in plugin.EntryAttributes
+func AtimeAttr() SyncableAttribute {
```
MikaelSmith: I think these should all return pointers to the type so that every Sync call shares a pointer to a single instance rather than a copy of the SyncableAttribute.

plugin/syncableAttribute.go
MikaelSmith: I agree with breaking it out in a separate file.

plugin/kubernetes/pod.go
MikaelSmith: It's kind of weird code. Is there another way we could refactor it?

plugin/entryAttributes.go
ekinanp: Was thinking symmetry reasons (`actions` really are an attribute) but now that you mention it, it could be useful as a quick optimization to the `[`IsSupportedOn` methods](https://github.com/puppetlabs/wash/blob/master/plugin/action.go#L16-L26)` b/c we can just calculate that once in `CachedList` and indefinitely store it.

plugin/syncableAttribute.go
ekinanp: I'll do that in a separate PR.

plugin/aws/ec2InstanceMetadataJSON.go
ekinanp: Or even a generalized `delta` for any file (that's probably too ambitious). I'll add a TODO in one of the metadata.json files to track this discussion later.

api/types/list.go
```diff
@@ -1,9 +1,21 @@
...
 
 import (
-	"github.com/puppetlabs/wash/plugin"
+	"os"
+	"time"
 )
 
+// EntryAttributes represents an entry's attributes.
+//
+// TODO: Add meta later
```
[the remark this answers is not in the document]
ekinanp: Ah yeah, that's all the info. that `Entry#Metadata` returns. I'm probably thinking we should include it for things that get it from `List` and maybe make querying the `Metadata` endpoint optional.

fuse/core.go
```diff
@@ -51,70 +52,90 @@ func getIDs() (uint32, uint32) {
...
+
+	if attr.HasMode() {
+		a.Mode = attr.Mode()
+	} else if plugin.ListAction.IsSupportedOn(f.entry) {
+		a.Mode = os.ModeDir | 0550
+	} else {
+		a.Mode = 0440
+	}
+
+	a.Size = ^uint64(0)
```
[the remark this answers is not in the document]
ekinanp: Ok

fuse/core.go
```diff
@@ -51,70 +52,90 @@ func getIDs() (uint32, uint32) {
...
-func attr(ctx context.Context, f fuseNode, a *fuse.Attr) error {
-	attr, err := plugin.Attr(ctx, f.Entry())
-	if _, ok := err.(plugin.ErrCouldNotDetermineSizeAttr); ok {
-		log.Warnf("FUSE: Warn[Attr,%v]: %v", f, err)
-	} else if err != nil {
-		log.Warnf("FUSE: Error[Attr,%v]: %v", f, err)
-		return err
+func (f *fuseNode) Attr(ctx context.Context, a *fuse.Attr) error {
+	// TODO: The code below's a temporary hack to show that the refreshing
+	// behavior works. It blows up after several successive ls calls b/c each
```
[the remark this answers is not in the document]
ekinanp: When I would repeatedly list things in `S3`, it significantly slowed down and even halted once we reached the refresh interval b/c of all the buckets were fetching their metadata to refresh their attributes.

Hmm, it may be worth _not_ triggering a refresh if the entry doesn't have any synced attributes. I'll add that.

plugin/externalPluginEntry.go
ekinanp: So I can look at it later once I get around to doing it. Don't mind removing it if that's preferred too.

plugin/helpers.go
ekinanp: This is needed by the API, or anything that needs the entry's name. That's b/c `Name` isn't part of the entry interface. Can clarify it a bit more.

plugin/syncableAttribute.go
ekinanp: Sure, I can do that.

plugin/externalPluginEntry.go
ekinanp: So I can look at it later once I get around to doing it. Don't mind removing it if that's preferred too.

plugin/docker/container-log.go
ekinanp: Oh, good idea. I'll update.

plugin/docker/container.go
ekinanp: OK, I can add a `plugin.NewMetadataJSONFile` helper.

The `Sync` behavior handles it by munging the sizes -- see https://github.com/ekinanp/wash/blob/common-metadata/plugin/syncableAttribute.go#L147-L170

ekinanp: Actually, let me add the helper once we see more use-cases some up, and when the `metadata.json` files are something that people like. I don't think keeping it the way it is is that bad right now.

plugin/aws/s3Object.go
MikaelSmith: I mean: something that makes sure that the string you pass in for `Sync` actually matches how you get it from `Metadata`.

ekinanp: Ah yeah. In that case, I don't know of another way either

plugin/docker/container-metadata.go
```diff
@@ -13,16 +13,36 @@ type containerMetadata struct {
 	container *container
 }
 
+func (cm *containerMetadata) Metadata(ctx context.Context) (plugin.EntryMetadata, error) {
```
MikaelSmith: This seems not ok. `CachedMetadata` -> `Metadata` -> `content` -> `CachedMetadata`. Either `Metadata` should call `CachedOpen` or `Open` should call `CachedMetadata`.

ekinanp: I think it should be b/c it calls container's CachedMetadata. However, these should use `Open` -- that makes a lot more sense. Will update.

ekinanp: Updated.

plugin/aws/ec2InstanceMetadataJSON.go
```diff
@@ -37,26 +37,17 @@ func newEC2InstanceMetadataJSON(ctx context.Context, inst *ec2Instance) (*ec2Ins
...
-	content, err := im.content(ctx)
-	if err != nil {
-		return nil, err
-	}
-
-	return bytes.NewReader(content), err
-}
-
-func (im *ec2InstanceMetadataJSON) content(ctx context.Context) ([]byte, error) {
 	meta, err := plugin.CachedMetadata(ctx, im.inst)
```
MikaelSmith: Oh, I mixed this up earlier. It's getting the instance's metadata, not a circular call.

plugin/kubernetes/pod.go
```diff
@@ -19,56 +19,98 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
+	pdInfo.pd = p
+
+	meta := pdInfo.toMeta()
+	attr := plugin.EntryAttributes{}
+	attr.SetCtime(p.CreationTimestamp.Time)
+	attr.SetAtime(attr.Ctime())
+	attr.SetMeta(meta)
```
MikaelSmith: This doesn't set Size anywhere (from `meta`) so it's not available when you list it.

ekinanp: Updated.

plugin/aws/s3Object.go
```diff
@@ -23,16 +22,38 @@ type s3Object struct {
...
+	//
+	// TODO: Export a mungeSize helper to abstract away the common
+	// logic of validating a negative size
+	mtime := awsSDK.TimeValue(o.LastModified)
+	attr := plugin.EntryAttributes{}
+	attr.
+		SetCtime(mtime).
+		SetMtime(mtime).
+		SetAtime(mtime).
+		SetSize(uint64(awsSDK.Int64Value(o.Size)))
```
MikaelSmith: Should this also include `SetMeta`?

ekinanp: No, b/c [head-object](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#S3.HeadObject) returns something different than what [list-objects returns](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#ListObjectsOutput).

This is brings up an important question -- do we want to maybe define `meta` to represent the complete metadata returned by a `List` endpoint? For example, I noticed that List for containers returns a decent amount of info (though not at the level of detail that RawInspect returns). That would make `find` more powerful and performant -- maybe we can add an option to use the detailed `meta` ?

PR #178, opened 2019-04-01
Create FUSE server obj, configure it to pass-in the PID

ekinanp:
This ensures that all of our FUSE methods write to the right journal
regardless of whether the PID's included in the method's request.

### DIFF
--- a/fuse/core.go
+++ b/fuse/core.go
@@ -10,6 +10,7 @@ import (
 
 	"bazil.org/fuse"
 	"bazil.org/fuse/fs"
+	"github.com/puppetlabs/wash/journal"
 	"github.com/puppetlabs/wash/plugin"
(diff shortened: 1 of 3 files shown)


fuse/core.go
```diff
@@ -177,8 +178,15 @@ func ServeFuseFS(filesys *plugin.Registry, mountpoint string) (chan<- bool, <-ch
 
 		log.Infof("FUSE: Serving filesystem")
 
+		serverConfig := &fs.Config{
+			WithContext: func(ctx context.Context, req fuse.Request) context.Context {
```
MikaelSmith: That seems easy...

ekinanp: Yup. Verified that it works by running a `stat` and `ls` and saw journal activity being recorded for them.

PR #181, opened 2019-04-02
Properly refresh the entry's attributes in FUSE

ekinanp:


plugin/cache.go
```diff
@@ -163,14 +167,56 @@ func CachedList(ctx context.Context, g Group) ([]Entry, error) {
...
+// should check the Boolean value first before checking the error.
+//
+// CachedListReadOnly is primarily needed by the FUSE layer because it is
+// optimized for frequent, concurrent reads. It should not be used anywhere
+// else.
+//
+// NOTE: If the entry's disabled caching for plugin.ListOp, then CachedListReadOnly
+// will return g.List(context.Background()) with the boolean set to true.
+// This is to optimize the FUSE layer by avoiding unnecessary lock acquisitons.
+func CachedListReadOnly(g Group) (map[string]Entry, bool, error) {
```
ekinanp: This is kind of a weird name, but the other option I considered (`GetCachedList`) also sounded weird.

Also, since this method's only needed by FUSE, would it be worth moving the FUSE package back into plugin so that this isn't included in the godoc? Just a thought.

fuse/dir.go
```diff
@@ -44,31 +44,31 @@ func (d *dir) Lookup(ctx context.Context, req *fuse.LookupRequest, resp *fuse.Lo
...
+		return nil, fuse.ENOENT
 	}
-	journal.Record(ctx, "FUSE: %v not found in %v", req.Name, d)
-	return nil, fuse.ENOENT
+
+	log.Infof("FUSE: Find[d,pid=%v] %v/%v", req.Pid, d, cname)
+	if plugin.ListAction.IsSupportedOn(entry) {
+		childdir := newDir(d.entry.(plugin.Group), entry.(plugin.Group))
+		journal.Record(ctx, "FUSE: Found directory %v", childdir)
+		// Prefetch directory entries into the cache
```
ekinanp: Do we still need to prefetch things? Would be nice to remove this logic if the answer's no so that the logs don't get cluttered up.

MikaelSmith: We could try removing it. I think it potentially makes the interface more responsive, but may not actually be a useful trade-off. Worth testing out.

ekinanp: Ok, can do that later. It's not too bad right now

PR #182, opened 2019-04-02
Remove attribute-metadata syncing logic

ekinanp:


plugin/entryBase.go
```diff
@@ -21,18 +19,11 @@ const (
...
-	*SyncableAttribute
-	key string
-}
-
 // EntryBase implements Entry, making it easy to create new entries.
 // You should use plugin.NewEntry to create new EntryBase objects.
 type EntryBase struct {
 	entryName          string
-	attrMux            *sync.RWMutex
 	attr               EntryAttributes
```
ekinanp: I considered exporting the field, but that breaks consistency with the other `Set*` methods so chose not to.

PR #183, opened 2019-04-03
Enable caching for an EC2 instance

ekinanp:
This was previously turned off because the instances' metadata.json and
console output files were stored as struct fields. However, the
attributes-metadata work moved those over to List. That work should have
re-enabled caching since entry attributes are now refreshed via a call
to the parent's CachedList.

### DIFF
--- a/plugin/aws/ec2Instance.go
+++ b/plugin/aws/ec2Instance.go
@@ -49,7 +49,7 @@ func newEC2Instance(ctx context.Context, inst *ec2Client.Instance, session *sess
 		ssmClient:        ssm.New(session),
 		cloudwatchClient: cloudwatchlogs.New(session),
 	}
-	ec2Instance.DisableDefaultCaching()
+	ec2Instance.SetTTLOf(plugin.ListOp, 30*time.Second)
 
 	metaObj := newDescribeInstanceResult(inst)
 


plugin/aws/ec2Instance.go
```diff
@@ -49,7 +49,7 @@ func newEC2Instance(ctx context.Context, inst *ec2Client.Instance, session *sess
 		ssmClient:        ssm.New(session),
 		cloudwatchClient: cloudwatchlogs.New(session),
 	}
-	ec2Instance.DisableDefaultCaching()
+	ec2Instance.SetTTLOf(plugin.ListOp, 30*time.Second)
```
ekinanp: Hmm, should we no longer cache `Open` and `Metadata` ? I think we were caching `Open` b/c we used it to compute the Size attribute, while `Metadata` was cached because we thought `wash meta` would be a frequent operation.

ekinanp: Could file an issue to re-consider the cached ops so we could look at it later.

MikaelSmith: CachedList is interesting - we work around the FUSE Node being persistent be recreating them. It makes sense to cache Open because that's called separate from Attr in the same workflow. Metadata could come directly from the node because it will have been created right before we use it.

It's inconvenient that we can't have caching work during creation.

PR #184, opened 2019-04-03
Improve debugging using activity journal

MikaelSmith:
Move to nanosecond granularity logging to debug slow operations.

Consistently print the journal ID for logged actions, and write all FUSE
activity to the journal.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -153,7 +153,8 @@ func loadLogger(levelStr string, logfile string) (*os.File, error) {
 
 	log.SetLevel(level)
 	log.SetFormatter(&log.TextFormatter{
-		FullTimestamp: true,
+		FullTimestamp:   true,
+		TimestampFormat: time.RFC3339Nano,
 	})
 
 	var logFH *os.File
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -16,6 +16,7 @@ import (
 	cmdutil "github.com/puppetlabs/wash/cmd/util"
 	"github.com/puppetlabs/wash/config"
 	"github.com/puppetlabs/wash/fuse"
+	"github.com/puppetlabs/wash/journal"
 	"github.com/puppetlabs/wash/plugin"
 	"github.com/puppetlabs/wash/plugin/aws"
 	"github.com/puppetlabs/wash/plugin/docker"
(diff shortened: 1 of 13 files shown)


plugin/aws/s3Bucket.go
```diff
@@ -217,7 +217,7 @@ func (b *s3Bucket) getRegion(ctx context.Context) (string, error) {
 	// Note that the callback to CachedOp also creates a new client for that region.
 	// We use CachedOp with a long expiration to ensure region is fetched infrequently.
 	// You can force a retry by deleting the cache entry if there was an error.
-	resp, err := plugin.CachedOp("Region", b, 24*time.Hour, func() (interface{}, error) {
+	resp, err := plugin.CachedOp(ctx, "Region", b, 24*time.Hour, func() (interface{}, error) {
```
ekinanp: I'd add a comment indicating that the `ctx` contains info. about the caller that's used for logging

MikaelSmith: Ok, will do.

cmd/server.go
```diff
@@ -68,6 +69,11 @@ func serverMain(cmd *cobra.Command, args []string) exitCode {
 		}()
 	}
 
+	// Setup finalizing steps that should be run even if we panic
+	defer func() {
+		journal.Finalize()
```
ekinanp: Nit: `Finalize` => `FlushOpenJournals` -- I like the latter better because it makes the behavior explicit in the name.

MikaelSmith: `journal.Finalize` doesn't say the same thing? Maybe `journal.FlushOpen`? `journal.FlushOpenJournals` feels a little redundant.

ekinanp: `Finalize` seemed a bit too abstract to me, like we're doing more things there than just flushing the open journals (which we're not).

`journal.FlushOpen` is fine too.

MikaelSmith: What about `journal.CloseAll`? Because it's both flushing and closing.

ekinanp: Yeah, `journal.CloseAll` sounds good.

MikaelSmith: Updated.

PR #187, opened 2019-04-05
Ensure new entries always have an ID

MikaelSmith:
For new entries created in a call to CachedList - which should be all of them - enable creating an ID as-needed by adding the parent ID to the context to be used by those cached operations. This should avoid any scenario where we can't determine a cache ID.

Also remove prefetching. It's mostly not needed and creates lots of implicit traffic that can
slow the rest of the experience. There are some scenarios where prefetching improved responsiveness, but it was equally likely to exceed rate limits on APIs and cause errors.

We could explore re-adding it with something like https://gobyexample.com/worker-pools (and maybe rate limiting) to limit concurrent requests so local responsiveness when not immediately using the prefetched data isn't impacted and to avoid running into rate limiting on APIs we use.

### DIFF
--- a/plugin/cache.go
+++ b/plugin/cache.go
@@ -8,8 +8,13 @@ import (
 	"time"
 
 	"github.com/puppetlabs/wash/datastore"
-	"github.com/puppetlabs/wash/journal"
 )
+
+// KeyType is used to create a unique key type for looking up context values.
+type keyType int
+
+// id is used to identify the parent's ID in a context.
+const parentID keyType = iota
 
 var cache datastore.Cache
 
@@ -144,7 +149,14 @@ func (c DuplicateCNameErr) Error() string {
 // querying a specific entry.
 func CachedList(ctx context.Context, g Group) (map[string]Entry, error) {
 	cachedEntries, err := cachedDefaultOp(ctx, ListOp, g, func() (interface{}, error) {
-		entries, err := g.List(ctx)
+		if g.id() == "" {
+			panic("cannot List an entry that you just created")
+		}
+
+		// Including the entry's ID allows plugin authors to use any Cached* methods defined on the
+		// children after their creation. This is necessary when the child's Cached* methods are used
+		// to calculate its attributes. Note that the child's ID is set in cachedOp.
+		entries, err := g.List(context.WithValue(ctx, parentID, g.id()))
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


plugin/cache.go
```diff
@@ -144,7 +149,13 @@ func (c DuplicateCNameErr) Error() string {
 // querying a specific entry.
 func CachedList(ctx context.Context, g Group) (map[string]Entry, error) {
 	cachedEntries, err := cachedDefaultOp(ctx, ListOp, g, func() (interface{}, error) {
-		entries, err := g.List(ctx)
+		if g.id() == "" {
+			panic("cannot List an entry that you just created")
+		}
+
+		// Include the current entry's ID in context so that operations done while listing children
```
ekinanp: I'd elaborate a bit more here, maybe something like "Including the entry's ID allows plugin authors to use any Cached* methods defined on the children _after_ their creation. This is necessary when the child's Cached* methods are used to calculate its attributes. Note that the child's ID is set in cachedOp."

Reason for ^ is because you can still cache operations while listing children, you'd just need to define those operations in the parent's type, which is very annoying.

plugin/cache.go
```diff
@@ -227,14 +238,13 @@ func cachedOp(ctx context.Context, opName string, entry Entry, ttl time.Duration
...
-		journal.Record(ctx, "Cached op %v on %v had no cache ID", opName, entry.name())
-		// This can happen if the entry's just been created, and it calls one of the Cached*
-		// methods to query its Metadata so that it can set its initial attributes. In that case,
-		// just invoke op. Ideally we should panic here, but panicking makes it significantly more
-		// annoying for plugin authors to write their code. Since the plugin package manages the
-		// entry's ID, and the entry's ID is set in CachedList, an empty ID is such a rare edge case
-		// that it's better to not annoy people.
-		return op()
+		// Try to set the ID based on parent ID
+		if obj := ctx.Value(parentID); obj != nil {
```
ekinanp: Could we centralize this? Maybe change `e.setID` => `e.setIDFromParent`.

plugin/cache.go
```diff
@@ -144,7 +149,13 @@ func (c DuplicateCNameErr) Error() string {
 // querying a specific entry.
 func CachedList(ctx context.Context, g Group) (map[string]Entry, error) {
```
ekinanp: Would also be useful to ad a comment [somewhere here](https://github.com/puppetlabs/wash/blob/master/plugin/cache.go#L169) that we're setting these IDs to ensure that all entries have an ID, even those that didn't need any of the Cached* methods in their parent's List call.

plugin/cache.go
MikaelSmith: The function would be a little awkward because we want to re-use a CName we've already calculated.

MikaelSmith:
Added comments.

PR #189, opened 2019-04-07
Add initial version of `wash find`

ekinanp:
This commit adds the initial support for `wash find`. `wash find` is
very similar to GNU find and BSD find in that it supports an expression
based language composed of primaries and operators. Primaries are
predicates constructed on a specific entry. This commit supports the -true,
-false, and -name primaries, where -true specifies a predicate that always
returns true; -false specifies a predicate that always returns false; and
-name <glob> specifies a predicate that returns true if the entry’s cname
matches the specified shell glob.

The supported operators are “()”, “!/-not”, “-a/-and”, and “-o/-or” (in order of
decreasing precedence). “-not” takes a single predicate p and returns a
new predicate q s.t. q(e) = ! p(e), where ! is the logical not operator. -and
takes predicates p1 and p2 and returns a new predicate q s.t.
q(e) = p1(e) && p2(e), where && is the short-circuited logical and operator.
Similarly, -or takes predicates p1 and p2 and returns q(e) = p1(e) || p2(e)
where || is the logical or operator. “()” is useful to control precedence. For
example, `wash find <path> -true -o -true -a -false` will evaluate to `true`
since -a has greater precedence than -o. However,
`wash find <path> ( -true -o - true ) -a -false` will evaluate to `false`.

Like the existing `find` command, `wash find` also supports concatenating predicates.
For example, something like `wash find <path> -true -false` will be
parsed as `wash find <path> -true -a -false`.

Subsequent commits will add more primaries, a better usage, and will also
properly recurse down into all the listed subdirectories.


cmd/find/namePrimary.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"fmt"
+
+	"github.com/gobwas/glob"
+	apitypes "github.com/puppetlabs/wash/api/types"
+)
+
+// namePrimary => -name ShellGlob
+var namePrimary = &atom{
```
ekinanp: The parser makes it easy to create subsequent primaries and independently test them. I plan on adding a time attribute primary to handle `ctime`, `mtime` and `atime` after this PR's merged.

cmd/find.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		cmdutil.ErrPrintf("%v\n", err)
+		return exitCode{1}
+	}
+
+	for _, entry := range entries {
+		if p(&entry) {
+			// TODO: Include the cwd's directory in path (so that find
+			// prints out absolute paths).
+			fmt.Printf("%v\n", entry.Path)
```
MikaelSmith: `find` by default prints a relative path. That seems fine to continue here.

MikaelSmith: Maybe just prefix with `.`.

cmd/find/grammar.go
```diff
@@ -0,0 +1,114 @@
...
+		// Find the ")" that's paired with our "(". Use the algorithm
+		// described in https://stackoverflow.com/questions/12752225/how-do-i-find-the-position-of-matching-parentheses-or-braces-in-a-given-piece-of
+		//
+		// TODO: Could optimize this by moving parens handling over to the evaluation
+		// stack. Not important right now because expressions to `wash find` will likely
+		// be simple.
+		tokens = tokens[1:]
+		ix := 0
+		counter := 1
+		for i, token := range tokens {
```
MikaelSmith: This looks like it won't correctly detect mismatched parentheses. You should probably error if `counter != 0` after exiting the loop.

MikaelSmith: Ah, handled elsewhere. Can you add a comment here that this is covered in the main parser?

cmd/find.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/spf13/cobra"
+)
+
+func findCommand() *cobra.Command {
+	findCmd := &cobra.Command{
+		Use: "find <path> [expression]",
+		// TODO: More detailed usage. Will need to use custom help text in order to
+		// properly enumerate all the primaries.
+		Short: "Finds stuff",
+		Args:  cobra.MinimumNArgs(1),
```
MikaelSmith: Improvements: it'd be nice to be able to say `wish find` and have that be equivalent to `wash find .`.

Also like to see it work at the plugin root level, and maybe even outside of the wash directory hierarchy.

cmd/find.go
ekinanp: Ah sure, I can do that.

cmd/find/grammar.go
```diff
@@ -0,0 +1,114 @@
...
+	notOp,
+	parensOp,
+	namePrimary,
+	truePrimary,
+	falsePrimary,
+}
+
+// Map of <token> => <atom>. This is initialized in ParsePredicate. Unfortunately,
+// we cannot intialize atoms here because doing so would result in an initialization
+// loop compiler error (e.g. notOp => atoms => notOp).
```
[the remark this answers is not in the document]
ekinanp: Yeah, it's very annoying. I think it's b/c when Go compiles notOp's `parsePredicate`, it sees a reference to `atoms` which references `notOp`, and Go just doesn't have a heuristic of "This function isn't invoked so the initialization loop here is OK" to bypass that.

cmd/find.go
ekinanp: I'll add those as TODOs

cmd/find.go
ekinanp: Oh wait, let me handle paths properly in a separate PR. The code there just prints the raw API path.

ekinanp:
Pushed an update addressing the feedback.

PR #197, opened 2019-04-09
Add `wash tail -f` to follow updates

### LINKED ISSUE #5: Write a new tail that works better
What does `tail` and `tail -f` look like when interacting directly with the APIs rather than via the filesystem.

It should work on non-wash resources as well.

MikaelSmith:
Requires `-f` as the only allowed semantic. Only displays new updates after `tail -f` is called. Works for streaming resources and normal files (using inotify).

Resolves #5.


cmd/tail.go
```diff
@@ -0,0 +1,137 @@
...
+			defer func() { errz.Log(stream.Close()) }()
+			// Start copying the stream to the aggregate channel
+			go func(src string) {
+				_, err := io.Copy(chanWriter{name: src, out: agg}, stream)
+				if err != nil {
+					agg <- line{Line: tail.Line{Time: time.Now(), Err: err}, source: src}
+				}
+			}(path)
+		} else {
+			if errObj, ok := err.(*apitypes.ErrorObj); ok && errObj.Kind == "unsupported-action" {
```
grimradical: This makes me wonder if we should return the list of supported actions when we throw an "unsupported-action" error. Another option may be to call "list" on the resource, to get the list of supported actions as a second step (only if the target isn't streamable)?

cmd/tail.go
```diff
@@ -0,0 +1,137 @@
...
+				// The resource exists but does not support the streaming action, try to read it as a file.
+				// TODO: should we go through the FS or try to use the /fs/read endpoint?
```
grimradical: Is there a practical difference between going through the FS and calling our API to get the contents? My understanding was that it was effectively the same thing.

MikaelSmith: The only one that might matter is that a directory could still implement the Readable interface, so you could read content via `/fs/read` but not through the filesystem. I don't think it's very likely to be used.

cmd/tail.go
MikaelSmith: Calling list (or metadata?) first would be reasonable, we already effectively trigger that finding the API key.

ekinanp: This seems to suggest adding something like a `/fs/info/<entry>` endpoint that would return what ListEntry returns + its full metadata. Maybe the existing `metadata` endpoint could become `/fs/info` ?

cmd/tail.go
```diff
@@ -0,0 +1,137 @@
...
+				cmdutil.ErrPrintf("%v\n", err)
+				return exitCode{1}
+			}
+		}
+	}
+
+	// Start at the end of the file
+	tailConf := tail.Config{
+		Follow:   true,
+		Location: &tail.SeekInfo{Offset: 0, Whence: 2},
```
ekinanp: What does setting the Location do functionality-wise? A comment would help here.

cmd/tail.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/spf13/viper"
+)
+
+func tailCommand() *cobra.Command {
+	tailCmd := &cobra.Command{
+		Use:   "tail -f <file> [<file>...]",
+		Short: "Displays new output of files or resources that support the stream action",
+	}
+
+	tailCmd.Flags().BoolP("follow", "f", false, "Set the output format (json or yaml)")
```
ekinanp: Copy-paste error? Follow doesn't update the output format.

cmd/tail.go
ekinanp: I think going through the fs makes sense for now.

cmd/tail.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return tailCmd
+}
+
+type line struct {
+	tail.Line
+	source string
+}
+
+type chanWriter struct {
```
ekinanp: Nit: `lineWriter` reads a bit better.

cmd/tail.go
```diff
@@ -0,0 +1,137 @@
...
+			files = append(files, path)
+			continue
+		}
+
+		if stream, err := conn.Stream(apiPath); err == nil {
+			defer func() { errz.Log(stream.Close()) }()
+			// Start copying the stream to the aggregate channel
+			go func(src string) {
+				_, err := io.Copy(chanWriter{name: src, out: agg}, stream)
+				if err != nil {
```
ekinanp: I don't see this error being handled anywhere

cmd/tail.go
MikaelSmith: That's exactly what the comment says. I guess I can clarify it to relate more clearly to Location.

ekinanp: Oh yeah, brain fart on my part. Maybe a bit more clarification would help in case other people are as mentally lazy as me.

cmd/tail.go
MikaelSmith: I don't think these are necessary changes right now, the error returned seems clear enough.

MikaelSmith:
Updated. Also added a little refactoring.

cmd/clear.go
```diff
@@ -36,14 +36,14 @@ func clearMain(cmd *cobra.Command, args []string) exitCode {
 
 	apiPath, err := client.APIKeyFromPath(path)
 	if err != nil {
-		cmdutil.ErrPrintf("%v\n", err)
+		log.Print(err)
```
ekinanp: This is weird to me because `log.Print` is generic but here we're only using it to print errors. Could we handle colored output separately? Or maybe change `log.Print` => `log.PrintError` ? Would prefer separately b/c it seems like there may be some design work there.

cmd/tail.go
```diff
@@ -0,0 +1,154 @@
...
+	tailConf := tail.Config{
+		Follow:   true,
+		Location: &tail.SeekInfo{Offset: 0, Whence: 2},
+		Logger:   tail.DiscardingLogger,
+	}
+	for _, path := range files {
+		if finfo, err := os.Stat(path); err != nil {
+			log.Print(err)
+			continue
+		} else if finfo.IsDir() {
```
ekinanp: A comment explaining why we're doing this check would be useful, b/c `tail` working on a directory is unintuitive.

ekinanp: Oh, I was talking about commenting that the `tail` package works on directories -- that was the unintuitive part and the reason for the `IsDir()` check.

MikaelSmith: The comment at line 107 doesn't clarify that?

MikaelSmith: Oh I see. I'll specifically mention the `tail` package behavior.

MikaelSmith: It might actually be waiting for the file to exist. I wonder if I should still call `TailFile` on these.

ekinanp: Up to you. What you have here seems OK to me.

PR #198, opened 2019-04-10
 Add primaries for the ctime, mtime, and atime attributes

ekinanp:

### DIFF
--- a/cmd/find/grammar.go
+++ b/cmd/find/grammar.go
@@ -16,6 +16,8 @@ type atom struct {
 // Map of <token> => <atom>.
 var atoms = make(map[string]*atom)
 
+// When creating a new atom with this function, be sure to comment nolint above the variable
+// so that CI does not mark it as unused. See notOp for an example.
 func newAtom(tokens []string, parsePredicate func(tokens []string) (Predicate, []string, error)) *atom {
 	a := &atom{
 		tokens:         tokens,
@@ -27,6 +29,7 @@ func newAtom(tokens []string, parsePredicate func(tokens []string) (Predicate, [
 	return a
 }
 
+//nolint
 var notOp = newAtom([]string{"!", "-not"}, func(tokens []string) (Predicate, []string, error) {
 	notToken := tokens[0]
 	tokens = tokens[1:]
@@ -46,6 +49,7 @@ var notOp = newAtom([]string{"!", "-not"}, func(tokens []string) (Predicate, []s
 	return nil, nil, fmt.Errorf("%v: no following expression", notToken)
 })
 
+//nolint
 var parensOp = newAtom([]string{"("}, func(tokens []string) (Predicate, []string, error) {
 	// Find the ")" that's paired with our "(". Use the algorithm
 	// described in https://stackoverflow.com/questions/12752225/how-do-i-find-the-position-of-matching-parentheses-or-braces-in-a-given-piece-of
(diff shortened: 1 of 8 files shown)


cmd/find/grammar.go
```diff
@@ -13,104 +13,98 @@ type atom struct {
...
-	parensOp,
-	namePrimary,
-	truePrimary,
-	falsePrimary,
-}
-
-// Map of <token> => <atom>. This is initialized in ParsePredicate. Unfortunately,
-// we cannot intialize atoms here because doing so would result in an initialization
-// loop compiler error (e.g. notOp => atoms => notOp).
+// Map of <token> => <atom>.
```
MikaelSmith: I think a comment that this is populated by calls to `newAtom` would be helpful.

cmd/find/grammar.go
```diff
@@ -13,104 +13,98 @@ type atom struct {
...
 	precedence int
 	combine    func(p1 Predicate, p2 Predicate) Predicate
 }
 
-var allBinaryOps = []*binaryOp{
-	andOp,
-	orOp,
-}
-
 // Map of <token> => <binaryOp>
```
MikaelSmith: Same comment here for newBinaryOp.

cmd/find/timeAttrPrimary.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	return duration, roundDiff
+}
+
+// timeAttrPrimary => -<name> (+|-)?((\d+)|(\d+[smhdw])
+//
+// where s => seconds, m => minutes, h => hours, d => days, w => weeks
+//
+// Example inputs:
+//   -mtime 1      (true if the difference between the entry's mtime and startTime is exactly 1 24-hour period)
```
MikaelSmith: Is there any rounding on that? Alternatively, is there any way to express "between 23 and 24 hours"? I guess that could be `-mtime +23h -and -mtime -24h`.

cmd/find/timeAttrPrimary.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+	return duration, roundDiff
+}
+
+// timeAttrPrimary => -<name> (+|-)?((\d+)|(\d+[smhdw])
+//
+// where s => seconds, m => minutes, h => hours, d => days, w => weeks
+//
+// Example inputs:
```
MikaelSmith: How can we document all these primaries? Starting to need a man-page :)

cmd/find/timeAttrPrimary.go
ekinanp: Yeah, that seems to be the only way to do it. We could consider adding alternate syntax for these primaries that supports something like that? Wouldn't be hard to do if people find `and`'ing the `mtimes` annoying.

cmd/find/timeAttrPrimary.go
ekinanp: Either a really long usage, or a man page. Could go with a long usage for now, then the man page?

I'd like to handle the usage separately though, at least after the `Size` primary's been added (and maybe after the full traversal logic [i.e. `depth`])

cmd/find/timeAttrPrimary.go
MikaelSmith: I think the way I outlined is fine. Not sure exact numbers will be that useful in practice.

cmd/find/timeAttrPrimary.go
MikaelSmith: Sure.

PR #201, opened 2019-04-10
Fix ps on some AWS instances

MikaelSmith:
For instances using the `dash` shell in EC2, shell escaping didn't work
right for tabs, newlines, and null characters as they'd be replaced by
their literal values after echoing. Escape those characters.

Some processes also have an empty cmdline file. Handle those cases.

### DIFF
--- a/plugin/aws/ec2Instance.go
+++ b/plugin/aws/ec2Instance.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


plugin/aws/ec2Instance.go
```diff
@@ -270,7 +271,12 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
 
 		// Handle Stdin by piping its content into our command.
 		// Note that this won't work well for large Stdin streams.
-		cmdStr = shellquote.Join("echo", "-n", string(input)) + " | " + cmdStr
+		quoted := shellquote.Join(string(input))
+		// Add extra escaping needed for some shells. This should probably be part of shellquote.
+		quoted = strings.Replace(quoted, "\\t", "\\\\t", -1)
```
ekinanp: Agreed. Too bad that repo isn't very well-maintained.

PR #202, opened 2019-04-11
Add the size primary to `wash find`

ekinanp:
The size primary was modeled off its counterpart in BSD's `find`
command.

### DIFF
--- a/cmd/find/sizePrimary.go
+++ b/cmd/find/sizePrimary.go
@@ -30,13 +30,13 @@ var sizeValueRegex = regexp.MustCompile(`^(\+|-)?((\d+)|(\d+[ckMGTP]))$`)
 
 // sizePrimary => -size (+|-)?((\d+)|(\d+[ckMGTP]))
 //
-// where c => character (byte), k => kibibyte, M => megabyte, G => gigabyte, T => terabyte, P => petabyte
+// where c => character (byte), k => kibibyte, M => mebibyte, G => gibibyte, T => tebibyte, P => pebibyte
 //
 // Example inputs:
 //   -size 2   (true if the entry's size in 512-byte blocks, rounded up, is 2)
 //   -size +2  (true if the entry's size in 512-byte blocks, rounded up, is greater than 2)
 //   -size -2  (true if the entry's size in 512-byte blocks, rounded up, is less than 2)
-//   -size +1k (true if the entry's size is greater than 1 kilobyte (1024 bytes))
+//   -size +1k (true if the entry's size is greater than 1 kibibyte (1024 bytes))
 //
 //nolint
 var sizePrimary = newAtom([]string{"-size"}, func(tokens []string) (Predicate, []string, error) {
--- a/cmd/find/sizePrimary.go
+++ b/cmd/find/sizePrimary.go
@@ -30,7 +30,7 @@ var sizeValueRegex = regexp.MustCompile(`^(\+|-)?((\d+)|(\d+[ckMGTP]))$`)
 
 // sizePrimary => -size (+|-)?((\d+)|(\d+[ckMGTP]))
 //
-// where c => character (byte), k => kilobyte, M => megabyte, G => gigabyte, T => terabyte, P => petabyte
+// where c => character (byte), k => kibibyte, M => megabyte, G => gigabyte, T => terabyte, P => petabyte
 //
 // Example inputs:
 //   -size 2   (true if the entry's size in 512-byte blocks, rounded up, is 2)
--- a/cmd/find/sizePrimary.go
+++ b/cmd/find/sizePrimary.go
@@ -63,19 +63,6 @@ var sizePrimary = newAtom([]string{"-size"}, func(tokens []string) (Predicate, [
 
 	p := func(e *apitypes.ListEntry) bool {
 		if !e.Attributes.HasSize() {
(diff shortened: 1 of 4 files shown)


cmd/find/sizePrimary.go
```diff
@@ -0,0 +1,112 @@
...
+			// b/c the output will contain things that don't have a `size` attribute
+			// (like EC2 instances). When we add the `state` attribute, then an expression
+			// like `wash find /aws -state running -o true` will include entries that don't
+			// have state, like S3 buckets and S3 objects.
+			//
+			// The more general issue here is how we handle entries that do not have the attributes
+			// required by the user's specified primaries when those primaries are or'ed
+			// with other primaries that could potentially return true for that entry. Do
+			// we ignore them? Do we still apply the predicate on them, even if that predicate
+			// may return true? We should figure this out soon.
```
ekinanp: I'm thinking it'd be a better UX if `wash find`'s predicate always returned false for entries that do not have the attribute(s) specified by the given expression (so something like `wash find /aws -size +1 -o true` will never display EC2 instances because size doesn't make sense for them). 

We don't have to enforce this for _all_ attributes, e.g. I think entries without `ctime`, `mtime` or `atime` attributes should still be processed. Hmm, maybe we shouldn't enforce it for `ctime`, but we should for the other attributes?

If the TODO here is not an issue or is something we should address if users become annoyed by it, I'm happy to remove it too.

cmd/find/sizePrimary.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+var sizeValueRegex = regexp.MustCompile(`^(\+|-)?((\d+)|(\d+[ckMGTP]))$`)
+
+// sizePrimary => -size (+|-)?((\d+)|(\d+[ckMGTP]))
+//
+// where c => character (byte), k => kilobyte, M => megabyte, G => gigabyte, T => terabyte, P => petabyte
+//
+// Example inputs:
+//   -size 2   (true if the entry's size in 512-byte blocks, rounded up, is 2)
```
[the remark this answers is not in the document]
MikaelSmith: Also, why 512-byte blocks? 4096 is a more common default for Linux.

cmd/find/sizePrimary.go
MikaelSmith: Why would you use `-o true`? I'm not clear on that.

ekinanp: Oh, the real use-case could be something llike `find /aws -size +1 -o mtime +1h` where an entry without a size attribute but w/ an mtime attribute could satisfy the predicate

cmd/find/sizePrimary.go
```diff
@@ -0,0 +1,112 @@
...
+		return nil, nil, fmt.Errorf("-size: requires additional arguments")
+	}
+	v := tokens[0]
+	if !sizeValueRegex.MatchString(v) {
+		return nil, nil, fmt.Errorf("-size: %v: illegal size value", v)
+	}
+
+	cmp := v[0]
+	if cmp == '+' || cmp == '-' {
+		v = strings.TrimPrefix(v, string(cmp))
```
MikaelSmith: So this allows `+++1`? Any reason to do that rather than just a string slice `v[1:]`?

cmd/find/sizePrimary.go
MikaelSmith: I'm comfortable saying that predicate returns false, but not going any further. We can do something more complicated if specific examples come up.

cmd/find/sizePrimary.go
ekinanp: It doesn't -- `sizeValueRegex` already validated v. That's a good test-case to add though.

Ah yeah, string slice is simpler. Will update.

cmd/find/sizePrimary.go
ekinanp: Sounds good, I'll remove the TODO.

cmd/find/sizePrimary.go
MikaelSmith: Did string slicing work?

ekinanp:
@MikaelSmith Updated the PR (and yes, string slicing worked). Only unaddressed feedback is whether we should use 512/4096 byte blocks as the default. Note that the Linux `find` command also defaults to 512 byte blocks.

cmd/find/sizePrimary.go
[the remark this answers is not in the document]
MikaelSmith: I think you missed the table on the right from my link. mebibyte, gibibyte, tebibyte, pebibyte.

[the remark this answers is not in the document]
MikaelSmith: Also line 39 isn't correct.

MikaelSmith:
Leaving it 512 is fine.

cmd/find/sizePrimary.go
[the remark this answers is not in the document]
ekinanp: Oh, you wanted all of them to be labeled as such. Sure, will update.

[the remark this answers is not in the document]
MikaelSmith: I want to follow the ISQ standard. I'm ok with the single-character form referring to the colloquial kilobyte (now kibibyte), etc. We could later differentiate with `KB` and `KiB`, which are the official IEC symbols.

PR #205, opened 2019-04-11
Simplify tail implementation

MikaelSmith:
Cleanup tail implementation to be more readable and attempt to open each
file in the order they're listed. Extracts attempting to tail streaming
and file entries to functions.

### DIFF
--- a/cmd/tail.go
+++ b/cmd/tail.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


cmd/tail.go
```diff
@@ -49,6 +50,80 @@ func (w lineWriter) Write(b []byte) (int, error) {
 	return len(b), nil
 }
 
+// Streams output via API to aggregator channel.
+// Returns nil if streaming's not supported on this path.
+func tailStream(conn client.DomainSocketClient, agg chan line, path string) io.Closer {
```
ekinanp: Hmm, why not error the tail command if any of these tail functions fail to open the resource?

MikaelSmith: It’s a little weird. Trying to come up with same semantics for: success, failed but you should try it as a file, and failed and you should skip it. Maybe that third doesn’t need to exist?

ekinanp: Yeah, I ended up typing close to a paragraph trying to come up with a more "simplified" set of semantics so I think skipping the file is OK.

PR #206, opened 2019-04-11
Improve wash ls subcommand

### LINKED ISSUE #204: Improve LS
Make it work everywhere rather than panicking. Output can be different based on context.

Make `wash list` a command, with `wash ls` aliased to it (for consistency with the API).

Add common attributes to LS:
- user who created the resource
- location the resource lives

MikaelSmith:
Renames the `wash ls` command to `wash list` to make it more consistent with the API. Also includes `wash ls` as an alias to `wash list` to preserve familiarity with the `ls` command.

Fixes listing
- files
- the wash root directory
- wash plugin roots

Also simplifies code around listing the current directory, and introduces a foundation for listing non-resource paths.

Updates `wash list` to work for normal files and directories. Emits similar table content to that used for wash resources.

Resolves #204.

### DIFF
--- a/api/helpers.go
+++ b/api/helpers.go
@@ -53,6 +53,13 @@ func getEntryFromPath(ctx context.Context, path string) (plugin.Entry, *errorRes
 		panic("path should never be empty")
 	}
 
+	// Get the registry from context (added by registry middleware).
+	registry := ctx.Value(pluginRegistryKey).(*plugin.Registry)
+	if path == "/" {
+		// Return the registry
+		return registry, nil
+	}
+
 	// Don't interpret trailing slash as a new segment, and ignore optional leading slash
 	path = strings.Trim(path, "/")
 	// Split into plugin name and an optional list of segments.
@@ -60,11 +67,13 @@ func getEntryFromPath(ctx context.Context, path string) (plugin.Entry, *errorRes
 	pluginName := segments[0]
 	segments = segments[1:]
 
-	// Get the registry from context (added by registry middleware).
-	registry := ctx.Value(pluginRegistryKey).(*plugin.Registry)
 	root, ok := registry.Plugins()[pluginName]
 	if !ok {
 		return nil, pluginDoesNotExistResponse(pluginName)
+	}
+	if len(segments) == 0 {
+		// Listing the plugin itself, so return it's root
+		return root, nil
 	}
 
 	return findEntry(ctx, root, segments)
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -1,12 +1,13 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
 
+	"github.com/Benchkram/errz"
 	"github.com/spf13/cobra"
 
 	"github.com/puppetlabs/wash/api/client"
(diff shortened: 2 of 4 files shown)


api/helpers.go
```diff
@@ -53,19 +53,28 @@ func getEntryFromPath(ctx context.Context, path string) (plugin.Entry, *errorRes
...
 	// Split into plugin name and an optional list of segments.
 	segments := strings.Split(path, "/")
 	pluginName := segments[0]
 	segments = segments[1:]
 
-	// Get the registry from context (added by registry middleware).
-	registry := ctx.Value(pluginRegistryKey).(*plugin.Registry)
 	root, ok := registry.Plugins()[pluginName]
 	if !ok {
 		return nil, pluginDoesNotExistResponse(pluginName)
```
ekinanp: Hmm, how important is it to differentiate a non-existent plugin from a non-existent entry? If it's not that important, we could probably move this code over to `findEntry`.

cmd/list.go
```diff
@@ -0,0 +1,180 @@
...
+		} else {
+			entries = []apitypes.ListEntry{target}
+		}
+	}
+
+	fmt.Print(formatListEntries(apiPath, entries))
+	return nil
+}
+
+func listPath(path string) error {
```
ekinanp: One thing I was thinking about is to make a core plugin that we can delegate to for "non-Wash" paths (this core-plugin's internal and is not something we expose to the user, but we include it in the plugin.Registry). This way, we can maintain consistency between the outputs in `wash ls` for Wash entries vs. non-Wash entries.

FS-specific things like inode number can be included in the metadata so that `wash find` could still query regular files via the `meta` primary.

What do you think? Asking b/c having commands with two different outputs based on context is a little weird. There's no reason why regular files cannot be interpreted as Wash resources (especially when we have an `sshfs` example).

cmd/list.go
```diff
@@ -0,0 +1,180 @@
...
+	apitypes "github.com/puppetlabs/wash/api/types"
+	cmdutil "github.com/puppetlabs/wash/cmd/util"
+	"github.com/puppetlabs/wash/config"
+	"github.com/puppetlabs/wash/plugin"
+)
+
+func listCommand() *cobra.Command {
+	listCmd := &cobra.Command{
+		Use:     "list [file]",
+		Aliases: []string{"ls"},
```
ekinanp: Cool, didn't know Cobra had this.

cmd/list.go
```diff
@@ -0,0 +1,180 @@
...
+	if apiPath == "/" {
+		// The root, definitely listable
+		ls, err := conn.List(apiPath)
+		if err != nil {
+			return err
+		}
+		entries = ls
+	} else {
+		// List the parent to see whether it's a single entry or a listable resource
+		parent, base := filepath.Split(apiPath)
```
ekinanp: Think this might be a good time to implement that `fs/info/<entry>` endpoint, because the casing here's a bit weird.

ekinanp: I'm starting to think that'd make the `List` endpoint cleaner because we can avoid returning `.` in the results. That way, people can see that `List` is strictly meant to invoke the `List` action.

cmd/list.go
ekinanp: Hmm, we could maybe make it an exported variable instead of adding it to the plugin.Registry.

ekinanp: This "plugin" could probably just be in the API, b/c it's needed for the API to work on regular files. That also means our endpoints need to take absolute paths.

api/helpers.go
MikaelSmith: I think it's useful to leave it differentiated.

cmd/list.go
MikaelSmith: I think the endpoints may need to take absolute paths anyway to support Windows. I'd prefer to keep that a separate refactor, as I think it threads through a lot more of the system.

cmd/list.go
MikaelSmith: That's fair, I think it would be useful.

cmd/list.go
ekinanp: Sounds good.

PR #208, opened 2019-04-12
Add the fs/info endpoint

ekinanp:


api/info.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/puppetlabs/wash/journal"
+)
+
+// TODO: Include the entry's full metadata?
```
MikaelSmith: 🤷‍♂️ that seems a little weird when we also have `/fs/meta`. Would definitely leave it out for now.

PR #209, opened 2019-04-12
Require absolute paths for the Wash API

ekinanp:
This removes the dependency on xattr for determining the API key. It
also paves the way for the API to support regular files/directories
(i.e. non-Wash entries) in subsequent commits.

### DIFF
--- a/api/list.go
+++ b/api/list.go
@@ -53,7 +53,12 @@ var listHandler handler = func(w http.ResponseWriter, r *http.Request, p params)
 
 	result := make([]apitypes.Entry, 0, len(entries))
 	for _, entry := range entries {
-		result = append(result, toAPIEntry(entry))
+		apiEntry := toAPIEntry(entry)
+		// Use += for efficiency
+		apiEntry.Path = path
+		apiEntry.Path += "/"
+		apiEntry.Path += apiEntry.CName
+		result = append(result, apiEntry)
 	}
 	journal.Record(ctx, "API: List %v %+v", path, result)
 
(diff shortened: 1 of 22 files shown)


cmd/find.go
```diff
@@ -49,21 +49,16 @@ func findMain(cmd *cobra.Command, args []string) exitCode {
...
-	e, err := conn.Info(apiPath)
+	e, err := conn.Info(path)
 	if err != nil {
 		cmdutil.ErrPrintf("%v\n", err)
 		return exitCode{1}
 	}
 	entries := []apitypes.Entry{e}
 	if e.Supports(plugin.ListAction) {
-		children, err := conn.List(apiPath)
+		children, err := conn.List(path)
```
ekinanp: One nice thing about this PR is that `wash find` now prints absolute paths.

MikaelSmith: It shouldn't though. In what circumstance does it make sense to print absolute paths?

ekinanp: Nevermind, brainfart on my part.

ekinanp: Fixed.

ekinanp:
We'll need `apitypes.Entry#Path` for `wash find`, specifically for the `path` primary and to make querying children easier w/o having to calculate and use the absolute path. Thus, I've kept it for now.

PR's been updated to only fix `wash find`.

api/list.go
```diff
@@ -53,7 +53,12 @@ var listHandler handler = func(w http.ResponseWriter, r *http.Request, p params)
 
 	result := make([]apitypes.Entry, 0, len(entries))
 	for _, entry := range entries {
-		result = append(result, toAPIEntry(entry))
+		apiEntry := toAPIEntry(entry)
+		// Use += for efficiency
```
MikaelSmith: What makes this more efficient? I'd have just gone `apiEntry.Path = filepath.Join(path, apiEntry.CName)`.

MikaelSmith: But I guess that'd be wrong, because on Windows that would add a backslash.

ekinanp: Thanks, that is much better. Updated.

PR #210, opened 2019-04-12
History

MikaelSmith:
Add a `history` command to view all commands that have operated through
wash since the wash server was started. `history <n>` can be used to
view logs related to that specific entry in history, where `<n>` is the
index shown in `history`.

In practice this looks like
```
$ wash history
 1  2019-04-15 12:14  /System/Library/CoreServices/coreservicesd
 2  2019-04-15 12:14  /System/Library/CoreServices/Finder.app/Contents/MacOS/Finder
 3  2019-04-15 12:14  /Applications/Dropbox.app/Contents/MacOS/Dropbox
 4  2019-04-15 12:14  git config --get oh-my-zsh.hide-status
 5  2019-04-15 12:14  git rev-parse --is-inside-work-tree
 6  2019-04-15 12:14  ls -pG
 7  2019-04-15 12:14  git config --get oh-my-zsh.hide-status
 8  2019-04-15 12:14  git rev-parse --is-inside-work-tree
 9  2019-04-15 12:14  -zsh
10  2019-04-15 12:15  wash meta aws
11  2019-04-15 12:15  git config --get oh-my-zsh.hide-status
12  2019-04-15 12:15  git rev-parse --is-inside-work-tree
13  2019-04-15 12:15  wash history
$ wash history 12
time="2019-04-15T12:15:02.170225-07:00" level=info msg="FUSE: Attr /"
time="2019-04-15T12:15:02.170344-07:00" level=info msg="FUSE: Attr finished /"
time="2019-04-15T12:15:02.170795-07:00" level=info msg="FUSE: Find .git in /"
time="2019-04-15T12:15:02.170938-07:00" level=info msg="FUSE: .git not found in /"
time="2019-04-15T12:15:02.171102-07:00" level=info msg="FUSE: Find .git in /"
time="2019-04-15T12:15:02.171206-07:00" level=info msg="FUSE: .git not found in /"
time="2019-04-15T12:15:02.171303-07:00" level=info msg="FUSE: Find HEAD in /"
time="2019-04-15T12:15:02.171357-07:00" level=info msg="FUSE: HEAD not found in /"
$ wash history 11
time="2019-04-15T12:15:02.147485-07:00" level=info msg="FUSE: Attr /"
time="2019-04-15T12:15:02.147572-07:00" level=info msg="FUSE: Attr finished /"
time="2019-04-15T12:15:02.147956-07:00" level=info msg="FUSE: Find .git in /"
time="2019-04-15T12:15:02.14803-07:00" level=info msg="FUSE: .git not found in /"
time="2019-04-15T12:15:02.148139-07:00" level=info msg="FUSE: Find .git in /"
time="2019-04-15T12:15:02.148196-07:00" level=info msg="FUSE: .git not found in /"
time="2019-04-15T12:15:02.148297-07:00" level=info msg="FUSE: Find HEAD in /"
time="2019-04-15T12:15:02.148364-07:00" level=info msg="FUSE: HEAD not found in /"
```


journal/history.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// History returns the entire history.
+func History() []Entry {
+	// We only ever add to history, so the slice wont be modified by other operations later.
+	return history
+}
+
+// Open returns a reader to read the journal specified by its ID.
+func Open(journalID string) (io.ReadCloser, error) {
```
ekinanp: Is thread-safety a concern here? For example, would it be bad if we open a journal that's also being written to?

journal/history.go
```diff
@@ -0,0 +1,63 @@
...
+	"path/filepath"
+	"sync"
+	"time"
+)
+
+var mux = sync.Mutex{}
+
+// Entry contains the command-line used to invoke the process associated with this item
+// and its journal ID.
+type Entry struct {
```
ekinanp: Hmm, could we include this information in the filename + journal file? I'm thinking something like `<server_pid>-<process_pid>-<start_time>-<executable>` for the filename, and then the first line of the journal content could be `<description>`. The remaining content would be the activity.

api/history.go
```diff
@@ -0,0 +1,66 @@
...
+			commands[i] = item.Start.Format("2006-01-02 15:04") + "  " + item.Description
+		}
+		jsonEncoder := json.NewEncoder(w)
+		if err := jsonEncoder.Encode(&commands); err != nil {
+			log.Warnf("Unable to encode history: %v", err)
+			return unknownErrorResponse(fmt.Errorf("Could not marshal %v: %v", history, err))
+		}
+	} else {
+		idx, err := strconv.Atoi(index)
+		if err != nil || idx < 0 || idx >= len(history) {
```
ekinanp: Shouldn't this be based off of PID? Idx seems like something that'd be hard to track for a user.

journal/history.go
```diff
@@ -0,0 +1,63 @@
...
+	ID, Description string
+	Start           time.Time
+}
+
+// A list of commands that have been invoked
+var history = make([]Entry, 0)
+
+// Lookup to identify whether a journal ID has already been recorded in history. There's a small
+// chance the journal ID may not be unique in history. We handle that by ignoring any entries we
```
ekinanp: When would it not be unique?

api/history.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/gorilla/mux"
+	apitypes "github.com/puppetlabs/wash/api/types"
+	"github.com/puppetlabs/wash/journal"
+	log "github.com/sirupsen/logrus"
+)
+
+var historyHandler handler = func(w http.ResponseWriter, r *http.Request) *errorResponse {
+	history := journal.History()
+
+	if index := mux.Vars(r)["index"]; index == "" {
```
ekinanp: Hmm, may be nice to split this into `historyHandler`/`historyEntryHandler`, respectively.

journal/history.go
```diff
@@ -0,0 +1,63 @@
...
+// find that are more than an hour old.
+var jidToHistory = make(map[string]time.Time)
+
+// registerCommand appends the command description to history if it hasn't been registered before.
+func (e Entry) registerCommand() {
```
ekinanp: Hmm, could we move

journal/helper.go
```diff
@@ -2,31 +2,48 @@ package journal
...
 // lots of new processes being created constantly, but makes fast things a *lot* faster.
 var pidToExec = datastore.NewMemCache()
 
-// PIDToID converts a process ID to a journal ID, appending the name of the executable
-// running that process if we can retrieve it.
-func PIDToID(pid int) string {
-	pidStr := strconv.FormatInt(int64(pid), 10)
+// PIDToJournalEntry converts a process ID to a journal ID, appending the name of the executable
+// running that process if we can retrieve it - and the command invokation of that process ID.
+func PIDToJournalEntry(pid int) Entry {
```
ekinanp: Hmm, could we move this over to https://github.com/puppetlabs/wash/blob/master/journal/core.go#L89, where we trigger this code _if_ the journal file doesn't exist (implying that it's being created for the first time)? We could add it to the history there + maintain a hash to handle the case where the journal file's deleted.

journal/history.go
MikaelSmith: OS's use a limited number for process IDs (I think 100k on macOS). If you somehow got the same process ID for the same executable, you'd have the same journal ID. I just realized I could include create_time (from https://github.com/shirou/gopsutil) in the journal ID to make it unique. Using `time.Now()` when generating the journal ID didn't seem right because I want it to be reproducible for the same process.

api/server.go
```diff
@@ -63,6 +53,28 @@ func (handle handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+type handler func(http.ResponseWriter, *http.Request) *errorResponse
+
+func (handle handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	log.Infof("API: %v %v", r.Method, r.URL)
+	wrapError(w, handle(w, r))
+}
+
+type entryHandler func(http.ResponseWriter, *http.Request, params) *errorResponse
+
+func (handle entryHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
MikaelSmith: So have a `params` argument, then pass that directly to `getEntryFromPath` and error if `path` isn't included?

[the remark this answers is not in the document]
ekinanp: Yeah, something like that. Could also just pass-in the request and extract the `path` param that way. `getEntryFromRequest` doesn't read too badly, and it makes sense.

journal/history.go
ekinanp: Ah, I see. A comment describing what you just said would be useful.

Good to see there's an easy way to make things unique.

journal/history.go
MikaelSmith: There shouldn't be. Writing will happen at the end. Worst-case we get a partially written line at the end.

journal/helper.go
MikaelSmith: The thing we add to context is an opaque blob everywhere else.

I'm not sure what it is you're suggesting be moved.

MikaelSmith:
I still want to add some tests, but updated to address some of the feedback.

journal/history.go
MikaelSmith: Made journal ID unique by including the timestamp.

README.md
```diff
@@ -49,6 +49,7 @@ We've implemented some neat features inside of `wash` to support the above goals
     * `wash find` - find resources using powerful selection predicates (WIP)
     * `wash tail -f` - follow updates to resources that support the `stream` primitive as well as normal files
     * `wash ps` - lists running processes on indicated compute instances that support the `exec` primitive
+    * `wash history` - lists all activity through `wash`; `wash history <id>` can be used to view logs for a specific entry
```
ekinanp: Nit: entry => activity (just so people don't confuse it with Wash entries)

activity/core.go
```diff
@@ -42,38 +42,39 @@ func CloseAll() {
...
-// Record writes a new entry to the journal identified by the ID at `journal.Key` in
+// Record writes a new entry to the journal identified by the ID at `activity.JournalKey` in
 // the provided context. It also writes to the server logs at the debug level. If no ID
-// is registered, the entry is written to the server logs at the warning level. If the
+// is registered, the entry is written to the server logs at the info level. If the
 // ID is an empty string, it uses the ID 'dead-letter-office'.
 //
 // Record creates a new journal for ID if needed, then appends the message to that journal.
 // Records are journaled in the user's cache directory under `wash/activity/ID.log`.
 func Record(ctx context.Context, msg string, a ...interface{}) {
```
ekinanp: I'd say the logic in https://github.com/MikaelSmith/wash/blob/history/activity/core.go#L83-L107 should be moved to a `Journal#Record` method. Otherwise, it's weird having an `activity.Journal` type but no way for that type to access its own file without having to go through `activity.Record`.

ekinanp: `journalCache` should also be `journalFileCache` or something similar, because it's not caching `Journal` objects but journal files.

activity/helper.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/puppetlabs/wash/datastore"
+	"github.com/shirou/gopsutil/process"
+	log "github.com/sirupsen/logrus"
+)
+
+// Cache pid to process names. This may end up getting the wrong process name if there are
+// lots of new processes being created constantly, but makes fast things a *lot* faster.
+var pidToExec = datastore.NewMemCache()
+
+// JournalForPID converts a process ID to a journal ID, and the command invocation of that
```
ekinanp: This comment should probably be re-worded. Maybe something like "JournalForPID creates a journal that can be used to record all Wash-related activity induced by the given process ID. The journal ID is formatted as `<pid>-<name>-<createtime>`" (I like explicitly specifying the format better because it's more concise)

activity/helper.go
```diff
@@ -0,0 +1,56 @@
...
+	"time"
+
+	"github.com/puppetlabs/wash/datastore"
+	"github.com/shirou/gopsutil/process"
+	log "github.com/sirupsen/logrus"
+)
+
+// Cache pid to process names. This may end up getting the wrong process name if there are
+// lots of new processes being created constantly, but makes fast things a *lot* faster.
+var pidToExec = datastore.NewMemCache()
```
ekinanp: This should be renamed to something like `pidJournalCache`

activity/journal.go
```diff
@@ -0,0 +1,67 @@
...
+func initHistory() ([]Journal, map[string]struct{}) {
+	return make([]Journal, 0), make(map[string]struct{})
+}
+
+// history: a list of commands that have been invoked
+// jidToHistory: lookup to identify whether a journal ID has already been recorded in history.
+var history, jidToHistory = initHistory()
+
+// registerCommand appends the command description to history if it hasn't been registered before.
+func (j Journal) registerCommand() {
```
ekinanp: `registerCommand` => `addToHistory` -- the latter better captures the generic nature of these journals since they do not have to be limited to just commands.

activity/journal.go
```diff
@@ -0,0 +1,67 @@
...
+
+var mux = sync.Mutex{}
+
+func initHistory() ([]Journal, map[string]struct{}) {
+	return make([]Journal, 0), make(map[string]struct{})
+}
+
+// history: a list of commands that have been invoked
+// jidToHistory: lookup to identify whether a journal ID has already been recorded in history.
+var history, jidToHistory = initHistory()
```
ekinanp: `jidToHistory` => `storedJournals` -- latter better reflects how this variable is used since `struct{}` is not a history object.

activity/journal.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"io"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+)
+
+// Journal contains the command-line used to invoke the process associated with this item
```
ekinanp: Based on our conversation yesterday, I'd reword this comment to reflect that these journals are more generic than the "command-line" + "process" level.

activity/journal.go
```diff
@@ -0,0 +1,67 @@
...
+// and its journal ID.
+type Journal struct {
+	ID, Description string
+	Start           time.Time
```
ekinanp: Based on how `Start` is used in the code, I'm guessing that it represents when the activity was first started right? If so, then I don't think it belongs in the `Journal` type because the `Journal` type's meant to be re-created multiple times, while `Start` is not supposed to change (this is especially true in `JournalForPID`, where API clients that use it will find that history will record a different Start time than the one specified in `JournalForPID`.).

I think it's worth creating a separate `historyEntry` type to encapsulate the `Start` time info, and have `Start` time be set when we're adding it to the history.

We could also add it to the history when the journal's first created (and maybe create the journal file along the way so that `history <ix>` won't return an error?).

api/history.go
```diff
@@ -0,0 +1,102 @@
...
+//     - application/octet-stream
+//
+//     Schemes: http
+//
+//     Responses:
+//       200: octetResponse
+//       400: errorResp
+//       404: errorResp
+//       500: errorResp
+var journalHandler handler = func(w http.ResponseWriter, r *http.Request) *errorResponse {
```
ekinanp: Hmm, I still like `historyEntryHandler` because it's more RESTy (from what I've seen, GET `/<collection>` is an array of items, and GET `/<collection>/<item_id>` returns more detailed info for a single item, while here, we go from thinking about history entries (collection) to a journal (item), which is weird).

I think the response for `/history/<ix>` could be JSON, maybe something like
```
startTime
journal: {
  id: <id>
  description: <description>
  path: <journal file path>
  content: <content>
}
```

^ We could probably change the `/history` endpoint to return an array of those objects, but omit the content.

ekinanp: Might also be worth adding a separate `journal/<id>` endpoint.

api/types/journal.go
```diff
@@ -1,4 +1,8 @@
 package apitypes
 
-// JournalIDHeader is the name of the HTTP Header used to provide a journal ID to assocate multiple actions.
+// JournalIDHeader is the name of the HTTP Header used to provide a activity ID to assocate multiple actions.
```
ekinanp: Typo? I think the last part should be something like "provide a journal ID that Wash will use to record the activity".

cmd/find.go
```diff
@@ -70,8 +70,7 @@ func findMain(cmd *cobra.Command, args []string) exitCode {
 		if p(&entry) {
 			entryPath := path
 			if len(entries) > 1 && i != 0 {
-				entryPath += "/"
-				entryPath += entry.CName
+				entryPath += "/" + entry.CName
```
ekinanp: https://github.com/puppetlabs/wash/pull/212 addresses this, so I'd take it out.

cmd/history.go
```diff
@@ -0,0 +1,78 @@
...
+		Short: "Prints the wash command history, or journal of a particular item",
+		Args:  cobra.MaximumNArgs(1),
+	}
+
+	historyCmd.RunE = toRunE(historyMain)
+
+	return historyCmd
+}
+
+func printJournalEntry(index string) error {
```
ekinanp: Yeah, I think a separate `historyEntry` type in the `activity` package would be cleaner.

activity/core.go
ekinanp: Could make the dead-letter-office its own Journal via an internal package variable

activity/journal.go
MikaelSmith: There's multiple roles being served here. Maybe I do need to break them out into separate things.

There's a type that's created from a process ID or server request, that's used to interact with the journal.

I think I can separate those into distinct things.

api/history.go
MikaelSmith: I eventually want `/history/<id>` to be able to stream updates. I feel like what you described is a more specialized thing.

activity/journal.go
MikaelSmith: Updated everything else. Still working on this.

api/history.go
ekinanp: Ok, that makes sense. Can we change the handler's name to `historyEntryHandler` (or something like it) so that it is at least more REST-y? `historyHandler` and `journalHandler` is still confusing to think about.

activity/core.go
```diff
@@ -60,51 +60,22 @@ func GetJournal(ctx context.Context) Journal {
...
-			Out:       f,
-			Level:     log.TraceLevel,
-			Formatter: &log.TextFormatter{TimestampFormat: time.RFC3339Nano},
-		}
-		return l, nil
-	})
-	if err != nil {
-		log.Warnf("Error creating journal %v: %v", id, err)
+	if journal.ID == "" {
+		journal.ID = "dead-letter-office"
```
ekinanp: I'd still like there to be a `deadLetterOfficeJournal` that we can set `journal` to, because mutating the type's a little weird. But it is not a blocker.

activity/journal.go
```diff
@@ -37,18 +40,48 @@ func (j Journal) registerCommand() {
...
+	log.Debugf(msg, a...)
+
+	// This is a single-use cache, so pass in an empty category.
+	obj, err := journalFileCache.GetOrUpdate("", j.ID, expires, true, func() (interface{}, error) {
+		jdir := Dir()
+		if err := os.MkdirAll(jdir, 0750); err != nil {
+			return nil, err
+		}
+
+		lpath := filepath.Join(jdir, j.ID+".log")
```
ekinanp: Nit: It'd be nice to define a `j.FilePath()` method to encapsulate the whole `j.ID + ".log"` computation since it is also used in `Open`.

api/types/journal.go
```diff
@@ -2,3 +2,7 @@ package apitypes
 
 // JournalIDHeader is the name of the HTTP Header used to provide a journal ID to assocate multiple actions.
 const JournalIDHeader = "JournalID"
```
ekinanp: Should we impose a regex on valid journal IDs? Asking b/c then we can have a single `JournalHeader` constructed as `<journal_id><sep><journal_description>`.

activity/journal.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Register the command.
+	storedJournals[j.ID] = len(history)
+	history = append(history, j)
+}
+
+// Record writes a new entry to the journal. It creates a new file for the journal if needed, then
+// appends the message to that journal. Journals are stored in the user's cache directory under
+// `wash/activity/ID.log`.
+func (j Journal) Record(msg string, a ...interface{}) {
```
ekinanp: Does this need to be exported?

api/history.go
MikaelSmith: Done.

api/types/journal.go
MikaelSmith: We could, but I don't feel there's a strong reason to.

activity/journal.go
MikaelSmith: No, but I don't see a reason not to. It seems like a valid part of normal operations for a Journal, and if for some reason you have direct access to it you might as well use it.

It might need some syncing on the printf operation though, I'm not sure that's thread-safe.

MikaelSmith:
Updated.

activity/journal.go
```diff
@@ -0,0 +1,119 @@
...
+
+// NewJournal creates a new journal entry with start time set to 'now'.
+func NewJournal(id, desc string) Journal {
+	// We set Start based on the time we first encounter the process, not when the process was
+	// started. This makes history make a little more sense when interacting with things like
+	// the shell, which was likely started before wash was.
+	return Journal{ID: id, Description: desc, start: time.Now()}
+}
+
+type historyBlob struct {
```
ekinanp: I like what you did with the `HistoryResponse` type in the API -- it's easy to think of history as an array of activities. It would be nice if `historyBlob` could reflect that symmetry, but it's not a blocker/could be something that's done later.

api/client/client.go
```diff
@@ -179,6 +185,47 @@ func (c *DomainSocketClient) Exec(path string, command string, args []string, op
...
+	var result apitypes.HistoryResponse
+	if err := json.Unmarshal(body, &result.Activities); err != nil {
+		return nil, fmt.Errorf("Non-JSON body at %v: %v", "/history", string(body))
+	}
+
+	return result.Activities, nil
+}
+
+// Journal returns a reader for the journal associated with a particular command in history.
+func (c *DomainSocketClient) Journal(index int) (io.ReadCloser, error) {
```
ekinanp: How about `ActivityJournal` ?

MikaelSmith: Sure.

activity/journal.go
MikaelSmith: I'm not sure what that would look like. You mean in terms of naming?

ekinanp: That and the type definition. So something like
```
type history struct {
        mux           sync.Mutex
	activities   []Activity
	stored       map[string]int
}
```

where activity may look like:
```
type Activity struct {
       start     time.Time
       journal Journal
}
```

PR #211, opened 2019-04-15
Extend Wash API to support local files and directories

ekinanp:
This commit adds the apifs package, which is used by the API to convert
local files/directories into Wash entries. This means that custom commands
built on the API are no longer required to special-case on them.

### DIFF
--- a/api/fs/core.go
+++ b/api/fs/core.go
@@ -4,8 +4,6 @@ package apifs
 
 import (
 	"os"
-	"syscall"
-	"time"
 
 	"github.com/puppetlabs/wash/plugin"
 )
@@ -17,15 +15,13 @@ type fsnode struct {
 }
 
 func newFSNode(finfo os.FileInfo, path string) *fsnode {
-	// TODO: Need to case this on platform since finfo.Sys()
-	// contains platform-specific data
-	statT := finfo.Sys().(*syscall.Stat_t)
+	// TODO: finfo.Sys() contains more detailed file attributes,
+	// but it's platform-specific. We should eventually use it for
+	// a more complete implementation of apifs.
 	attr := plugin.EntryAttributes{}
 	attr.
-		SetCtime(timespecToTime(statT.Ctimespec)).
-		SetAtime(timespecToTime(statT.Atimespec)).
-		SetMtime(timespecToTime(statT.Mtimespec)).
-		SetMode(os.FileMode(statT.Mode)).
+		SetMtime(finfo.ModTime()).
+		SetMode(finfo.Mode()).
 		SetSize(uint64(finfo.Size())).
 		SetMeta(plugin.ToMeta(finfo))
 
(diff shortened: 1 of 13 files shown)


plugin/cache.go
```diff
@@ -240,6 +236,10 @@ func cachedOp(ctx context.Context, opName string, entry Entry, ttl time.Duration
 		}
 	}
 
+	if ttl < 0 {
```
ekinanp: Had to do this because of the `apifs` package, where those entries do not have parents (that's also why I removed the g.id() == "" panic in `CachedList`).

Times when an empty ID would be bad is if someone tries to use a `CachedOp` (which the apifs entries don't), or if someone tries to use `plugin.ID`. I can extend `plugin.ID` to panic on an empty entry ID.

api/fs/core.go
```diff
@@ -0,0 +1,56 @@
...
+	// contains platform-specific data
+	statT := finfo.Sys().(*syscall.Stat_t)
+	attr := plugin.EntryAttributes{}
+	attr.
+		SetCtime(timespecToTime(statT.Ctimespec)).
+		SetAtime(timespecToTime(statT.Atimespec)).
+		SetMtime(timespecToTime(statT.Mtimespec)).
+		SetMode(os.FileMode(statT.Mode)).
+		SetSize(uint64(finfo.Size())).
+		SetMeta(plugin.ToMeta(finfo))
```
ekinanp: There's some weird behavior going on with JSON serialization b/c `wash meta` outputs an empty hash for local files/dirs when it shouldn't. I'd like to address that in a separate PR though.

api/fs/core.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if finfo.IsDir() {
+		return newDir(finfo, path), nil
+	}
+	return newFile(finfo, path), nil
+}
+
+func timespecToTime(t syscall.Timespec) time.Time {
```
ekinanp: Didn't add this to the munge package b/c syscall.Timespec is platform-specific, and I don't think we'll need any platform-specific stuff in munge.

plugin/cache.go
MikaelSmith: In practice the extra branches for failure checking are pretty much always worth it.

MikaelSmith:
Other than agreeing we should panic if `plugin.ID` should panic if the ID is empty, this code looks fine.

ekinanp:
@MikaelSmith I thought about adding a wrapper to the client that would implement the local file functionality (e.g. something like `client.AddLocalFSSupport(c)`), but quite a bit of that code is basically copy-pasting parts of the Wash server which to me indicates a bad abstraction. I think I'll leave it as-is for now.

One nice thing about doing the `client.AddLocalFSSupport` (or at least adapting local files/dirs into Wash entries) was that local files/dirs would automatically work with `wash find` w/o having to modify the command itself, but I think I'm ok with `wash find` not working on local files/dirs.

PR #212, opened 2019-04-16
 Implement depth-first recursion in `wash find` + path primary

ekinanp:

### DIFF
--- a/cmd/find/entry.go
+++ b/cmd/find/entry.go
@@ -6,12 +6,8 @@ import (
 )
 
 type entry struct {
-	*apitypes.Entry
+	apitypes.Entry
 	NormalizedPath string
-}
-
-func newEntry() entry {
-	return entry{Entry: &apitypes.Entry{}}
 }
 
 // info is a wrapper to c.Info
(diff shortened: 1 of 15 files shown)


cmd/find/entry.go
```diff
@@ -0,0 +1,47 @@
+package cmdfind
+
+import (
+	"github.com/puppetlabs/wash/api/client"
+	apitypes "github.com/puppetlabs/wash/api/types"
+)
+
+type entry struct {
+	*apitypes.Entry
```
MikaelSmith: I'm curious why this is a pointer. Not clear it's used that way.

ekinanp: Oh, so we avoid doing a bunch of copy-by-values, which I think may be bad for hundreds of thousands of entries. Can change it back if you prefer the non-pointer version.

MikaelSmith: I don't think it's actually avoiding those copies in lots of cases, and it's a small object so I wouldn't worry about it in the first place.

ekinanp: Updated.

PR #213, opened 2019-04-17
Split up `wash find` into multiple packages, default path to ".", add depth option

ekinanp:


cmd/find.go
```diff
@@ -1,17 +1,19 @@
...
+	"github.com/puppetlabs/wash/cmd/internal/find"
 	"github.com/spf13/cobra"
 )
 
 func findCommand() *cobra.Command {
 	findCmd := &cobra.Command{
-		Use: "find <path> [expression]",
+		// `wash find` parses its own flags to keep its syntax consistent with the
+		// existing `find` command
+		DisableFlagParsing: true,
```
ekinanp: It isn't hard to extend the parser to handle options (very similar to primaries), which is why I would like to implement familiar syntax for `wash find` (e.g. like `wash find <path> -mindepth 1 -maxdepth 10 -true -a -false ...`). Unfortunately, Cobra does not support `-<word>` flags based on what I'm seeing in https://github.com/spf13/cobra/issues/740.

We could have Cobra handle the flag parsing for us, but then the input would have to be something like
```
find --mindepth 1 --maxdepth 10 <path> -true -a -false ...
```

which is different from the existing find commands. Also, something like `find --mindepth 1 --maxdepth 10 -true -a -false` wouldn't work, because Cobra would try to parse `-true` and `-false` as flags, so we cannot default the path to `.` there.

I don't know which approach is best. Personally, I'd rather keep the syntax familiar, but I can understand it being undesirable for us to handle the command-line parsing ourselves.

ekinanp:
The `meta` primary is the biggest reason why I split `find` into separate packages, because I'm likely going to put `meta`'s code in a separate `primary/meta` package since the logic is likely going to be a bit complex.

cmd/find.go
ekinanp: Hmm, maybe. I'll look into that tomorrow. Hopefully it won't cause any confusion with the primaries.

ekinanp: Cool, looks like we can use it, and I can enforce syntax like `wash find [path] [options] [expression]`. One issue is that options can't be included in `[expression]`, which is something that the original `find` commands allowed. However, I think I'm OK with breaking consistency there because there isn't really a good reason to do that. The Linux find command also outputs a warning if you try something like `find path -true -o -depth`, here the warning would be "find: warning: you have specified the -depth option after a non-option argument -true, but options are not positional (-depth affects tests specified before it as well as those specified after it).  Please specify options before other arguments"

ekinanp: There may also be inconsistent syntax between options and primaries (e.g. like `find -mindepth=10 -true -o -false`) but meh, it's not that big of a deal.

cmd/internal/find/walker.go
```diff
@@ -0,0 +1,59 @@
...
+		opts: r.Options,
+		conn: conn,
+	}
+}
+
+func (w *walker) Walk(e types.Entry) {
+	w.walk(e, 0)
+}
+
+func (w *walker) walk(e types.Entry, depth uint) {
```
ekinanp: Go's stack limit is 1 GB. On my machine, https://gist.github.com/ekinanp/c79b4d164665d23057e65496785cf816 ran for 5592386 iterations. I don't think we're going to have directory hierarchies that are that large, so I'd rather keep the (simple) recursive algorithm if I can. Happy to switch to a stack, but it'll be a bit more complicated.

PR #214, opened 2019-04-17
Make EC2 and Docker instance names more identifiable

MikaelSmith:


plugin/aws/ec2Instance.go
```diff
@@ -43,8 +44,17 @@ const (
 )
 
 func newEC2Instance(ctx context.Context, inst *ec2Client.Instance, session *session.Session, client *ec2Client.EC2) *ec2Instance {
+	id := awsSDK.StringValue(inst.InstanceId)
+	name := id
+	for _, tag := range inst.Tags {
+		if awsSDK.StringValue(tag.Key) == "Name" {
```
ekinanp: This seems to suggest that our present use of `e.Name()` should really be `e.ID()` (where ID is the entry's ID w.r.t to the plugin API) and `e.Name` is `e.CName` (or what's displayed in the shell). Both `CName` and `ID` are configurable by plugin authors.

I don't think we currently need to change anything in the plugin architecture, but this is an interesting thing to keep in mind to see if other plugins will adopt similar conventions.

plugin/docker/containers.go
```diff
@@ -25,8 +26,14 @@ func (cs *containers) List(ctx context.Context) ([]plugin.Entry, error) {
 	activity.Record(ctx, "Listing %v containers in %v", len(containers), cs)
 	keys := make([]plugin.Entry, len(containers))
 	for i, inst := range containers {
+		name := inst.ID
+		if len(inst.Names) > 0 {
+			// Names must currently be unique in the docker runtime.
```
ekinanp: I'm not sure how this comment clarifies the `TrimPrefix` call here. A bit more explanation would be useful.

plugin/aws/ec2Instance.go
ekinanp: Also, a comment explaining what's in the `Name` tag would be nice.

plugin/docker/containers.go
MikaelSmith: Oh, that was why I didn't append the ID. I'll clarify the comment.

MikaelSmith:
Done.

PR #217, opened 2019-04-18
Extract client to interface; re-define meta + simplify plugins; fix bug in apifs

ekinanp:

### DIFF
--- a/plugin/entryAttributes.go
+++ b/plugin/entryAttributes.go
@@ -160,11 +160,15 @@ func (a *EntryAttributes) SetSize(size uint64) *EntryAttributes {
 	return a
 }
 
-// Meta returns the entry's metadata. All entries have a
-// meta attribute, which by default is an empty map.
+// Meta returns the portion of the entry's metadata that's returned by the
+// plugin API's List endpoint. This may or may not match what's returned by
+// e.Metadata (i.e. the entry's full metadata). That detail is plugin-specific.
+//
+// If a.SetMeta(m) was called, then this returns m. Otherwise, it returns
+// a.ToMap(false).
 func (a *EntryAttributes) Meta() EntryMetadata {
 	if a.meta == nil {
-		return EntryMetadata{}
+		return a.ToMap(false)
 	}
 
 	return a.meta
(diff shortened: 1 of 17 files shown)


ekinanp:
Here's what meta looks like for a local file:
```
enis:wash enis.inan$ ./wash meta docs/external_plugins/plugin_script.md
{
  "isDir": false,
  "modTime": "2019-04-17T13:50:46.054230642-07:00",
  "mode": 420,
  "name": "plugin_script.md",
  "size": 10168,
  "sys": {
    "Atimespec": {
      "Nsec": 234368346,
      "Sec": 1555538525
    },
    "Birthtimespec": {
      "Nsec": 54102807,
      "Sec": 1555534246
    },
    "Blksize": 4194304,
    "Blocks": 24,
    "Ctimespec": {
      "Nsec": 54230642,
      "Sec": 1555534246
    },
    "Dev": 16777220,
    "Flags": 0,
    "Gen": 0,
    "Gid": 20,
    "Ino": 23335103,
    "Lspare": 0,
    "Mode": 33188,
    "Mtimespec": {
      "Nsec": 54230642,
      "Sec": 1555534246
    },
    "Nlink": 1,
    "Pad_cgo_0": [
      0,
      0,
      0,
      0
    ],
    "Qspare": [
      0,
      0
    ],
    "Rdev": 0,
    "Size": 10168,
    "Uid": 503
  }
}
```

plugin/entryAttributes.go
```diff
@@ -160,11 +160,15 @@ func (a *EntryAttributes) SetSize(size uint64) *EntryAttributes {
 	return a
 }
 
-// Meta returns the entry's metadata. All entries have a
-// meta attribute, which by default is an empty map.
+// Meta returns the portion of the entry's metadata that's returned by the
```
MikaelSmith: Main thing I can't tell from looking at these changes is whether Meta will still be refreshed with new data when some cache expires. Also should Meta be cached at all?

ekinanp: Since it's tied to the `List` endpoint, it will be refreshed when the entry's refreshed.

I don't think we need to cache meta. For entries where `meta` represents the full metadata, we could probably remove the `Metadata` method (e.g. EC2 instances).

PR #218, opened 2019-04-18
Speed up volume listing, enable refresh of volume data, and fix /cache

MikaelSmith:


api/cache_test.go
```diff
@@ -86,13 +86,15 @@ func (suite *CacheHandlerTestSuite) TestClearCache() {
 	group.SetTestID("/dir")
 	group.On("List", mock.Anything).Return([]plugin.Entry{}, nil)
 
+	reqCtx := context.WithValue(context.Background(), mountpointKey, "/")
```
MikaelSmith: This could use some tests around the notCachedPathResponse error.

api/helpers.go
```diff
@@ -62,30 +62,54 @@ func findEntry(ctx context.Context, root plugin.Entry, segments []string) (plugi
...
 	}
 	if !filepath.IsAbs(path) {
-		return nil, "", relativePathResponse(path)
+		return "", relativePathResponse(path)
 	}
 
+	return path, nil
+}
+
+func getWashPathFromFullPath(ctx context.Context, path string) (string, *errorResponse) {
```
MikaelSmith: These might benefit from some unit tests.

plugin/docker/volume.go
```diff
@@ -35,7 +35,6 @@ func newVolume(c *client.Client, v *types.Volume) (*volume, error) {
 		EntryBase: plugin.NewEntry(v.Name),
 		client:    c,
 	}
-	vol.SetTTLOf(plugin.ListOp, 60*time.Second)
```
MikaelSmith: There may be other places I should tweak cache TTLs.

volume/dir_test.go
```diff
@@ -6,24 +6,32 @@ import (
...
 
+	"github.com/puppetlabs/wash/datastore"
 	"github.com/puppetlabs/wash/plugin"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestVolumeDir(t *testing.T) {
 	dmap, err := StatParseAll(strings.NewReader(fixture), mountpoint)
 	assert.Nil(t, err)
+	listcb := func(ctx context.Context) (DirMap, error) {
```
MikaelSmith: These tests feel terrible. Might try to clean them up.

api/cache_test.go
ekinanp: Add a TODO?

api/helpers.go
```diff
@@ -62,30 +62,54 @@ func findEntry(ctx context.Context, root plugin.Entry, segments []string) (plugi
...
 	mountpoint := ctx.Value(mountpointKey).(string)
 	trimmedPath := strings.TrimPrefix(path, mountpoint)
 	if trimmedPath == path {
+		return path, notCachedPathResponse(path)
+	}
+
+	return trimmedPath, nil
+}
+
+func getWashPathFromRequest(r *http.Request) (string, *errorResponse) {
```
ekinanp: I'm guessing you added this to avoid unnecessary API requests for endpoints that do not require an Entry object? If so, a comment explaining that would be nice.

api/helpers.go
```diff
@@ -62,30 +62,54 @@ func findEntry(ctx context.Context, root plugin.Entry, segments []string) (plugi
 	return curEntry, nil
 }
 
-func getPathFromRequest(r *http.Request) (string, *errorResponse) {
+func getFullPathFromRequest(r *http.Request) (string, *errorResponse) {
```
ekinanp: Hmm, what was wrong with `getPathFromRequest` ?

volume/dir.go
```diff
@@ -7,44 +7,36 @@ import (
...
 // NewDir creates a Dir populated from dirs.
-func NewDir(name string, attr plugin.EntryAttributes, cb ContentCB, path string, dirs DirMap) *Dir {
+func NewDir(name string, attr plugin.EntryAttributes, cacheEntry plugin.Entry, lb ListCB, cb ContentCB, path string) *Dir {
 	vd := &Dir{
-		EntryBase: plugin.NewEntry(name),
-		contentcb: cb,
-		path:      path,
-		dirs:      dirs,
+		EntryBase:  plugin.NewEntry(name),
+		cacheEntry: cacheEntry,
```
ekinanp: This package is getting a bit complicated with the callbacks + cacheEntry. It'd be nice to clean some of this up if possible, but that can be done in a separate PR.

api/helpers.go
MikaelSmith: getWashPathFromPath seemed weird, and getWashPathFromFullPath + getFullPathFromRequest has symmetry.

MikaelSmith:
Updated.

api/cache_test.go
```diff
@@ -116,6 +120,27 @@ func (suite *CacheHandlerTestSuite) TestClearCache() {
 	group.AssertNumberOfCalls(suite.T(), "List", 2)
 }
 
+func (suite *CacheHandlerTestSuite) TestCleareCacheErrors() {
```
ekinanp: Typo in `Cleare`

api/helpers.go
```diff
@@ -62,30 +62,59 @@ func findEntry(ctx context.Context, root plugin.Entry, segments []string) (plugi
...
-		return nil, "", relativePathResponse(path)
+		return "", relativePathResponse(path)
 	}
 
+	return path, nil
+}
+
+// Common subset used by getEntryFromRequest and getWashPathFromRequest.
+// getEntryFromRequest needs both the path and the wash path.
+func getWashPathFromFullPath(ctx context.Context, path string) (string, *errorResponse) {
```
ekinanp: `toWashPath` might be a better name here, b/c there's some munging going on.

volume/core.go
```diff
@@ -20,6 +20,16 @@ import (
 	"github.com/puppetlabs/wash/plugin"
 )
 
+// Interface presents methods to access the volume.
+type Interface interface {
```
ekinanp: This is a lot better. Only suggestions I have are:
* `Interface` => `Volume`
* `VolumeList` => `ListDir`
* `VolumeRead` => `ReadFile` (or `OpenFile` to keep it consistent w/ `Readable#Open`)
* `MakeEntries` => `List`

MikaelSmith: Then it would be `volume.Volume`. `<package>.Interface` is not uncommon in the go standard library. I'm open to other suggestions, but not sure `volume.Volume` is an improvement.

I went with VolumeList and VolumeRead to namespace them since we're adding on to the Entry interface.

PR #220, opened 2019-04-23
Find and view log files on execable resources

### LINKED ISSUE #216: Find and view log files for services on a VM
Mount `/var/log` on targets to `<vm>/fs/var/log`.

The default method is to implement over exec. Add a helper to make that behavior easy to opt into for anything that implements `Group` and `Execable`. For other modes, such as downloading files using Docker's API, plugins will need to implement that behavior themselves.

MikaelSmith:
Adds a `volume.FS` Entry-type helper for listing a remote volume on things that are Execable. Currently only exposes `/var/log` to keep performance reasonable for systems that have a lot of files.

Resolves #216.


ekinanp:
After some thought, I think the best approach might be to move the `volume` package over to plugin. Here's why I think that.

* The current approach requires us to export `SetParent` + `SlashReplacementChar` in the Entry interface, which introduces the possibility of plugin authors overriding those methods. We could add `plugin.SetParent` + `plugin.SlashReplacementChar` wrappers around those, but that still exports methods that people shouldn't have to care about.

* The caching behavior is part of the plugin interface, so it should all be in the plugin package (even more so when the cache config is embedded in EntryBase). I don't think it's worth compromising an important part of the design due to a Go-specific limitation.

So, even though it sucks, I'd rather we bloat `plugin/` with more files than compromise the simplicity of the existing plugin interface + architecture. I think we should be able to get away with exporting only `volume.Interface`, and maybe do some type-assertion in CachedList on it so that we don't have to export a separate `plugin.VolumeList` method?

plugin/entryBase.go
```diff
@@ -68,16 +69,18 @@ func (e *EntryBase) attributes() EntryAttributes {
...
 }
 
 func (e *EntryBase) id() string {
 	return e.washID
 }
 
-func (e *EntryBase) setID(id string) {
-	e.washID = id
+// SetParent sets the entry's ID based on its parent's ID. Used by plugincached.List.
+func (e *EntryBase) SetParent(parent Entry) {
```
ekinanp: Nit: `setIDFromParent` is more reflective of what's going on here.

MikaelSmith: Does that matter to the people calling it?

MikaelSmith: Does that matter to the people calling it?

ekinanp: Maybe for someone new to the codebase, but even then it's probably not a big deal. That's why I marked it as a nit.

MikaelSmith:
Reworked. It's quite a bit simpler now, with better tests.

volume/core.go
```diff
@@ -95,6 +98,29 @@ func StatParse(line string) (plugin.EntryAttributes, string, error) {
 // A DirMap is a map of directory names to a map of their children and the children's attributes.
 type DirMap = map[string]map[string]plugin.EntryAttributes
 
+func makedirs(attrs *map[string]map[string]plugin.EntryAttributes, newpath string) map[string]plugin.EntryAttributes {
```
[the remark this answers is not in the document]
ekinanp: I think it'd be better if all the stat code was moved over to a separate `volume/stat.go` file because it clutters `core.go` and makes it harder to understand.

volume/fs.go
```diff
@@ -0,0 +1,90 @@
...
+func (d *FS) List(ctx context.Context) ([]plugin.Entry, error) {
+	return List(ctx, d, "")
+}
+
+func consumeExec(ctx context.Context, result plugin.ExecResult) (*bytes.Buffer, error) {
+	var buf bytes.Buffer
+	for chunk := range result.OutputCh {
+		if chunk.Err != nil {
+			activity.Record(ctx, "Error on exec: %v", chunk.Err)
+		}
```
MikaelSmith: I think this should be `else if`. If there was an error, we shouldn't try to read data.

volume/fs.go
```diff
@@ -0,0 +1,89 @@
...
+// a directory tree based on supplied configuration (defaulting to `/var/log` if not specified).
+func (d *FS) List(ctx context.Context) ([]plugin.Entry, error) {
+	return List(ctx, d, "")
+}
+
+func consumeExec(ctx context.Context, result plugin.ExecResult) (*bytes.Buffer, error) {
+	var buf bytes.Buffer
+	for chunk := range result.OutputCh {
+		if chunk.Err != nil {
+			activity.Record(ctx, "Error on exec: %v", chunk.Err)
```
ekinanp: Hmm, this code implies that `consumeExec` will not return an error even if the command executed successfully, but something went wrong transmitting its output over the network (i.e. ExitCodeCB returning a zero exit code despite a streaming error). Is that intentional and if so, why?

plugin/docker/container.go
```diff
@@ -51,7 +52,8 @@ func (c *container) List(ctx context.Context) ([]plugin.Entry, error) {
 	clfAttr.SetSize(uint64(content.Size()))
 	clf.SetAttributes(clfAttr)
 
-	return []plugin.Entry{cm, clf}, nil
+	// Include a view of the remote filesystem using volume.FS
+	return []plugin.Entry{cm, clf, vol.NewFS("fs", c)}, nil
```
ekinanp: Hmm, we may need a config setting for this to translate to external plugins. Or maybe external plugins don't get to have this benefit. I guess the latter's the likely case here, given that introducing this for external plugins would cause a dependency cycle.

volume/fs_test.go
```diff
@@ -0,0 +1,135 @@
...
+	fs.SetTestID("/instance/fs")
+
+	entry := suite.dig(fs, "var", "log", "path1", "a file")
+	suite.Equal("a file", plugin.Name(entry))
+
+	execResult := createResult("hello")
+	exec.On("Exec", mock.Anything, "cat", []string{"/var/log/path1/a file"}, plugin.ExecOptions{}).Return(execResult, nil)
+	rdr, err := entry.(plugin.Readable).Open(context.Background())
+	suite.NoError(err)
+	suite.Equal(int64(5), rdr.Size())
```
ekinanp: Why not assert the content here?

volume/fs_test.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func TestCache(t *testing.T) {
```
ekinanp: Typo here

volume/fs_test.go
```diff
@@ -0,0 +1,135 @@
...
+	cmd := StatCmd("/var/log")
+	exec.On("Exec", mock.Anything, cmd[0], cmd[1:], plugin.ExecOptions{}).Return(createResult(varLogFixture), nil)
+	return exec
+}
+
+func (suite *fsTestSuite) dig(grp plugin.Group, names ...string) plugin.Entry {
+	entry := plugin.Entry(grp)
+	for _, name := range names {
+		entries, err := entry.(plugin.Group).List(context.Background())
+		suite.NoError(err)
```
ekinanp: I'd probably fail the test if an error was returned. Don't think `NoError` does that.

volume/fs_test.go
```diff
@@ -0,0 +1,135 @@
...
+func createExec() *mockExecutor {
+	exec := &mockExecutor{EntryBase: plugin.NewEntry("instance")}
+	// Used when recording activity.
+	exec.SetTestID("/instance")
+	cmd := StatCmd("/var/log")
+	exec.On("Exec", mock.Anything, cmd[0], cmd[1:], plugin.ExecOptions{}).Return(createResult(varLogFixture), nil)
+	return exec
+}
+
+func (suite *fsTestSuite) dig(grp plugin.Group, names ...string) plugin.Entry {
```
ekinanp: `find(grp plugin.Group, path string)` would make the tests easier to read.

volume/fs_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"context"
+	"testing"
+
+	"github.com/puppetlabs/wash/datastore"
+	"github.com/puppetlabs/wash/plugin"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/suite"
+)
+
+const varLogFixture = `
```
ekinanp: A brief comment indicating that this fixture represents the output of `StatCmd(/var/log)` would help.

volume/fs_test.go
```diff
@@ -0,0 +1,135 @@
...
+	entries, err := entry.List(context.Background())
+	suite.NoError(err)
+	suite.Equal(3, len(entries))
+
+	suite.Equal("path", plugin.Name(entries[0]))
+	suite.Equal("path1", plugin.Name(entries[1]))
+	suite.Equal("path2", plugin.Name(entries[2]))
+	for _, entry := range entries {
+		_, ok := entry.(plugin.Group)
+		suite.True(ok)
```
ekinanp: Think we'd want to fail the test if this assertion fails so that Lines 92 + 99 don't panic. May be worth doing a similar assertion in Line 79.

plugin/docker/container.go
MikaelSmith: I still think external plugins should be a separate package, which would solve the dependency cycle.

volume/core.go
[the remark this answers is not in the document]
MikaelSmith: I wasn't sure whether adding new entries to the map would be safe, and thought I was seeing an issue caused by it. I'll revert to using the reference; I think only arrays are different because you can create a new array with `append`.

plugin/docker/container.go
ekinanp: Yes, I think we could place them in a `plugin/internal` package so they're not exported in godocs. May need to fix some things with the Registry when registering external plugins in order to avoid a `plugin => internal => plugin` dep. cycle

volume/fs_test.go
MikaelSmith: NoError will flag the test as failed if err is an error.

ekinanp: Oh, I meant that it won't terminate the test.

volume/fs.go
MikaelSmith: I guess it makes sense to return that error. I wasn't sure the error would be fatal enough to avoid returning partial information.

volume/fs.go
```diff
@@ -0,0 +1,90 @@
...
+			activity.Record(ctx, "Stdout: %v", chunk.Data)
+			fmt.Fprint(&buf, chunk.Data)
+		} else {
+			activity.Record(ctx, "Stderr: %v", chunk.Data)
+		}
+	}
+
+	exitcode, err := result.ExitCodeCB()
+	if err != nil {
+		activity.Record(ctx, "Error exiting exec: %v", err)
```
ekinanp: I think these errors should include the failed command. Otherwise, it's not obvious what failed to exec without looking through the code.

volume/fs.go
```diff
@@ -0,0 +1,90 @@
...
+func (d *FS) List(ctx context.Context) ([]plugin.Entry, error) {
+	return List(ctx, d, "")
+}
+
+func consumeExec(ctx context.Context, result plugin.ExecResult) (*bytes.Buffer, error) {
+	var buf bytes.Buffer
+	for chunk := range result.OutputCh {
+		if chunk.Err != nil {
+			activity.Record(ctx, "Error on exec: %v", chunk.Err)
+			return nil, chunk.Err
```
ekinanp: Would be nice to indicate that this is a streaming error. Matches what we do in https://github.com/puppetlabs/wash/blob/master/api/error.go#L37.

Also, I think the early return may cause some issues with the channel closing because it wouldn't be able to read any data after the error so the caller would block.

MikaelSmith: It will. I guess it's probably better to close the channel and have the writer error.

MikaelSmith: Actually, can't close it because it's receive-only. So I think the context cancellation has to take care of it.

MikaelSmith: Should we keep reading until the end, then return the error?

volume/fs.go
ekinanp: I guess the activity records that info, so maybe it's not that important.

volume/fs.go
ekinanp: Yeah, I think reading until the end makes the most sense and is simple enough.

volume/fs.go
MikaelSmith: Clarified it a bit anyway.

plugin/types.go
```diff
@@ -31,9 +31,14 @@ import (
 	"time"
 )
 
-// Entry is a basic named resource type. It is a sealed
-// interface, meaning you must use plugin.NewEntry when
+// Entry is the interface for things that are representable by Wash's filesystem. This includes
+// plugin roots; resources like containers and volumes; placeholders (e.g. the containers directory
+// in the Docker plugin); read-only files like the metadata.json files for conainers and EC2
```
ekinanp: Typo in `conainers`.

volume/stat.go
```diff
@@ -0,0 +1,138 @@
...
+// Keep as its own specialized function as it will be faster than munge.ToTime.
+func parseTime(t string) (time.Time, error) {
+	epoch, err := strconv.ParseInt(t, 10, 64)
+	if err != nil {
+		return time.Time{}, err
+	}
+	return time.Unix(epoch, 0), nil
+}
+
+// StatParse parses a single line of the output of StatCmd into EntryAttributes and a name.
```
ekinanp: I'd change `name` => `path` since that's what it is returning. Basing that off of `StatParseAll`'s code, specifically the `fullpath` reference.

volume/stat.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// StatParseAll an output stream that is the result of running StatCmd. Strips 'base' from the
+// file paths, and maps each directory to a map of files in that directory and their attr
+// (attributes).
+func StatParseAll(output io.Reader, base string) (DirMap, error) {
+	scanner := bufio.NewScanner(output)
+	// Create lookup table for directories to contents, and prepopulate the root entry because
+	// the mount point won't be included in the stat output.
+	attrs := DirMap{"": make(Dir)}
```
ekinanp: I'd rename `attrs` to `dirMap` or `dirMp`, i.e. something that is more consistent with the return type.

volume/stat.go
```diff
@@ -0,0 +1,138 @@
...
+	for scanner.Scan() {
+		text := strings.TrimSpace(scanner.Text())
+		if text != "" {
+			attr, fullpath, err := StatParse(text)
+			if err != nil {
+				return nil, err
+			}
+
+			relative := strings.TrimPrefix(fullpath, base)
```
ekinanp: Hmm, I think we could move this part + `makedirs` over to a `DirMap#add(<attr>, <path>)` helper. That would better reflect what all this logic's doing.

May be worth creating a `volume.Node` type to encapsulate `<attr>` + `<path>`, but that's not that important. However, at least commenting that `<attr>` + `<path`> represent a volume node would be nice.

volume/fs.go
```diff
@@ -30,11 +30,17 @@ func (d *FS) List(ctx context.Context) ([]plugin.Entry, error) {
 	return List(ctx, d, "")
 }
 
-func consumeExec(ctx context.Context, result plugin.ExecResult) (*bytes.Buffer, error) {
+func exec(ctx context.Context, executor plugin.Execable, cmdline []string) (*bytes.Buffer, error) {
+	result, err := executor.Exec(ctx, cmdline[0], cmdline[1:], plugin.ExecOptions{})
```
ekinanp: Don't see the cmdline being used anywhere else in the code, including the errors. Is that intentional?

plugin/docker/container.go
MikaelSmith: Yes, tweaks to the Registry would be needed.

volume/fs.go
MikaelSmith: Yes, the callers generally add it to their errors.

volume/stat.go
MikaelSmith: I think that's trickier than it seems, and isn't something that needs to be exposed. DirMap is an alias, I'd rather not promote it to a new type.

ekinanp: Ok, it's not that important anyways.

PR #221, opened 2019-04-23
Add the initial version of the `meta` primary

ekinanp:


cmd/internal/find/primary/meta/numericPredicate.go
```diff
@@ -0,0 +1,39 @@
...
+func parseNumericPredicate(tokens []string) (predicate, []string, error) {
+	if len(tokens) == 0 {
+		return nil, nil, errz.NewMatchError("expected a +, -, or a digit")
+	}
+	token := tokens[0]
+	numericP, _, err := numeric.ParsePredicate(
+		token,
+		numeric.ParsePositiveInt,
+		numeric.Negate(numeric.ParsePositiveInt),
+		numeric.ParseSize,
```
ekinanp: We can add a `numeric.ParseFrequency` value here when the time comes.

cmd/internal/find/primary/meta/timePredicate.go
```diff
@@ -0,0 +1,59 @@
...
+		// to return true if 'timeV - StartTime' > 1h. Unfortunately, there
+		// isn't a clean way to mathematically enforce this without complicating
+		// the parsing logic. However, we can get a pretty good approximation
+		// by inverting numericP. For the +{1h} example, inverting numericP means
+		// that numericP returns true if 'StartTime - timeV' <= -1h which reduces
+		// to returning true if 'timeV - StartTime' >= 1h. This is not quite correct
+		// because numericP will return true if timeV == startTime, but given that
+		// Go's time.Time objects have nanosecond precision, this is a very rare
+		// (if not impossible) edge case so negating numericP should be good enough.
+		numericP = numeric.Not(numericP)
```
ekinanp: We can document this as something like:

`+1h` can be thought of as "more than one hour ago", while `+{1h}` can be thought of as "more than one hour from now". Thus, `{}` shifts the semantics from "ago" to "from now".

cmd/internal/find/primary/meta/stringPredicate.go
```diff
@@ -0,0 +1,29 @@
+package meta
+
+import (
+	"fmt"
+
+	"github.com/puppetlabs/wash/cmd/internal/find/primary/errz"
+)
+
+// StringPredicate => [^-]+
```
ekinanp: I think regex/glob support should be next, since this isn't really that useful right now. I'm thinking the user would type something like
```
find instances .Tags[] r'termination_date'
```
where the regex grammar is `r<regex>`. Note that `r'termination_date'` would be passed as `rtermination_date` into the parser (`'`/`"` is syntactic sugar).

cmd/internal/find/primary/numeric/predicate.go
```diff
@@ -0,0 +1,72 @@
...
+	if cmp == '+' || cmp == '-' {
+		str = str[1:]
+	} else {
+		cmp = '='
+	}
+
+	var parserID int
+	var n int64
+	var err error
+	for i, parser := range parsers {
```
ekinanp: This logic's similar to the `try` function in `meta/grammar.go`. Unfortunately, I can't abstract that logic out b/c I don't want to do unnecessary type assertions in numeric predicates.

cmd/internal/find/primary/meta/arrayPredicate.go
```diff
@@ -0,0 +1,145 @@
...
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/puppetlabs/wash/cmd/internal/find/primary/errz"
+)
+
+// ArrayPredicate => EmptyPredicate      |
+//                   ‘[' ‘]’ Predicate   |
```
ekinanp: Thinking we could express `some` as `[?]` to make it more explicit.

cmd/internal/find/primary/meta/stringPredicate.go
ekinanp: We may need to distinguish string literals by requiring users to double quote them. Or we could preserve the existing behavior of "fallback to literal string".

ekinanp:
I'll add the "default to mindepth/maxdepth 1" logic in a separate PR. Right now, I recommend testing this with `-mindepth 1 -maxdepth 1`.

cmd/internal/find/params/core.go
```diff
@@ -0,0 +1,8 @@
+// Package params represents `wash find`'s parameters. These are typically
+// set in `wash find`'s main function.
+package params
```
ekinanp: Open to a better name for this, or another location. Main reason for the separate package is to avoid an import cycle (`meta` => `primary` => `meta`).

cmd/internal/find/primary/meta/timePredicate.go
ekinanp: To check if something expired "now", users can do `+1s` (meaning "more than a second ago").

cmd/internal/find/primary/meta/stringPredicate.go
ekinanp: Also, on my Iterm shell, we don't need to quite `.Tags[]`. It works as-is.

MikaelSmith:
I think it'd be useful to include the google doc specification you wrote for this somewhere.

cmd/internal/find/params/core.go
MikaelSmith: None stand out to me. Params seems fine.

cmd/internal/find/primary/size.go
```diff
@@ -3,35 +3,13 @@ package primary
...
-	}
-	panic(fmt.Sprintf("cmdfind.bytesOf received an unexpected unit %v", unit))
-}
-
-var sizeValueRegex = regexp.MustCompile(`^(\+|-)?((\d+)|(\d+[ckMGTP]))$`)
-
-// sizePrimary => -size (+|-)?((\d+)|(\d+[ckMGTP]))
-//
-// where c => character (byte), k => kibibyte, M => mebibyte, G => gibibyte, T => tebibyte, P => pebibyte
+// sizePrimary => -size (+|-)?(\d+ | util.SizeRegex)
```
MikaelSmith: Why does this say `util.SizeRegex`?

cmd/internal/find/primary/meta/predicate.go
```diff
@@ -0,0 +1,29 @@
...
+func parsePredicate(tokens []string) (predicate, []string, error) {
+	p, tokens, err := try(
+		tokens,
+		parseObjectPredicate,
+		parseArrayPredicate,
+		parsePrimitivePredicate,
+	)
+	if err != nil {
+		if errz.IsMatchError(err) {
+			return nil, nil, errz.NewMatchError("expected either a primitive, object, or array predicate")
```
MikaelSmith: This error message may not be particularly useful. It almost certainly requires reading some docs that don't yet exist.

cmd/internal/find/primary/meta/stringPredicate.go
MikaelSmith: I think the pattern needs something more, `[^-]+` is incomplete unless you mean to say the string can't have a dash anywhere in it. That doesn't seem to be enforced by the predicate.

MikaelSmith: On the fence here. I like making the regex behavior easy (I think `ag` went that way). But that can also be annoying. I guess I'm comfortable saying we'll need some sigil to denote a regex, unless we think people will almost always use a regex instead of an exact string match.

cmd/internal/find/primary/meta/stringPredicate_test.go
```diff
@@ -0,0 +1,42 @@
...
+	// Test the happy cases first
+	suite.runTestCases(
+		nPTC("foo -size", "-size", "foo"),
+	)
+
+	// Now test that the predicate returns false for a non-string
+	// value or if value != s
+	p, _, err := parseStringPredicate(toTks("foo"))
+	if suite.NoError(err) {
+		suite.False(p(200))
```
MikaelSmith: I'd like to also see a 'true' test.

cmd/internal/find/primary/meta/timePredicate.go
MikaelSmith: And `-1h` means less than one hour ago, `-{1h}` means less than one hour from now?

MikaelSmith: Also, does `+0` not work?

cmd/internal/find/primary/meta/timePredicate_test.go
```diff
@@ -0,0 +1,61 @@
...
+		nPETC("", `expected a \+, -, or a digit`, true),
+		nPETC("200", "expected a duration", true),
+		nPETC("+{", ".*closing.*}", false),
+	)
+}
+
+func (suite *TimePredicateTestSuite) TestValidInput() {
+	// Test the happy cases first
+	suite.runTestCases(
+		nPTC("+2h -size", "-size", addTST(-3*numeric.DurationOf('h'))),
```
MikaelSmith: I think it'd be useful to have more negative tests where a valid value fails the predicate.

cmd/internal/find/primary/meta/objectPredicate.go
```diff
@@ -0,0 +1,79 @@
...
+	// EmptyPredicate did not match, so try '.' Key Predicate
+	if len(tokens) == 0 {
+		return nil, nil, errz.NewMatchError("expected a key sequence")
+	}
+	tk := tokens[0]
+	if tk[0] != '.' {
+		return nil, nil, errz.NewMatchError("key sequences must begin with a '.'")
+	}
+	tk = tk[1:]
+	loc := keyRegex.FindStringIndex(tk)
```
MikaelSmith: This seems to allow a key sequence that doesn't immediately follow the `.`, because the regex doesn't start with `^`.

MikaelSmith: What would a valid instance of this look like?

cmd/internal/find/primary/meta/objectPredicate.go
```diff
@@ -0,0 +1,79 @@
...
+	if loc == nil {
+		return nil, nil, fmt.Errorf("expected a key sequence after '.'")
+	}
+	key := tk[loc[0]:loc[1]]
+	if len(tk) == loc[1] {
```
MikaelSmith: I'm not really sure what either of these comments mean. Can you explain a bit more?

cmd/internal/find/primary/meta/arrayPredicate.go
MikaelSmith: Hmm. Not no. This is a bit distinct from jq in that it's explicitly a filter, rather than a map operation.

cmd/internal/find/primary/meta/timePredicate.go
ekinanp: Yup, `-{1h}` does mean less than one hour from now. `+0` doesn't work b/c that'll be parsed as a NumericPredicate, but `+0s` does work. I guess people could use that. It just seemed weird to think about `more than 0 hours ago`, but maybe that's just me.

cmd/internal/find/primary/meta/stringPredicate.go
ekinanp: Yeah, I think people using a regex is more likely than an exact string match so we should make it easy. Guess that would mean `StringPredicate`'s semantics would be matching the regex compiled by the token. If exact string matches are common, and those matches are likely to contain regexp characters, we could always add quoted string literals to the grammar.

If you like ^, I can go ahead and add it to the PR. Can do it separately too since it's big enough.

cmd/internal/find/primary/meta/arrayPredicate.go
ekinanp: Yeah, I think `[?]` is better. I'll update the PR later with that modification.

cmd/internal/find/primary/size.go
ekinanp: Typo -- `numeric` was originally called `util`. Will fix that in the morning.

cmd/internal/find/primary/meta/predicate.go
ekinanp: Hmm, it isn't used anywhere b/c `ArrayPredicate` + `ObjectPredicate` only check the match error then return their own error messages. Should I just return `err` here?

cmd/internal/find/primary/meta/stringPredicate.go
ekinanp: You're right, that was a brainfart on my part. Should be `[^-].*`. Will update later.

cmd/internal/find/primary/meta/stringPredicate_test.go
ekinanp: That should be addressed on line 26. `nPTC` is `(<input>, <rem>, <trueV>)`.

cmd/internal/find/primary/meta/timePredicate_test.go
ekinanp: Ok, can add those later this morning.

cmd/internal/find/primary/meta/objectPredicate.go
ekinanp: Hmm, was thinking of being as flexible as possible there, but the regex does allow for spaces and such (on top of missing the `^`). I'll revise this later.

Given `.key1`, valid instance would be `key1`. Given `.key1.key2`, valid instance would be `key1`; postmatch would be `.key2`

cmd/internal/find/primary/meta/objectPredicate.go
ekinanp: Sure.

cmd/internal/find/primary/meta/stringPredicate.go
ekinanp: Hmm, maybe we do need a sigil to avoid weird magic. For example, if someone types `[abc]`, that will be parsed as an ArrayPredicate and a syntax error will be thrown. I could make ArrayPredicate return match errors so it can fall-back to StringPredicate, but that seems pretty hacky and can be hard to reason about if `meta` gets more complicated.

I guess telling people "Hey, you need to type r'<regex>'" for regex support isn't that bad.

ekinanp:
Wait, one more thing I want to fix.

ekinanp:
Ok, updated.

cmd/internal/find/primary/meta/predicate.go
MikaelSmith: Sure.

cmd/internal/find/primary/meta/stringPredicate_test.go
MikaelSmith: Ah, I guess that works. I haven't figured out how to read some of the inputs yet. Why is `-size` in so many of them?

ekinanp: Just to assert that the parse functions return the rest of `tokens`. Seemed easy enough to to test.

cmd/internal/find/primary/meta/arrayPredicate.go
```diff
@@ -0,0 +1,151 @@
...
+	// parseArrayPredicateType will return an error.
+	ptype, token, err := parseArrayPredicateType(rawToken)
+	if err != nil {
+		return nil, nil, err
+	}
+	if len(token) == 0 {
+		tokens = tokens[1:]
+	} else {
+		// token may be part of a key sequence (e.g. something like
+		// [].key2 or [][])
```
MikaelSmith: Lots of examples of `[]` still around in code comments. Worth updating them to `[?]`?

ekinanp: Updated.

PR #226, opened 2019-04-25
Add action primary; return false for time-mismatches in TimePredicate

ekinanp:

### DIFF
--- a/cmd/internal/find/primary/action.go
+++ b/cmd/internal/find/primary/action.go
@@ -9,43 +9,26 @@ import (
 	"github.com/puppetlabs/wash/cmd/internal/find/types"
 )
 
-// actionPrimary => <action> (, <action>)?
+// actionPrimary => <action>
 //nolint
 var actionPrimary = grammar.NewAtom([]string{"-action"}, func(tokens []string) (types.Predicate, []string, error) {
 	tokens = tokens[1:]
 	if len(tokens) == 0 {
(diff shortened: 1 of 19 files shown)


cmd/internal/find/primary/action.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/puppetlabs/wash/plugin"
+	"github.com/puppetlabs/wash/cmd/internal/find/grammar"
+	"github.com/puppetlabs/wash/cmd/internal/find/types"
+)
+
+// actionPrimary => <action> (, <action>)?
```
MikaelSmith: What's the behavior of multiple arguments? To get and/or, you can use `-action exec -and|-or -action list`. `-action exec,list` is a short-hand for which of those? Looking at the code, it looks like `-or`; I'd like the comment to clarify that as well.

MikaelSmith: I'm not sure the short-hand provided is useful enough. I'd be tempted to leave it out for now since it can be accomplished more clearly by combining primaries.

ekinanp: Hmm, I was thinking the shorthand could save some typing because I don't feel it's likely people will query only single actions at a time. That being said, shorthands are only when stuff's annoying and we haven't had enough Wash users to figure out what sucks and what doesn't, so I'll go ahead and remove it for now.

PR #229, opened 2019-04-25
Stream logs from VMs

### LINKED ISSUE #227: Stream logs from VMs
Expanding on #216, we should be able to stream logs from VMs rather than just open them.

Implement the Streamable interface for volumes, probably with tail -f as the backend. Tricky part is (maybe) stopping the command on the target.

MikaelSmith:
Add Stream support to the volume package. Add Tty support to Exec on Docker and Kubernetes to support correctly cleaning up processes after terminating streaming.

Resolves #227.

### DIFF
--- a/plugin/types.go
+++ b/plugin/types.go
@@ -63,9 +63,25 @@ type Root interface {
 }
 
 // ExecOptions is a struct we can add new features to that must be serializable to JSON.
-// Examples of potential features: user, privileged, tty, map of environment variables, string of stdin, timeout.
+// Examples of potential features: user, privileged, map of environment variables, timeout.
 type ExecOptions struct {
+	// Stdin can be used to pass a stream of input to write to stdin when executing the command.
 	Stdin io.Reader
+
+	// Tty instructs the executor to allocate a TTY (pseudo-terminal), which lets Wash communicate
+	// with the running process via its Stdin. The TTY is used to send a process termination signal
+	// (Ctrl+C) via Stdin when the passed-in Exec context is cancelled.
+	//
+	// NOTE TO PLUGIN AUTHORS: The Tty option is only relevant for executors that do not have an API
+	// endpoint to stop a running command (e.g. Docker, Kubernetes). If your executor does have an
+	// API endpoint to stop a running command, then ignore the Tty option. Note that the reason we
+	// make Tty an option instead of having the relevant executors always attach a TTY is because
+	// attaching a TTY can change the behavior of the command that's being executed.
+	//
+	// NOTE TO CALLERS: The Tty option is useful for executing your own stream-like commands (e.g.
+	// tail -f), because it ensures that there are no orphaned processes after the request is
+	// cancelled/finished.
+	Tty bool
 }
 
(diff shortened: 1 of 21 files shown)


MikaelSmith:
EC2 probably needs something here as well, but it seems more complicated. I think we would probably need to send another message over SSM to perform cleanup. I'm not sure how much it's worth investing in SSM given the interface and how rarely it's used compared to SSH.

plugin/types.go
```diff
@@ -77,6 +78,9 @@ type ExecOutputChunk struct {
 	Err       error
 }
 
+// StreamTypes provides the name of the corresponding StreamID.
+var StreamTypes = []string{"Stdout", "Stderr"}
```
ekinanp: Hmm, think we should just promote `StreamID` to a type and attach a `Name` method. Then callers can do `chunk.StreamID.Name()` to get this info.

Probably worth re-using that logic in https://github.com/puppetlabs/wash/blob/master/api/types/exec.go#L38-L40

MikaelSmith: I'll explore this a bit. Can probably be added on as a separate commit or PR.

plugin/types.go
```diff
@@ -63,9 +63,14 @@ type Root interface {
 }
 
 // ExecOptions is a struct we can add new features to that must be serializable to JSON.
-// Examples of potential features: user, privileged, tty, map of environment variables, string of stdin, timeout.
+// Examples of potential features: user, privileged, map of environment variables, timeout.
 type ExecOptions struct {
+	// Stdin can be used to pass
```
ekinanp: Missing some info here

plugin/types.go
```diff
@@ -63,9 +63,14 @@ type Root interface {
...
 	Stdin io.Reader
+	// Tty instructs the execution environment to allocate a TTY. In some cases this changes the
+	// behavior of the command being executed. For commands that don't naturally exit, it also
+	// instructs the execution environment to ensure the command is terminated when the context ends.
+	Tty bool
```
ekinanp: Hmm, how about something like:

> Tty instructs the executor to allocate a TTY (pseudo-terminal), which lets Wash communicate with the running process via its Stdin. The TTY is used to send a process termination signal (Ctrl+C) via Stdin when the passed-in Exec context is cancelled.
> 
> NOTE TO PLUGIN AUTHORS: The Tty option is only relevant for executors that do not have an API endpoint to stop a running command (e.g. Docker, Kubernetes). If your executor does have an API endpoint to stop a running command, then ignore the Tty option. Note that the reason we make Tty an option instead of having the relevant executors always attach a TTY is because attaching a TTY can change the behavior of the command that's being executed.
>
> NOTE TO CALLERS: The Tty option is useful for executing your own stream-like commands (e.g. `tail -f`), because it ensures that there are no orphaned processes after the request is cancelled/finished.
>

^ This wording is basically a compilation of all your answers to my questions about this, which I imagine are questions that other plugin authors/callers might have when they see this option.

MikaelSmith: Done.

PR #234, opened 2019-05-07
Implement EC2 over SSH

### LINKED ISSUE #230: Support SSH for AWS targets
SSM for AWS is not commonly enabled and is pretty slow. It also has very different permission profiles than SSH.

We need to support SSH for AWS targets, and it should probably be the default behavior.

Use golang SSH library initially for speed of development, relying on a correctly configured local ssh-agent. We want to keep open a path for using Bolt's inventory and/or SSH transport later as our customization needs increase. Should also consider using scp for file transfers.

Should we support configuring use of SSM instead of SSH for targets? Should that be on startup, or something you can request for a specific Exec call? Allowing it per Exec introduces extra complexity for everything that builds on Exec, so maybe we should have a separate file target representing the SSM transport.

MikaelSmith:
Use SSH to execute against EC2 instances. This performs better and is the default connection method setup for all (Linux-based) EC2 instances.

Move SSM implementation to a file so it's still available as an experimental option. Eventually we'll need to implement something different - either WinRM or use SSM - for Windows machines.

Resolves #230.


transport/ssh.go
```diff
@@ -0,0 +1,111 @@
...
+func sshConnect(host, port, user string) (*ssh.Client, error) {
+	agent, err := newAgent()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to find config from SSH_AUTH_SOCK: %s", err)
+	}
+
+	sshConfig := &ssh.ClientConfig{
+		User: user,
+		Auth: []ssh.AuthMethod{agent},
+		// TODO: use https://skarlso.github.io/2019/02/17/go-ssh-with-host-key-verification/
```
MikaelSmith: TODO for security.

transport/ssh.go
```diff
@@ -0,0 +1,111 @@
...
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+	}
+
+	// TODO: cache connections
```
MikaelSmith: TODO for performance.

plugin/aws/ec2Instance.go
```diff
@@ -238,457 +191,13 @@ func (inst *ec2Instance) checkLatestConsoleOutput(ctx context.Context) (*ec2Inst
...
-
-	execResult.OutputCh = outputCh
-	execResult.ExitCodeCB = func() (int, error) {
-		if exitCodeErr != nil {
-			return 0, exitCodeErr
-		}
-
-		return exitCode, nil
-	}
+	dnsname := meta["PublicDnsName"].(string)
```
MikaelSmith: This may not be configured. Need to try public DNS name, then public IP, then error cleanly.

plugin/aws/ec2Instance.go
```diff
@@ -238,457 +191,13 @@ func (inst *ec2Instance) checkLatestConsoleOutput(ctx context.Context) (*ec2Inst
...
-	deleteCtx, cancelFunc := context.WithTimeout(context.Background(), 3*time.Second)
-	defer cancelFunc()
-
-	// DeleteLogStream's response does not contain any useful information
-	// so we can discard it
-	_, err := s.client.DeleteLogStreamWithContext(deleteCtx, request)
-	if err != nil {
-		activity.Record(ctx, "Failed to delete the logs for %v: %v", s.name, err)
-	}
+	return transport.ExecSSH(ctx, transport.Identity{Host: dnsname, User: "ec2-user"}, append([]string{cmd}, args...), opts)
```
MikaelSmith: How do we decide what user to connect as? New option for exec?

Ideally we'd have a way to determine the user we should connect as from metadata, but not always exposed. We need a default (or configurable) user for using `fs`.

transport/ssh.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	// Run command via session
+	session, err := connection.NewSession()
+	if err != nil {
+		return result, fmt.Errorf("Failed to create session: %s", err)
+	}
+
+	outputch, stdout, stderr := plugin.CreateExecOutputStreams(ctx)
+	session.Stdin, session.Stdout, session.Stderr = opts.Stdin, stdout, stderr
```
MikaelSmith: Needs additional work to support TTY and context completion.

transport/ssh.go
ekinanp: Would be nice to note that in the TODO

transport/ssh.go
ekinanp: Would be nice to note that in the TODO

transport/ssh.go
ekinanp: Why do we need TTY support? If it's to kill the process, couldn't we just use https://godoc.org/golang.org/x/crypto/ssh#Session.Signal ?

plugin/aws/ec2Instance.go
```diff
@@ -186,6 +132,13 @@ func (inst *ec2Instance) List(ctx context.Context) ([]plugin.Entry, error) {
 		entries = append(entries, latestConsoleOutput)
 	}
 
+	entries = append(entries, &ec2InstanceSSM{
+		EntryBase:        plugin.NewEntry("ssm"),
```
ekinanp: This is weird, because SSM is not a "thing" that belongs to an EC2 instance. Could we remove this?

plugin/aws/ec2Instance.go
ekinanp: I think we could go with ec2-user based on what I'm seeing in https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AccessingInstancesLinux.html.

plugin/aws/ec2Instance.go
MikaelSmith: I guess I can remove it completely. Wanted to keep it around to experiment with somehow; it should probably end up being a config and I don't want to deal with the config yet.

ekinanp: We could always prefix unused files with `_` ?

transport/ssh.go
MikaelSmith: Good point. I'd noticed that. I think I could probably always send a signal if still waiting when the context finishes.

plugin/aws/ec2Instance.go
MikaelSmith: ec2-user is only the default for Amazon AMIs. It's not a bad default, using ssh config to override it. If I can find an authoritative source I can query or hard-code I might make it smarter.

transport/ssh.go
```diff
@@ -0,0 +1,180 @@
...
+
+		hostKeyCallback, err := getHostKeyCallback()
+		if err != nil {
+			return nil, fmt.Errorf("Loading SSH known hosts file: %v", err)
+		}
+
+		sshConfig := &ssh.ClientConfig{
+			User:            user,
+			Auth:            []ssh.AuthMethod{agent},
+			HostKeyCallback: hostKeyCallback,
```
MikaelSmith: I'm not sure it's worth reading known hosts yet, since it's very likely the fingerprint won't have been added for an arbitrary EC2 instance yet. I could also recommend adding
```
Host *.compute.amazonaws.com
  StrictHostKeyChecking no
```
and disable it that way.

ekinanp: When would reading known hosts be bad/annoying? How likely are those scenarios?

transport/ssh.go
```diff
@@ -0,0 +1,180 @@
...
+	if err := session.Start(cmdStr); err != nil {
+		return result, err
+	}
+
+	// Wait for session to complete and stash result. This is done asynchronously because the
+	// exitcode callback won't be called until stdout/stderr are closed, but they're not closed by
+	// session.Start.
+	var exitErr error
+
+	go func() {
```
ekinanp: This pattern is becoming common enough that it might be worth including some sort of cancellation function in `ExecResult`. That way as a plugin author, I don't have to worry about contexts, which is Go-specific.

ekinanp: We could require the post-condition that if Exec returns without an error, then the command was successfully started. That would make it easier to reason about how this stuff works.

transport/ssh.go
MikaelSmith: Ok, made that work.

transport/ssh.go
MikaelSmith: That makes some sense.

MikaelSmith:
Switching to SSH for AWS means trying to list directories may encounter "permission denied". Figuring out how I want to deal with that.

plugin/kubernetes/pod.go
```diff
@@ -146,31 +146,20 @@ func (p *pod) Exec(ctx context.Context, cmd string, args []string, opts plugin.E
 
 	outputCh, stdout, stderr := plugin.CreateExecOutputStreams(ctx)
 	exitcode := 0
+	var cancelWriter io.WriteCloser
 	go func() {
```
ekinanp: Hmm, think there's a race condition here (Exec returning before cancelWriter's set).

MikaelSmith: Oh yup, definitely.

MikaelSmith: Fixed.

MikaelSmith:
Ok, I think I've addressed most of the issues. `Elevate` is full of compromises, but I think is the most sensible default. I do wonder if `fs` should fall-back to trying with `Elevate: false` if it fails.

plugin/kubernetes/pod.go
```diff
@@ -186,6 +175,16 @@ func (p *pod) Exec(ctx context.Context, cmd string, args []string, opts plugin.E
 	}()
 
 	execResult.OutputCh = outputCh
+	if cancelWriter != nil {
+		execResult.CancellationCB = func() {
```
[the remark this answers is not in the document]
MikaelSmith: Oh yeah, moving more out of the go-routine cleaned that up.

PR #239, opened 2019-05-08
Support predicate expressions in the `meta` primary

ekinanp:
This commit adds support for predicate expressions to the `meta`
primary. Predicate expressions allow users to create more expressive
and powerful predicates. For example, the changes here reduce the
AWS reaper case from the verbose “-m .Tags[?].key termination_date -a -m .Tags[?].value +{1h}”
to the more expressive “-m .Tags[?] .key termination_date -a .value +{1h}”.
See the test cases in primary/meta_test.go for more examples.

### DIFF
--- a/cmd/internal/find/parser/expression/parser.go
+++ b/cmd/internal/find/parser/expression/parser.go
@@ -64,7 +64,7 @@ func NewParser(predicateParser predicate.Parser) *Parser {
 		}
 	}
 	p.atom = &predicate.CompositeParser{
-		ErrMsg: "expected an atom",
+		MatchErrMsg: "expected an atom",
 		Parsers: []predicate.Parser{
 			notOpParser(p),
 			parensOpParser(p),
--- a/cmd/internal/find/parser/expression/parser.go
+++ b/cmd/internal/find/parser/expression/parser.go
@@ -183,15 +183,22 @@ func (parser *Parser) Parse(tokens []string) (predicate.Predicate, []string, err
 	}
 	if _, ok := parser.stack.Peek().(*BinaryOp); ok {
 		// This codepath is possible via something like "p1 -and" or "p1 -and <unknown_token>"
-		if err != nil {
-			// We have "p1 -and <unknown_token>"
-			return nil, nil, err
+		if err == nil {
+			// We have "p1 -and"
+			return nil, nil, fmt.Errorf(
+				"%v: no expression after %v",
+				parser.stack.mostRecentOpToken,
+				parser.stack.mostRecentOpToken,
+			)
 		}
-		return nil, nil, fmt.Errorf(
-			"%v: no expression after %v",
-			parser.stack.mostRecentOpToken,
-			parser.stack.mostRecentOpToken,
-		)
+		// We have "p1 -and <unknown_token>". Pop the binary op off the stack and include
+		// it as part of the remaining tokens. This is useful in case our expression is inside
+		// another expression, where the top-level expression handles combining our parsed
+		// predicate p with whatever's parsed by the "<unknown_token>" bit. For example, it
+		// ensures that the top-level `wash find` parser correctly parses something like
+		// "-m .key foo -o -m .key bar" as "Meta(.key, foo) -o Meta(.key, bar)".
+		parser.stack.Pop()
+		tokens = append([]string{parser.stack.mostRecentOpToken}, tokens...)
 	}
 	// Call s.evaluate() to handle cases like "p1 -and p2"
 	parser.stack.evaluate()
--- a/cmd/internal/find/primary/meta/arrayPredicate.go
+++ b/cmd/internal/find/primary/meta/arrayPredicate.go
@@ -18,10 +18,11 @@ func parseArrayPredicate(tokens []string) (predicate.Predicate, []string, error)
 		return p, tokens, err
 	}
 	// EmptyPredicate did not match
+	parsePredicate := predicate.ToParser(parsePredicate)
 	return parseArrayP(
 		tokens,
-		predicate.ToParser(parsePredicate),
-		predicate.ToParser(parsePredicate),
+		parsePredicate,
+		parsePredicate,
 	)
 }
 
--- a/cmd/internal/find/primary/meta/arrayPredicate.go
+++ b/cmd/internal/find/primary/meta/arrayPredicate.go
@@ -6,17 +6,27 @@ import (
 	"strings"
 
 	"github.com/puppetlabs/wash/cmd/internal/find/parser/errz"
+	"github.com/puppetlabs/wash/cmd/internal/find/parser/predicate"
 )
 
 // ArrayPredicate => EmptyPredicate      |
 //                   ‘[' ? ‘]’ Predicate |
 //                   ‘[' * ‘]’ Predicate |
 //                   ‘[' N ‘]’ Predicate |
-func parseArrayPredicate(tokens []string) (Predicate, []string, error) {
+func parseArrayPredicate(tokens []string) (predicate.Predicate, []string, error) {
 	if p, tokens, err := parseEmptyPredicate(tokens); err == nil {
 		return p, tokens, err
 	}
 	// EmptyPredicate did not match
+	return parseArrayP(
+		tokens,
+		predicate.ToParser(parsePredicate),
+		predicate.ToParser(parsePredicate),
+	)
+}
+
+// This helper's used by parseArrayPredicate and parseArrayExpression
+func parseArrayP(tokens []string, baseCaseParser, keySequenceParser predicate.Parser) (predicate.Predicate, []string, error) {
 	if len(tokens) == 0 {
 		return nil, nil, errz.NewMatchError("expected an opening '['")
 	}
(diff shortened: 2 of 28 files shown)


cmd/internal/find/primary/meta_test.go
```diff
@@ -1,43 +1,103 @@
...
 
-	_, _, err = metaPrimary.parse([]string{"-m", "."})
-	suite.Regexp("-m: expected a key sequence after '.'", err)
+func (s *MetaPrimaryTestSuite) TestMetaPrimaryValidInputTruePredicates() {
+	s.RunTestCases(
+		s.NPTC("-m .architecture x86_64 -primary", "-primary", s.e),
+		s.NPTC("-m .blockDeviceMappings[?] .deviceName /dev/sda1 -primary", "-primary", s.e),
+		s.NPTC("-m .cpuOptions.coreCount 4 -primary", "-primary", s.e),
+		s.NPTC("-m .tags[?] .key termination_date -a .value +1h -primary", "-primary", s.e),
+		s.NPTC("-m .tags[?] .key foo -o .key department -primary", "-primary", s.e),
```
ekinanp: As a follow-up, I'd like to extend ObjectPredicate/ArrayPredicate to also support the `wash find` operators. That would reduce this test-case to
```
-m .tags[?] .key \( foo -o department \) -primary
```

(Parens are necessary to distinguish ObjectPredicate/ArrayPredicate expressions from the top-level meta primary expression).

ekinanp: May also be worth introducing some simple regex support for `meta` keys, because typing out e.g. `blockDeviceMappings` can be annoying, especially when you have to do it more than once. I'm thinking of something like `blockDev*[?] .deviceName /dev/sda1 -primary` where `blockDev*` means "key matching blockDev*".

ekinanp:
I recommend reviewing the updated grammar in `primary/meta.go` first.

Note that much of this PR's implementing negation semantics for the individual predicates (see `genericPredicate` in `meta/predicate.go`) + tests.

cmd/internal/find/primary/meta_test.go
ekinanp: With all the parens stuff in here, it might also not be a bad idea to have the expression parser support something like `\(foo` where it'd strip the parens off. Then the `department` test case could further be shortened to:
```
-m .tags[?] .key \(foo -o department\) -primary
```

cmd/internal/find/primary/meta/timePredicate.go
```diff
@@ -39,28 +40,45 @@ func parseTimePredicate(tokens []string) (Predicate, []string, error) {
...
+				// past. diff < 0 there means timeV is from the future, so the query
+				// doesn't make sense. Similarly, +{1h} means "more than one hour from now",
+				// which queries the future. diff < 0 there means timeV is from the past,
+				// so the query doesn't make sense. Thus, diff < 0 is a time-mismatch.
+				// Time-mismatches always return false.
+				return false
+			}
+			return p(diff)
+		},
+		subFromStartTime: true,
```
MikaelSmith: It looks like `subFromStartTime` is always true, so there's no reason to have it be a parameter of `timePredicate`.

cmd/internal/find/primary/meta_test.go
MikaelSmith: Those are interesting. I'm not against spending a little time on them, but probably don't want to spend a lot at the moment since you can reasonable accomplish your goal.

MikaelSmith: How does parsing distinguish between `-m .key foo -o .key bar` and `-m .key foo -o -m .key bar`? In other words, how does it know `-o` is part of the meta expression rather than after the meta expression?

P.S. If I read the answer earlier I may have just forgotten it and now can't find it.

cmd/internal/find/primary/numeric/predicate.go
```diff
@@ -25,6 +56,7 @@ func ParsePredicate(str string, parsers ...Parser) (Predicate, int, error) {
 		panic("numeric.ParsePredicate called without any parsers")
 	}
 
+	// TODO: Introduce "+="/"-=" to represent ">="/"<="?
```
MikaelSmith: We're using +/- primarily to avoid shell piping operators? I guess the symmetry makes sense then.

cmd/internal/find/primary/meta/arrayPredicate.go
```diff
@@ -6,17 +6,27 @@ import (
...
 //                   ‘[' N ‘]’ Predicate |
-func parseArrayPredicate(tokens []string) (Predicate, []string, error) {
+func parseArrayPredicate(tokens []string) (predicate.Predicate, []string, error) {
 	if p, tokens, err := parseEmptyPredicate(tokens); err == nil {
 		return p, tokens, err
 	}
 	// EmptyPredicate did not match
+	return parseArrayP(
+		tokens,
+		predicate.ToParser(parsePredicate),
```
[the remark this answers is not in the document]
ekinanp: Oh, I could extract that into a variable.

cmd/internal/find/primary/meta/predicateExpression_test.go
```diff
@@ -0,0 +1,219 @@
...
+		// Tests for -or
+		s.NPETC("-o", "-o: no expression before -o"),
+		s.NPETC("-true -o", "-o: no expression after -o"),
+		s.NPETC("-true -o -o", "-o: no expression after -o"),
+	)
+}
+
+func (s *PredicateExpressionTestSuite) TestAndOpEval() {
+	s.RunTestCases(
+		s.NPOETC("-true -a -false -primary", "-primary", false),
```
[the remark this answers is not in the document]
ekinanp: Sure.

cmd/internal/find/primary/meta/timePredicate.go
ekinanp: Oh, that was a mistake. Thanks for catching it. I'm surprised the tests didn't fail. Will fix.

cmd/internal/find/primary/meta_test.go
ekinanp: Don't think it should take too long, maybe a day.

And to answer your question, meta's expression parser stops if it encounters an unknown token. For your examples, `-m .key foo -o .key bar` will be parsed as `-m .key ( foo -o .key bar)`. Here, the predicate's on `meta["key"]` and it returns true if that value is the string "foo" OR if that value is an object with its "key" key set to the string "bar" (if that makes sense). The distinguishing factor there's the `KeySequence Expression` part in the grammar, where `.key` is the KeySequence and `foo -o .key bar` is the expression.

The second example  `-m .key foo -o -m .key bar` should be parsed by the top-level parser as `Meta(.key, foo) -o Meta(.key, bar)` where the `meta` primary will return an `UnknownTokenError` when it encounters the second `-m` token so that it hands it off to the top-level expression parser. However, thanks to https://github.com/puppetlabs/wash/blob/master/cmd/internal/find/parser/expression/parser.go#L184-L195, it currently does not do this -- it returns a `-m: no expression after -o` error so it is a bug. Will fix that.

cmd/internal/find/primary/numeric/predicate.go
ekinanp: I used +/- b/c BSD/GNU find uses them, but I'm guessing their reasoning for using them is as you said -- to avoid the shell piping operators.

I can add the symmetry as a follow-up.

cmd/internal/find/primary/meta/timePredicate.go
ekinanp: Nevermind, looks like it isn't needed. The `if diff < 0` check already handles time-mismatches. Will remove it.

cmd/internal/find/primary/meta/timePredicate_test.go
```diff
@@ -67,13 +67,22 @@ func (s *TimePredicateTestSuite) TestTimeP_Negation_NotATime() {
...
 	tp := timeP(true, func(n int64) bool {
-		return n > d 
+		return n > d
 	})
-	/// Negating tp still preserves time mismatches. 
 	s.False(tp.Negate().IsSatisfiedBy(addTST(d+1)))
 	s.False(tp.Negate().IsSatisfiedBy(addTST(d-1)))
+
+	// Test future queries
+	tp = timeP(false, func(n int64) bool {
```
ekinanp: Adding this test triggered the failure in the old code.

cmd/internal/find/parser/expression/parser.go
```diff
@@ -183,15 +183,22 @@ func (parser *Parser) Parse(tokens []string) (predicate.Predicate, []string, err
...
+				parser.stack.mostRecentOpToken,
+				parser.stack.mostRecentOpToken,
+			)
 		}
-		return nil, nil, fmt.Errorf(
-			"%v: no expression after %v",
-			parser.stack.mostRecentOpToken,
-			parser.stack.mostRecentOpToken,
-		)
+		// We have "p1 -and <unknown_token>". Pop the binary op off the stack and include
```
MikaelSmith: This ends up being a kind of two-look-ahead parser. Interesting set of special-casing for binary ops. I worry about other edge-cases we might be missing, but don't have any specifically in mind.

PR #244, opened 2019-05-09
Support predicate expressions in ObjectPredicate/ArrayPredicate

ekinanp:

### DIFF
--- a/cmd/internal/find/parser/expression/atom.go
+++ b/cmd/internal/find/parser/expression/atom.go
@@ -4,7 +4,7 @@ import "github.com/puppetlabs/wash/cmd/internal/find/parser/predicate"
 import "github.com/puppetlabs/wash/cmd/internal/find/parser/errz"
 import "fmt"
 
-func notOpParser(parser *Parser) predicate.Parser {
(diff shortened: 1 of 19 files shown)


ekinanp:
I'm open to additional test cases for the inner predicate expression stuff.

Also, think I'm going to hold off on the key regex support b/c with the changes here, I don't think people are likely going to type the same key more than once.

cmd/internal/find/parser/expression/atom.go
```diff
@@ -45,6 +45,18 @@ func parensOpParser(parser *Parser) predicate.Parser {
 			parser.stack = stack
 			parser.numOpenParens--
 		}()
-		return parser.Parse(tokens)
+		p, tokens, err := parser.Parse(tokens)
+		if err != nil && err != ErrEmptyExpression {
+			return p, tokens, err
+		}
+		// err == nil || err == ErrEmptyExpression
+		if len(tokens) == 0 || tokens[0] != ")" {
```
MikaelSmith: This check looks wrong. `tokens` is empty but then you try to index into it. That suggests a test case is also missing.

PR #245, opened 2019-05-09
Wash shell

MikaelSmith:


cmd/clear.go
```diff
@@ -19,7 +19,7 @@ func clearCommand() *cobra.Command {
 
 	clearCmd.Flags().BoolP("verbose", "v", false, "Print paths that were cleared from the cache")
 	if err := viper.BindPFlag("verbose", clearCmd.Flags().Lookup("verbose")); err != nil {
-		cmdutil.ErrPrintf("%v\n", err)
+		log.Print(err)
```
ekinanp: My concerns in https://github.com/puppetlabs/wash/pull/197#discussion_r274101636 still stand for this commit, unless it was accidentally introduced here?

cmd/server.go
```diff
@@ -80,27 +80,44 @@ func serverMain(cmd *cobra.Command, args []string) exitCode {
...
+		return exitCode{1}
+	}
+	return exitCode{0}
+}
+
+type serverOpts struct {
+	cpuprofile, externalPluginsPath string
+}
+
+func runServer(mountpoint string, readyCh chan struct{}, sigCh chan os.Signal, opts serverOpts) error {
```
ekinanp: This indicates that it might be a good time to introduce a `plugin.Server` object to encapsulate some of this code. Then you can have something like `server.Start() readyCh chan error`, `server.Stop() stoppedCh chan error`. Thinking `plugin.Server` could replace `plugin.Registry`, so you can also do `server.RegisterCorePlugin()`, `server.RegisterExternalPlugin()`, both of which read well. These changes would also make it easier for third parties to use the Wash server if they need to.

cmd/shell.go
```diff
@@ -0,0 +1,109 @@
...
+
+	readyCh := make(chan struct{}, 1)
+	sigCh := make(chan os.Signal, 1)
+	doneCh := make(chan struct{}, 1)
+	go func() {
+		// Configure logrus to omit simple text
+		logrus.SetLevel(logrus.WarnLevel)
+		logrus.SetFormatter(&logrus.TextFormatter{DisableTimestamp: true})
+
+		// TODO: handle multiple active shells using a single server
```
ekinanp: Not sure what you mean by this TODO.

cmd/shell.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	// Add wash to PATH, set PS1. These are useful fallbacks in case PATH and prompt aren't setup.
+	washPath, err := os.Executable()
+	if err != nil {
+		return cleanup(err)
+	}
+	washDir := filepath.Dir(washPath)
+	pathEnv := os.Getenv("PATH")
+	if err := os.Setenv("PATH", pathEnv+string(os.PathListSeparator)+washDir); err != nil {
```
ekinanp: Are we still planning on aliasing the Wash commands? It's a bit annoying having a `wash$` prompt but still having to type `wash ls`, `wash find`, etc. to do Wash-specific stuff.

cmd/shell.go
ekinanp: Also, how can I access my server logs?

cmd/clear.go
MikaelSmith: Oh right. I can add something. I think I had some theory of unifying logging, but I think I ended up just customizing logrus to be similar instead.

cmd/server.go
MikaelSmith: I don't think it should be part of plugin (for one, that introduces a circular dependency between fuse and plugin), but an object seems to make sense.

cmd/shell.go
MikaelSmith: The server logs don't exist. They're written to stdout, or are part of activity journals.

cmd/shell.go
MikaelSmith: Probably. I don't think LS is useful enough to be the default yet though.

cmd/server.go
ekinanp: Agreed. I think `api/server.go` could contain the object, and we could move the FUSE package over to an internal package in the API since I don't think that's something we want people to use.

`api.NewServer()` also reads well.

cmd/shell.go
ekinanp: So if I want to debug something, I'd need to use the old workflow of starting my server in one shell, then doing stuff in another shell?

MikaelSmith: No. Logging still goes to the console. I haven't added a flag to mess with verbosity yet.

ekinanp: That could cause interleaved log/shell output, which might be confusing. Maybe include the logfile as an option too?

ekinanp:
Oh, since we're now requiring Go 1.12+, it'd also be worth cleaning up https://github.com/puppetlabs/wash/blob/master/plugin/helpers.go#L140-L146

cmd/server.go
MikaelSmith: I started this with in an independent internal package for now to give us a little more flexibility. I'm not sure it's generally useful yet.

cmd/shell.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"github.com/puppetlabs/wash/cmd/internal/server"
+	cmdutil "github.com/puppetlabs/wash/cmd/util"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func writeAlias(path, subcommand string) error {
+	f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0750)
```
ekinanp: Oh I see, you're creating a script that just invokes the subcommand. A quick comment explaining that would be nice.

cmd/internal/server/core.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/puppetlabs/wash/plugin"
+	"github.com/puppetlabs/wash/plugin/aws"
+	"github.com/puppetlabs/wash/plugin/docker"
+	"github.com/puppetlabs/wash/plugin/kubernetes"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// Opts exposes additional configuration for server operation.
+type Opts struct {
```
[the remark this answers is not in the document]
ekinanp: Should we include log config? That also seems like a part of starting the server.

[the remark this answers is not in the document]
MikaelSmith: "additional". `mountpoint` is required, while the rest can safely be left empty.

Including log config makes sense to me though.

cmd/internal/server/core.go
```diff
@@ -0,0 +1,192 @@
...
+	mountpoint string
+	registry   *plugin.Registry
+	opts       Opts
+	api        controlChannels
+	fuse       controlChannels
+}
+
+// New creates a new Server.
+func New(mountpoint string, opts Opts) (*Server, error) {
+	registry := plugin.NewRegistry()
```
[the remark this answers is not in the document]
MikaelSmith: I was going off the principal of checking for errors earlier, but that's also a reasonable breakdown. Having `New` return an error does feel weird to me, so I guess I'll change it up.

cmd/internal/server/core.go
```diff
@@ -0,0 +1,192 @@
...
+
+	fuseServerStopCh, fuseServerStoppedCh, err := fuse.ServeFuseFS(s.registry, s.mountpoint)
+	if err != nil {
+		s.stopAPIServer()
+		return err
+	}
+	s.fuse = controlChannels{stopCh: fuseServerStopCh, stoppedCh: fuseServerStoppedCh}
+
+	if s.opts.Cpuprofile != "" {
+		f, err := os.Create(s.opts.Cpuprofile)
```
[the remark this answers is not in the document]
MikaelSmith: k

cmd/internal/server/core.go
```diff
@@ -0,0 +1,192 @@
...
+		pprof.StopCPUProfile()
+	}
+
+	// Close any open journals on shutdown to ensure remaining entries are flushed to disk.
+	activity.CloseAll()
+}
+
+// Wait blocks until the server exits due to an error or SIGINT/SIGTERM.
+// Only one of Wait or Stop should be called.
+func (s *Server) Wait() {
```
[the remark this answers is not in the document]
MikaelSmith: That would result in blocking even if there's a failure and the server stops running. I am amenable to passing in the signal channel (thus making it more configurable), but I think the `Wait` semantics still make sense.

cmd/root.go
```diff
@@ -37,7 +43,8 @@ func rootCommand() *cobra.Command {
 		// TODO: Set this to "" when we're ready to ship so that
 		// when we alias our custom commands, someone typing in
 		// e.g. `meta --help` will not see `wash meta` in the usage
-		Use: "wash",
+		Use:  "wash",
+		RunE: toRunE(shellMain),
```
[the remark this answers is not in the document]
MikaelSmith: I'm not sure that actually makes the code any clearer. Root does other things too.

cmd/internal/server/core.go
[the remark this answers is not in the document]
ekinanp: That's a good point. Ok, I think `Wait` should take the signal channel then.

cmd/shell.go
```diff
@@ -0,0 +1,143 @@
...
+	if err := os.Symlink(washPath, newWashPath); err != nil {
+		cmdutil.ErrPrintf("Error linking wash executable to %v: %v\n", newWashPath, err)
+		return exitCode{1}
+	}
+	// Executable file can't override shell built-ins, so use wexec instead of exec.
+	// List also isn't very feature complete so we don't override ls.
+	// These are executables instead of aliases because putting alias declarations at the beginning
+	// of stdin for the command doesn't work right.
+	aliases := map[string]string{
+		"wexec": "exec",
```
MikaelSmith: This should probably include history.

cmd/shell.go
```diff
@@ -0,0 +1,143 @@
...
+	// TODO: instead of running a server in-process, can we start one in a separate process that can
+	//       be shared between multiple invocations of `wash`?
+	srv := server.New(mountpath, level, server.Opts{LogFile: logfile})
+	if err := srv.Start(); err != nil {
+		cmdutil.ErrPrintf("Unable to start server: %v\n", err)
+		return exitCode{1}
+	}
+
+	// TODO: expand Welcome to Wash spiel
+	fmt.Println("Welcome to Wash!")
```
MikaelSmith: This should introduce history, and mention find/wexec/wls.

cmd/shell.go
ekinanp: What about tail?

cmd/root.go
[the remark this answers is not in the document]
ekinanp: Could we rename `shell.go` to `rootMain.go` then? That would still enforce the "Wash subcommands are all of the non-root *.go files in the cmd/" invariant.

cmd/internal/server/core.go
```diff
@@ -0,0 +1,192 @@
...
+	<-s.fuse.stoppedCh
+}
+
+func (s *Server) shutdown() {
+	if s.opts.Cpuprofile != "" {
+		pprof.StopCPUProfile()
+	}
+
+	// Close any open journals on shutdown to ensure remaining entries are flushed to disk.
+	activity.CloseAll()
```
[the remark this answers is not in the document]
MikaelSmith: This feels like bike-shedding and unrelated to this PR. Not going to touch it right now.

MikaelSmith:
Updated.

cmd/root.go
```diff
@@ -79,3 +92,24 @@ func Execute() int {
 
 	return exitCode.value
 }
+
+var levelMap = map[string]log.Level{
```
ekinanp: Should we move this logging stuff over to the `cmdutil` package?

cmd/internal/server/core.go
```diff
@@ -0,0 +1,203 @@
...
+	mountpoint string
+	level      log.Level
+	opts       Opts
+	logFH      *os.File
+	api        controlChannels
+	fuse       controlChannels
+}
+
+// New creates a new Server.
+func New(mountpoint string, level log.Level, opts Opts) *Server {
```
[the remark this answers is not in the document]
MikaelSmith: Mostly didn't want the default in two places, but sure.

cmd/root.go
MikaelSmith: Done.

PR #246, opened 2019-05-10
 Default to maxdepth 1 if the meta primary is set 

ekinanp:
The meta primary's a specialized filter. This means that for a
given entry, it is unlikely that the entry's siblings or children
will have the same meta schema. Thus, it does not make sense for
`wash find` to recurse when the user specifies the meta primary
because recursing could lead to wasteful API requests.

Thus to avoid the unnecessary recursion, this commit sets the maxdepth
option to 1 if the meta primary's specified and the flag was not
provided on the command line.

### DIFF
--- a/cmd/internal/find/primary/meta.go
+++ b/cmd/internal/find/primary/meta.go
@@ -1,6 +1,9 @@
 package primary
 
 import (
+    "fmt"
+    "os"
+
     "github.com/puppetlabs/wash/cmd/internal/find/types"
     "github.com/puppetlabs/wash/cmd/internal/find/primary/meta"
 )
@@ -71,9 +74,10 @@ var metaPrimary = Parser.add(&primary{
             // because those entries don't have tags and, even if they did, they'd likely be under a
             // different key (e.g. like "Labels" for Docker containers). Thus to avoid the unnecessary
             // recursion, we default maxdepth to 1 if the flag was not set by the user. Note that users
-            // who want to recurse down into subdirectories can just specify the maxdepth option, which
+            // who want to recurse down into subdirectories can just set maxdepth to -1. The recursion
             // is useful when running `wash find` inside a directory whose entries and subdirectory entries
             // all have the same `meta` schema (e.g. like in an S3 bucket).
+            fmt.Fprintln(os.Stderr, "The meta primary is being used. Setting maxdepth to 1...")
             opts.Maxdepth = 1
         }
     },
--- a/cmd/internal/find/primary/meta.go
+++ b/cmd/internal/find/primary/meta.go
@@ -53,4 +53,7 @@ StringPredicate     => [^-].*
 N                   => \d+ (i.e. some number > 0)
 */
 //nolint
-var metaPrimary = Parser.newPrimary([]string{"-meta", "-m"}, meta.Parse)
+var metaPrimary = Parser.add(&primary{
+    tokens: []string{"-meta", "-m"},
+    parseFunc: meta.Parse,
+})
--- a/cmd/internal/find/primary/meta.go
+++ b/cmd/internal/find/primary/meta.go
@@ -1,6 +1,9 @@
 package primary
 
-import "github.com/puppetlabs/wash/cmd/internal/find/primary/meta"
+import (
+    "github.com/puppetlabs/wash/cmd/internal/find/types"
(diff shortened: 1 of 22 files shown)


cmd/internal/find/primary/meta.go
```diff
@@ -56,4 +59,22 @@ N                   => \d+ (i.e. some number > 0)
...
+            // primary since it is likely that siblings or children will have a different meta
+            // schema. For example, if we're filtering EC2 instances based on a tag, then `wash find`
+            // shouldn't recurse down into the EC2 instance's console output + metadata.json files
+            // because those entries don't have tags and, even if they did, they'd likely be under a
+            // different key (e.g. like "Labels" for Docker containers). Thus to avoid the unnecessary
+            // recursion, we default maxdepth to 1 if the flag was not set by the user. Note that users
+            // who want to recurse down into subdirectories can just specify the maxdepth option, which
+            // is useful when running `wash find` inside a directory whose entries and subdirectory entries
+            // all have the same `meta` schema (e.g. like in an S3 bucket).
+            opts.Maxdepth = 1
```
ekinanp: I'm considering whether it's worth allowing something like `wash find -maxdepth` where a no-value `maxdepth` means "recurse." This is useful for people who want to use the `meta` primary and also want `wash find` to recurse all the way down (e.g. it makes the S3 bucket case less annoying). However, implementing this feature would require some extra, not-very-simple code b/c Go's `flag` package requires `int` options to have a value. Thus to avoid the extra code, I wonder if we should add a separate `recurse` option instead and document that the `recurse` option should be used in conjunction with the `meta` primary.

cmd/internal/find/primary/meta.go
```diff
@@ -56,4 +59,22 @@ N                   => \d+ (i.e. some number > 0)
 var metaPrimary = Parser.add(&primary{
     tokens: []string{"-meta", "-m"},
     parseFunc: meta.Parse,
+    optionsSetter: func(opts *types.Options) {
```
ekinanp: This capability will be useful for other primaries as well, like e.g. `delete` where for `delete`, the `depth` option must be set (otherwise, the parent would be deleted before the children).

cmd/internal/find/primary/meta.go
ekinanp: Also wondering if I should print something to stderr here so the user knows about this magic.

cmd/internal/find/types/options.go
```diff
@@ -18,9 +19,30 @@ func NewOptions() Options {
 		Depth:    false,
 		Mindepth: 0,
 		Maxdepth: ^uint(0),
+		setFlags: make(map[string]struct{}),
 	}
 }
 
+// DepthFlag is the name of the depth option's flag
+var DepthFlag = "depth"
```
MikaelSmith: These could be `const`.

cmd/internal/find/primary/meta.go
MikaelSmith: Check for `∞`? I guess the simplest is probably `-maxdepth -1`. Having a separate option feels like it makes it harder to figure out what's happening, but the behavior you're adding does seem useful.

ekinanp: I'll go with `-maxdepth -1` -- that makes more sense.

cmd/internal/find/primary/meta.go
```diff
@@ -71,9 +74,10 @@ var metaPrimary = Parser.add(&primary{
             // because those entries don't have tags and, even if they did, they'd likely be under a
             // different key (e.g. like "Labels" for Docker containers). Thus to avoid the unnecessary
             // recursion, we default maxdepth to 1 if the flag was not set by the user. Note that users
-            // who want to recurse down into subdirectories can just specify the maxdepth option, which
+            // who want to recurse down into subdirectories can just set maxdepth to -1. The recursion
             // is useful when running `wash find` inside a directory whose entries and subdirectory entries
             // all have the same `meta` schema (e.g. like in an S3 bucket).
+            fmt.Fprintln(os.Stderr, "The meta primary is being used. Setting maxdepth to 1...")
```
ekinanp: Is this message good enough?

MikaelSmith: Seems ok.

PR #256, opened 2019-05-15
Change plugin.ExecResult => plugin.RunningCommand

ekinanp:
plugin.RunningCommand better reflects how we've been using plugin.ExecResult. It also removes some cleanup burden from the plugin author, from clients of plugin.RunningCommand, and also simplifies the plugin interface by removing the ExitCodeCB so that the implementations better reflect the flow of an exec'ing command.

### DIFF
--- a/plugin/docker/container.go
+++ b/plugin/docker/container.go
@@ -56,9 +56,7 @@ func (c *container) List(ctx context.Context) ([]plugin.Entry, error) {
 	return []plugin.Entry{cm, clf, vol.NewFS("fs", c)}, nil
 }
 
-func (c *container) Exec(ctx context.Context, cmd string, args []string, opts plugin.ExecOptions) (plugin.ExecCommand, error) {
-	execCommand := plugin.ExecCommand{}
-
+func (c *container) Exec(ctx context.Context, cmd string, args []string, opts plugin.ExecOptions) (*plugin.ExecCommand, error) {
 	command := append([]string{cmd}, args...)
 	activity.Record(ctx, "Exec %v on %v", command, c.Name())
 
@@ -68,21 +66,21 @@ func (c *container) Exec(ctx context.Context, cmd string, args []string, opts pl
 	}
 	created, err := c.client.ContainerExecCreate(ctx, c.id, cfg)
 	if err != nil {
-		return execCommand, err
+		return nil, err
 	}
 
 	resp, err := c.client.ContainerExecAttach(ctx, created.ID, types.ExecStartCheck{})
 	if err != nil {
-		return execCommand, err
+		return nil, err
 	}
 
-	// Asynchronously copy container exec output to an exec output channel.
-	outputCh, stdout, stderr := plugin.CreateExecOutputStreams(ctx)
+	execCommand := plugin.NewExecCommand(ctx)
+
+	// Asynchronously copy container exec output
 	go func() {
-		_, err := stdcopy.StdCopy(stdout, stderr, resp.Reader)
+		_, err := stdcopy.StdCopy(execCommand.Stdout(), execCommand.Stderr(), resp.Reader)
 		activity.Record(ctx, "Exec on %v complete: %v", c.Name(), err)
-		stdout.CloseWithError(err)
-		stderr.CloseWithError(err)
+		execCommand.CloseStreamsWithError(err)
 		resp.Close()
 	}()
 
@@ -101,7 +99,7 @@ func (c *container) Exec(ctx context.Context, cmd string, args []string, opts pl
 		}()
 	}
 
-	execCommand.OutputCh = outputCh
+
 	execCommand.StopFunc = func() {
 		// Close the response on cancellation. Copying will block until there's more to read from the
 		// exec output. For an action with no more output it may never return.
--- a/plugin/docker/container.go
+++ b/plugin/docker/container.go
@@ -100,7 +100,7 @@ func (c *container) Exec(ctx context.Context, cmd string, args []string, opts pl
 	}
 
 
-	execCommand.StopFunc = func() {
+	execCommand.SetStopFunc(func() {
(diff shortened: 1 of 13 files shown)


plugin/docker/container.go
```diff
@@ -68,29 +66,30 @@ func (c *container) Exec(ctx context.Context, cmd string, args []string, opts pl
...
 
-	// Asynchronously copy container exec output to an exec output channel.
-	outputCh, stdout, stderr := plugin.CreateExecOutputStreams(ctx)
-	go func() {
-		_, err := stdcopy.StdCopy(stdout, stderr, resp.Reader)
-		activity.Record(ctx, "Exec on %v complete: %v", c.Name(), err)
-		stdout.CloseWithError(err)
-		stderr.CloseWithError(err)
+	cmdObj := plugin.NewRunningCommand(ctx)
+	cmdObj.SetStopFunc(func() {
```
ekinanp: I figured moving this closer to `cmdObj` would make the "This command is currently running, here's how we stop it" part more obvious. Can move it back if it just makes things more confusing.

plugin/docker/container.go
```diff
@@ -101,34 +100,26 @@ func (c *container) Exec(ctx context.Context, cmd string, args []string, opts pl
...
-	execResult.CancelFunc = func() {
-		// Close the response on cancellation. Copying will block until there's more to read from the
-		// exec output. For an action with no more output it may never return.
-		if opts.Tty {
-			// If resp.Conn is still open, send Ctrl-C over resp.Conn before closing it.
-			_, err := resp.Conn.Write([]byte{0x03})
-			activity.Record(ctx, "Sent ETX on context termination: %v", err)
-		}
+	// Asynchronously copy container exec output, then send over the exit code
+	go func() {
```
ekinanp: I think this reads better now without the `ExitCodeCB`. Same with `ssh.go`.

plugin/outputStream.go
```diff
@@ -41,13 +41,13 @@ func (s *OutputStream) Write(data []byte) (int, error) {
 	return len(data), err
 }
 
-// Close ensures the channel is closed when the last OutputStream is closed.
-func (s *OutputStream) Close() {
+// close ensures the channel is closed when the last OutputStream is closed.
+func (s *OutputStream) close() {
```
ekinanp: Output streams should be closed by calling methods of the `RunningCommand` object, so I don't see a good reason to continue exporting these methods.

plugin/runningCommand.go
```diff
@@ -0,0 +1,157 @@
...
+	mux           sync.Mutex
+	streamsClosed bool
+	exitCodeCh    chan int
+	exitCodeErrCh chan error
+}
+
+// NewRunningCommand creates a new RunningCommand object that's tied to
+// the passed-in execution context. When the execution context is cancelled,
+// the returned command's stdout/stderr streams will automatically be closed.
+// This means that you do not have to worry about handling dangling output
```
ekinanp: Handling the cleanup here happened to fix a [bug in ssh.go](https://github.com/puppetlabs/wash/blob/master/transport/ssh.go#L144-L153) where the output streams were not closed prior to exiting the function.

plugin/runningCommand.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// CloseStreams closes the command's stdout and stderr streams with
+// the specified error. This should be called with err == nil when the
+// command's finished its execution, or when an IO error occurs.
+//
+// NOTE: CloseStreams is automatically called whenever the execution
+// context is cancelled, so you do not have to worry about handling
+// dangling output streams.
+func (cmd *RunningCommand) CloseStreams(err error) {
```
ekinanp: One reason why I removed `CloseStreamsWithError`and just stuck it to `CloseStreams`'s interface is to centralize the "we handle the clean up for you, so you only worry about closing these streams once your plugin API's finished executing your command" comment.

transport/ssh.go
```diff
@@ -119,68 +117,61 @@ func ExecSSH(ctx context.Context, id Identity, cmd []string, opts plugin.ExecOpt
...
-	go func() {
-		exitErr = session.Wait()
-		activity.Record(ctx, "Closing session for %v: %v", id.Host, session.Close())
-		stdout.Close()
-		stderr.Close()
-	}()
-
-	result.OutputCh = outputch
-	result.CancelFunc = func() {
+	cmdObj.SetStopFunc(func() {
```
ekinanp: Moving this closer to `Start` makes me think of this as "Ok, the command's started so here we're setting the stop function". Can move back if it makes things more confusing.

plugin/runningCommand.go
```diff
@@ -0,0 +1,156 @@
...
+	cmd.Stdout().closeWithError(err)
+	cmd.Stderr().closeWithError(err)
+	cmd.streamsClosed = true
+}
+
+// Wait waits for the command to finish, passing in each chunk
+// of the command's output to processChunk. It returns the command's
+// exit code, or an error if it failed to fetch the command's exit
+// code.
+func (cmd *RunningCommand) Wait(processChunk func(ExecOutputChunk)) (int, error) {
```
ekinanp: I wonder if `Wait` is doing too much. I can split this up to `StreamOutput() <-chan Exec` (which just returns outputCh), and put the exit-code stuff in a `cmd.ExitCode()` method.

ekinanp: Hmm, I think I'll go ahead and do that. `Wait` with a callback is kind of weird.

plugin/runningCommand.go
ekinanp: Oh, nevermind. According to https://stackoverflow.com/questions/8593645/is-it-ok-to-leave-a-channel-open, channels can be left open. I'll update the PR to simplify some of the cleanup code.

ekinanp:
I've significantly deviated from this implementation, so going to close this PR and open a new one instead.

PR #257, opened 2019-05-16
Refactor Exec's plugin interface

ekinanp:
This commit refactors Exec's plugin interface to be more plugin-author
friendly, and simpler to understand. It includes the following changes:

* Rename plugin.ExecResult => plugin.ExecCommand

plugin.ExecCommand better reflects what is being returned.

* Enforce a strict separation between what an Exec caller needs to care
about (plugin.ExecCommand) vs. what a plugin author needs to care about
(plugin.ExecCommandImpl) when implementing Exec

Exec callers should only care about a command's output channel and exit
code. They should not have to worry about context-cancellation cleanup
like ensuring that the command's stopped. That cleanup is the plugin
author's responsibility. On the other hand, it would be annoying for a
plugin author to have to remember to handle context-cancellation
cleanup.

The plugin.ExecCommand interface and the plugin.ExecCommandImpl object
solve both of these problems. Execable#Exec returns a plugin.ExecCommand
object, which only includes OutputCh and ExitCode. These are precisely the
things that Exec callers need to care about. Conversely, the
plugin.ExecCommandImpl implements the plugin.ExecCommand interface, but
also includes additional helpers that are meant to help plugin authors
write their code, including helpers that handle context cancellation
cleanup.

Finally, plugin.ExecCommandImpl replaces the previous ExitCodeCB via
helpers that set the exit-code/exit-code errors. This is more idiomatic
with how one thinks about what exec'ing a command would look like
("Stream the output" => "Get/Set the exit code").

### DIFF
--- a/plugin/execCommand.go
+++ b/plugin/execCommand.go
@@ -1,0 +1,128 @@
+package plugin
+
+import (
+	"context"
+	"fmt"
+)
+
+// ExecCommandImpl implements the plugin.ExecCommand interface.
+// Use plugin.NewExecCommand to create instances of these objects.
+type ExecCommandImpl struct {
+	ctx           context.Context
+	outputCh      chan ExecOutputChunk
+	stdout        *OutputStream
+	stderr        *OutputStream
+	exitCodeCh    chan int
+	exitCodeErrCh chan error
+}
+
+// NewExecCommand creates a new ExecCommandImpl object whose
+// lifetime is tied to the passed-in execution context.
+func NewExecCommand(ctx context.Context) *ExecCommandImpl {
+	cmd := &ExecCommandImpl{
+		ctx: ctx,
+		exitCodeCh: make(chan int, 1),
+		exitCodeErrCh: make(chan error, 1),
+	}
+
+	// Create the output streams
+	cmd.outputCh = make(chan ExecOutputChunk)
+	closer := &multiCloser{ch: cmd.outputCh, countdown: 2}
+	cmd.stdout = &OutputStream{ctx: cmd.ctx, id: Stdout, ch: cmd.outputCh, closer: closer}
+	cmd.stderr = &OutputStream{ctx: cmd.ctx, id: Stderr, ch: cmd.outputCh, closer: closer}
+
+	// Ensure that the output streams are closed when the context
+	// is cancelled. This guarantees that callers won't be blocked
+	// when they are streaming our command's output.
+	go func() {
+		<-cmd.ctx.Done()
+		cmd.CloseStreamsWithError(ctx.Err())
+	}()
+
+	return cmd
+}
+
+// SetStopFunc sets the function that stops the running command. stopFunc
+// is called when the execution context completes to perform necessary
+// termination.
+func (cmd *ExecCommandImpl) SetStopFunc(stopFunc func()) {
+	// Thankfully, goroutines are cheap. Otherwise, mixing this in with
+	// the goroutine in NewExecCommand heavily complicates things.
+	// For example, we'd have to worry about the possibility that the
+	// NewExecCommand goroutine is invoked before the Execable#Exec
+	// implementation can set a stopFunc, which can happen if the context
+	// is prematurely cancelled. That can result in an orphaned process
+	// in some plugin APIs, which is bad.
+	if stopFunc != nil {
+		go func() {
+			<-cmd.ctx.Done()
+			stopFunc()
+		}()
+	}
+}
+
+// Stdout returns the command's stdout stream. Attach this to your
+// plugin API's stdout stream.
+func (cmd *ExecCommandImpl) Stdout() *OutputStream {
+	return cmd.stdout
+}
+
+// Stderr returns the command's stderr stream. Attach this to your
+// plugin API's stderr stream.
+func (cmd *ExecCommandImpl) Stderr() *OutputStream {
+	return cmd.stderr
+}
+
+// CloseStreamsWithError closes the command's stdout/stderr streams
+// with the given error.
+func (cmd *ExecCommandImpl) CloseStreamsWithError(err error) {
+	cmd.Stdout().CloseWithError(err)
+	cmd.Stderr().CloseWithError(err)
+}
+
+// SetExitCode sets the command's exit code.
+func (cmd *ExecCommandImpl) SetExitCode(exitCode int) {
+	select {
+	case <-cmd.ctx.Done():
+		// Don't send anything if the context is cancelled.
+	default:
+		cmd.exitCodeCh <- exitCode
+	}
+}
+
+// SetExitCodeErr sets the exit code error, which occurs when the
+// plugin API fails to fetch the comand's exit code.
+//
+// NOTE: You should only use this function if your plugin API requires
+// a separate request to fetch the command's exit code. Otherwise,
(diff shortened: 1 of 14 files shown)


plugin/outputStream.go
```diff
@@ -41,22 +43,27 @@ func (s *OutputStream) Write(data []byte) (int, error) {
...
 
-// Close ensures the channel is closed when the last OutputStream is closed.
-func (s *OutputStream) Close() {
-	s.closer.Close()
-}
-
-// CloseWithError sends the given error before calling Close()
+// CloseWithError sends the given error before closing the OutputStream.
+// It will noop if the OutputStream's already closed.
 func (s *OutputStream) CloseWithError(err error) {
```
ekinanp: I erased `Close` because I think it's simpler to expose a single method for closing an OutputStream. I also don't see a good reason why the OutputStream object should implement the `io.WriteCloser` interface when it just needs to implement `io.Writer` in order to do its job. Similar reasons explain why there is a `CloseStreamsWithError` method but not a `CloseStreams` method in `ExecCommandImpl`.

If there are good reasons for us to support the `Close` method in OutputStream, I'm happy to add it back.

plugin/outputStream_test.go
```diff
@@ -110,27 +109,21 @@ func (suite *OutputStreamTestSuite) assertClosedChannel(ch <-chan ExecOutputChun
...
 
-func (suite *OutputStreamTestSuite) TestClose() {
-	ch := make(chan ExecOutputChunk)
-	stream := OutputStream{ch: ch, closer: &multiCloser{ch: ch, countdown: 1}}
-
-	stream.Close()
-	suite.assertClosedChannel(stream.ch)
-}
-
 func (suite *OutputStreamTestSuite) TestCloseWithError() {
```
ekinanp: I should eventually clean up some of these tests.

plugin/execCommand.go
```diff
@@ -0,0 +1,128 @@
...
+	// Create the output streams
+	cmd.outputCh = make(chan ExecOutputChunk)
+	closer := &multiCloser{ch: cmd.outputCh, countdown: 2}
+	cmd.stdout = &OutputStream{ctx: cmd.ctx, id: Stdout, ch: cmd.outputCh, closer: closer}
+	cmd.stderr = &OutputStream{ctx: cmd.ctx, id: Stderr, ch: cmd.outputCh, closer: closer}
+
+	// Ensure that the output streams are closed when the context
+	// is cancelled. This guarantees that callers won't be blocked
+	// when they are streaming our command's output.
+	go func() {
```
ekinanp: Note that this cleanup cleanly handles edge-cases like a failing Stop call leaving dangling `Exec`-related goroutines (b/c the command would still be running, so the output streams would still be open and hence the caller blocked).

It might be worthwhile (in the future) to add some retry logic to `SetStopFunc` in case a Stop fails due to some connection error, though I don't know how likely that scenario is (or if it's even possible).

plugin/outputStream.go
ekinanp: Also, exporting this is still a good idea in case the plugin API has disjoint Stdout/Stderr streams (e.g. like in the previous SSM Exec implementation). This way, plugin authors can still capture stream-specific failures.

plugin/outputStream_test.go
```diff
@@ -151,28 +144,37 @@ func (suite *OutputStreamTestSuite) TestCloseWithError() {
...
 	// Now, test that if err == ctx.Err(), then ctx.Err() is _not_ sent if
 	// a previous Write sent it
 	stream = newOutputStream(ctx)
 	stream.sentCtxErr = true
+	suite.False(stream.closed)
 	stream.CloseWithError(ctx.Err())
 	suite.assertClosedChannel(stream.ch)
+	suite.True(stream.closed)
 
+	// TODO: Add test to ensure that consecutive closes noop
```
MikaelSmith: I'd rather not leave these TODOs around.

plugin/types.go
```diff
@@ -104,25 +104,26 @@ type ExecOutputChunk struct {
...
+// OutputCh returns a channel containing timestamped chunks of the command's
+// stdout/stderr.
+//
+// ExitCode returns the command's exit code. It will block until the command's
+// exit code is set, or until the execution context is cancelled. ExitCode will
+// return an error if it fails to fetch the command's exit code.
+type ExecCommand interface {
+	OutputCh() <-chan ExecOutputChunk
+	ExitCode() (int, error)
+	sealed()
```
MikaelSmith: This seems unnecessarily restrictive. I'm not sure we need to stop anyone from writing a different ExecCommand implementation. But I'm ok waiting until someone asks for it.

plugin/execCommand.go
```diff
@@ -0,0 +1,128 @@
...
+	ctx           context.Context
+	outputCh      chan ExecOutputChunk
+	stdout        *OutputStream
+	stderr        *OutputStream
+	exitCodeCh    chan int
+	exitCodeErrCh chan error
+}
+
+// NewExecCommand creates a new ExecCommandImpl object whose
+// lifetime is tied to the passed-in execution context.
```
MikaelSmith: We should document what other functions you need to call to successfully use ExecCommandImpl.

plugin/outputStream_test.go
ekinanp: I can address them. `ExecCommand` could use some tests anyways.

ekinanp:
Updated with tests and additional documentation for `ExecCommandImpl`.

plugin/execCommandImpl_test.go
```diff
@@ -0,0 +1,182 @@
...
+	// Ensure that the context is cancelled to avoid dangling
+	// goroutines
+	suite.cancelFunc()
+}
+
+func (suite *ExecCommandImplTestSuite) NewExecCommand() *ExecCommandImpl {
+	return NewExecCommand(suite.ctx)
+}
+
+func (suite *ExecCommandImplTestSuite) TestNewExecCommand_CreatesOutputStreams() {
```
ekinanp: This previously tested `CreateExecOutputStreams`.

plugin/outputStream.go
MikaelSmith: I can't think of a reason we need to keep the Closer functionality.

plugin/outputStream_test.go
```diff
@@ -109,68 +119,72 @@ func (suite *OutputStreamTestSuite) assertClosedChannel(ch <-chan ExecOutputChun
...
-	sentErr := fmt.Errorf("an arbitrary error")
-	suite.False(stream.closed)
-	stream.CloseWithError(sentErr)
-	assertSentError(stream, sentErr)
-	suite.assertClosedChannel(stream.ch)
-	suite.True(stream.closed)
+func (suite *OutputStreamTestSuite) TestCloseWithError_NoError() {
+	stream := newOutputStream(context.Background())
+	stream.CloseWithError(nil)
+	suite.assertClosed(stream)
```
MikaelSmith: A negative assertion would make sense here.

plugin/types.go
ekinanp: Yeah, let's wait for it. I'd like to enforce the "This is the way we want you to do it" thing at the type-level since it makes it easier for people to figure out how they should start implementing Exec.

PR #258, opened 2019-05-16
Add ability to follow Wash history

MikaelSmith:
Adds `-f,--follow` option to the history command so you can watch for new entries added to history or follow updates to a journal for an in-progress command. This is particularly useful when a command has been running for awhile and you want to see a more detailed look at what it's doing.


api/helpers.go
```diff
@@ -159,3 +161,17 @@ func getEntryFromRequest(r *http.Request) (plugin.Entry, string, *errorResponse)
 	entry, err := findEntry(ctx, root, segments)
 	return entry, path, err
 }
+
+func getBoolParam(u *url.URL, key string) (bool, *errorResponse) {
+	vals := u.Query()[key]
+	if len(vals) > 0 {
+		// Take last value
```
ekinanp: A `getParam`-type helper would be useful to centralize the "we take the last value of a non-array query param" concept.

api/history.go
```diff
@@ -29,18 +40,59 @@ import (
...
+		return err
+	}
 
-	w.WriteHeader(http.StatusOK)
+	if follow {
+		last := len(history)
+		for {
+			// Continue sending updates
+			select {
+			case <-r.Context().Done():
```
ekinanp: Since we're starting to have multiple streamed responses now (with stream, exec, history, and historyEntry/journal), I think we should extend the streamableResponseWriter to centralize context-cancellation related cleanup (+ the logging related to that).

MikaelSmith: I'm not sure they're directly related, but something to provide a common cancellation pattern makes sense. This starts to conflict with your PR though.

ekinanp: I was thinking of something like wrapping streamableResponseWriter#Write to invoke a cleanup function upon context cancellation (and also outputting the log message). That shouldn't conflict with my PR.

MikaelSmith: I don't think that would work. Write is only invoked if more data comes in; it could sit open indefinitely if no more data is received even though the context is cancelled. A separate goroutine waiting on cancellation makes more sense.

ekinanp: Ah yeah, you're right. It wouldn't.

api/streamableResponseWriter.go
```diff
@@ -20,3 +23,12 @@ func (wf *streamableResponseWriter) Write(b []byte) (n int, err error) {
 	wf.Flush()
 	return n, err
 }
+
+type closeFunc = func() error
+
+func streamCleanup(ctx context.Context, desc string, closer closeFunc) {
```
ekinanp: We should also wrap streamableResponseWriter's  `Write` method to ensure that nothing's written on a cancelled context. This would eliminate https://github.com/puppetlabs/wash/blob/master/api/exec.go#L18-L28. Then, all of this context-cancellation stuff can be baked into `newStreamableResponseWriter(ctx, w, endpoint, cleanup)` (I like cleanup better b/c it's more generic). The goroutine can go inside `new`, like in `NewExecCommand`.

MikaelSmith: I think combining them makes it more awkward. And guarding Write isn't really needed because we'll get back an error when we try to write anyway (the write will have already been closed).

ekinanp: I added the `sendPacket` helper in the Exec endpoint because I noticed the server logs were cluttered with context-cancellation related errors. Having the guarded write is nice b/c it guarantees that we avoid redundant error logging for all of our endpoints. I also think it is reasonable for a streamableResponseWriter's Write method to noop on a cancelled context.

I can understand not combining the cleanup func though. It is weird to think of a streamableResponseWriter as handling non-response related cleanup.  Would still be nice if `closer` => `cleanup` to reflect `streamCleanup`'s genericness.

api/streamableResponseWriter.go
```diff
@@ -13,15 +13,23 @@ type flushableWriter interface {
...
 }
 
 // Write flushes the data immediately after every write operation
 func (wf *streamableResponseWriter) Write(b []byte) (n int, err error) {
-	n, err = wf.flushableWriter.Write(b)
-	wf.Flush()
-	return n, err
+	select {
+	case <-wf.ctx.Done():
+		// Don't send anything if the context's finished. Pretend nothing happened.
```
MikaelSmith: @ekinanp I'm not really sure this is always desirable, but it reproduces the behavior you had for sendPacket.

api/streamableResponseWriter.go
```diff
@@ -1,22 +1,42 @@
...
+		return len(b), nil
+	default:
+		n, err = wf.writer.Write(b)
+		wf.writer.Flush()
+		return n, err
+	}
+}
+
+type closeFunc = func() error
```
ekinanp: Nit: Should be `cleanupFunc`

api/streamableResponseWriter.go
ekinanp: Hmm, I guess not including it forces callers to think about proper cleanup, b/c if your streaming endpoint keeps writing it likely means that you're not closing a reader when you're supposed to. That's what led to `sendPacket` in Exec.

I guess it is better to remove it, even if it conceptually makes sense. Sorry for the extra time-waste.

api/streamableResponseWriter.go
MikaelSmith: Done.

PR #262, opened 2019-05-17
Clarify meta; cleanup some plugins; add `wash info`

ekinanp:


ekinanp:
I think we should take some time to double-check the plugin implementations and make sure they're consistent, clean, and that we're using `meta` where we're supposed to. As an example of the latter, would `meta` make sense for K8 contexts? I see some sort of object being returned in https://github.com/puppetlabs/wash/blob/master/plugin/kubernetes/root.go#L25-L40

cmd/info.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/puppetlabs/wash/api/client"
+	cmdutil "github.com/puppetlabs/wash/cmd/util"
+	"github.com/puppetlabs/wash/config"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func infoCommand() *cobra.Command {
+	infoCmd := &cobra.Command{
+		Use:   "info <path>",
```
ekinanp: This is mostly useful for checking what an entry's meta attribute looks like, which is useful when you're trying to figure out a meta primary query. Also open to other ways of exposing the `meta` attribute.

docs/api/api.json
```diff
@@ -399,8 +399,8 @@
       "title": "EntryAttributes represents an entry's attributes. We use a struct\ninstead of a map for efficient memory allocation/deallocation,\nwhich is needed to make Group#List fast.",
       "x-go-package": "github.com/puppetlabs/wash/plugin"
```
MikaelSmith: Looks like  forgot to remove these (`docs/api`) when I rearranged the website.

PR #263, opened 2019-05-18
Share CLI flags for shell and server setup

MikaelSmith:
Since the `wash` shell sets up a server, it should handle all the same
flags tath the server accepts. Create shared utilities for doing so.


MikaelSmith:
This removes `viper` for these options, which we may need to add back for config. But I think we'll have to look carefully at how to structure the config file for this anyway, so not sure what we had before would work either.

cmd/internal/server/levels.go
```diff
@@ -14,7 +14,7 @@ var levelMap = map[string]log.Level{
 	"trace": log.TraceLevel,
 }
 
-func ParseLevel(s string) (log.Level, error) {
+func parseLevel(s string) (log.Level, error) {
```
ekinanp: Maybe bake this in with the Server object's code? It's small enough.

cmd/root.go
```diff
@@ -39,21 +37,16 @@ func rootCommand() *cobra.Command {
...
 		SilenceUsage:  true,
 		SilenceErrors: true,
 	}
-
-	rootCmd.Flags().String("loglevel", "info", "Set the logging level")
-	errz.Fatal(viper.BindPFlag("loglevel", rootCmd.Flags().Lookup("loglevel")))
-
-	rootCmd.Flags().String("logfile", "", "Set the log file's location. Defaults to stdout")
-	errz.Fatal(viper.BindPFlag("logfile", rootCmd.Flags().Lookup("logfile")))
+	opts := serverFlags(rootCmd)
```
ekinanp: Have you looked into something like
```
for _, flag := range serverCmd().Flags() {
    rootCmd.Flags().AddFlag(flag)
}
```
? That should give you what you want.

cmd/serverFlags.go
```diff
@@ -0,0 +1,17 @@
+package cmd
+
+import (
+	"github.com/puppetlabs/wash/cmd/internal/server"
+	"github.com/spf13/cobra"
+)
+
+// SetFlags adds server flags to the specified command.
+func serverFlags(cmd *cobra.Command) *server.Opts {
```
ekinanp: I think this should be in `cmd/server.go`, and renamed to `addServerFlags`. We should also add a `newServer(cmd)`-type method in `server.go` that creates the server object from these flags. This way, we can still bind to Viper, which we will need to do because that is the correct way to parse them from the config file (see https://github.com/puppetlabs/wash/blob/master/config/core.go#L31-L38. Note that with the current version of Wash, these options can be set from the environment). 

I like the symmetry here in binding the option variables to the flag; however, the `makeRootMain`/`makeServerMain` code is weird (but also the only way I can think of to get these set in one place w/ the options, so I get why you added it).

MikaelSmith: I'll experiment more. I'm not sure why the Viper bindings were such an issue when I tried to centralize them.

MikaelSmith: Running into https://github.com/spf13/viper/issues/233.

cmd/internal/server/levels.go
MikaelSmith: Done. I wish there were a more concise way to construct the error, but not finding good options.

cmd/serverFlags.go
MikaelSmith: Fixed.

PR #265, opened 2019-05-20
Update the external plugin interface with the recent changes

ekinanp:

### DIFF
--- a/plugin/externalPluginEntry.go
+++ b/plugin/externalPluginEntry.go
@@ -233,10 +233,10 @@ func (e *externalPluginEntry) Exec(ctx context.Context, cmd string, args []strin
 	// Start the command.
 	cmdObj := e.script.NewInvocation(ctx, "exec", e, append([]string{cmd}, args...)...)
 	execCmd := NewExecCommand(ctx)
-	cmdObj.Stdout = execCmd.Stdout()
-	cmdObj.Stderr = execCmd.Stderr()
+	cmdObj.SetStdout(execCmd.Stdout())
+	cmdObj.SetStderr(execCmd.Stderr())
 	if opts.Stdin != nil {
-		cmdObj.Stdin = opts.Stdin
+		cmdObj.SetStdin(opts.Stdin)
 	}
 	activity.Record(ctx, "Starting %v", cmdObj)
 	if err := cmdObj.Start(); err != nil {
@@ -249,7 +249,7 @@ func (e *externalPluginEntry) Exec(ctx context.Context, cmd string, args []strin
 	go func() {
 		err := cmdObj.Wait()
 		execCmd.CloseStreamsWithError(nil)
-		exitCode := cmdObj.ProcessState.ExitCode()
+		exitCode := cmdObj.ProcessState().ExitCode()
 		if exitCode < 0 {
 			execCmd.SetExitCodeErr(err)
 		} else {
--- a/plugin/externalPluginEntry.go
+++ b/plugin/externalPluginEntry.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/puppetlabs/wash/activity"
 	"github.com/puppetlabs/wash/plugin/internal"
 )
@@ -93,14 +94,13 @@ func (e *externalPluginEntry) List(ctx context.Context) ([]Entry, error) {
 	}
 	var decodedEntries []decodedExternalPluginEntry
 	if err := json.Unmarshal(stdout, &decodedEntries); err != nil {
-		activity.Record(
+		return nil, newStdoutDecodeErr(
 			ctx,
-			"could not decode the entries from stdout\nreceived:\n%v\nexpected something like:\n%v",
-			strings.TrimSpace(string(stdout)),
-			"[{\"name\":\"<name_of_first_entry>\",\"wmethods\":[\"list\"]},{\"name\":\"<name_of_second_entry>\",\"wmethods\":[\"list\"]}]",
+			"the entries",
+			err,
+			stdout,
+			"[{\"name\":\"entry1\",\"methods\":[\"list\"]},{\"name\":\"entry2\",\"methods\":[\"list\"]}]",
 		)
-
-		return nil, fmt.Errorf("could not decode the entries from stdout: %v", err)
 	}
 	entries := make([]Entry, len(decodedEntries))
 	for i, decodedExternalPluginEntry := range decodedEntries {
@@ -134,13 +134,13 @@ func (e *externalPluginEntry) Metadata(ctx context.Context) (JSONObject, error) 
 	}
 	var metadata JSONObject
 	if err := json.Unmarshal(stdout, &metadata); err != nil {
-		activity.Record(
+		return nil, newStdoutDecodeErr(
 			ctx,
-			"could not decode the metadata from stdout\nreceived:\n%v\nexpected something like:\n%v",
-			strings.TrimSpace(string(stdout)),
+			"the metadata",
+			err,
+			stdout,
 			"{\"key1\":\"value1\",\"key2\":\"value2\"}",
 		)
-		return nil, fmt.Errorf("could not decode the metadata from stdout: %v", err)
 	}
 	return metadata, nil
 }
@@ -274,4 +274,19 @@ func (s *stdoutStreamer) Read(p []byte) (int, error) {
 
 func (s *stdoutStreamer) Close() error {
 	return s.cmd.Wait()
+}
+
+func newStdoutDecodeErr(ctx context.Context, decodedThing string, reason error, stdout []byte, example string) error {
+	logMsg := fmt.Sprintf(
+		"could not decode %v from stdout\nreceived:\n%v\nexpected something like:\n%v",
+		decodedThing,
+		strings.TrimSpace(string(stdout)),
+		example,
+	)
+	if ctx == nil {
+		log.Warn(logMsg)
+	} else {
+		activity.Record(ctx, logMsg)
+	}
+	return fmt.Errorf("could not decode %v from stdout: %v", decodedThing, reason)
 }
--- a/plugin/externalPluginEntry.go
+++ b/plugin/externalPluginEntry.go
@@ -7,12 +7,11 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"os/exec"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/puppetlabs/wash/activity"
+	"github.com/puppetlabs/wash/plugin/internal"
 )
 
 type decodedCacheTTLs struct {
@@ -24,7 +23,7 @@ type decodedCacheTTLs struct {
 // decodedExternalPluginEntry describes a decoded serialized entry.
 type decodedExternalPluginEntry struct {
 	Name                 string           `json:"name"`
-	SupportedActions     []string         `json:"supported_actions"`
+	Methods             []string          `json:"methods"`
 	SlashReplacementChar string           `json:"slash_replacement_char"`
 	CacheTTLs            decodedCacheTTLs `json:"cache_ttls"`
 	Attributes           EntryAttributes  `json:"attributes"`
@@ -35,13 +34,12 @@ func (e decodedExternalPluginEntry) toExternalPluginEntry() (*externalPluginEntr
 	if e.Name == "" {
 		return nil, fmt.Errorf("the entry name must be provided")
 	}
-	if e.SupportedActions == nil {
-		return nil, fmt.Errorf("the entry's supported actions must be provided")
+	if e.Methods == nil {
+		return nil, fmt.Errorf("the entry's methods must be provided")
 	}
-
 	entry := &externalPluginEntry{
 		EntryBase:        NewEntry(e.Name),
-		supportedActions: e.SupportedActions,
+		methods:          e.Methods,
 		state:            e.State,
 	}
 	entry.SetAttributes(e.Attributes)
@@ -54,17 +52,14 @@ func (e decodedExternalPluginEntry) toExternalPluginEntry() (*externalPluginEntr
 
 		entry.SetSlashReplacementChar([]rune(e.SlashReplacementChar)[0])
 	}
-
 	return entry, nil
 }
 
-// externalPluginEntry represents an entry from an external plugin. It consists
-// of its name, its object (as serialized JSON), and the path to its
-// main plugin script.
+// externalPluginEntry represents an external plugin entry
 type externalPluginEntry struct {
 	EntryBase
 	script           externalPluginScript
-	supportedActions []string
+	methods         []string
 	state            string
 }
 
@@ -80,25 +75,33 @@ func (e *externalPluginEntry) setCacheTTLs(ttls decodedCacheTTLs) {
 	}
 }
 
-// List lists the entry's children, if it has any.
+// implements returns true if the entry implements the given method,
+// false otherwise
+func (e *externalPluginEntry) implements(method string) bool {
+	for _, m := range e.methods {
+		if method == m {
+			return true
+		}
+	}
+	return false
+}
+
 func (e *externalPluginEntry) List(ctx context.Context) ([]Entry, error) {
-	stdout, err := e.script.InvokeAndWait(ctx, "list", e.id(), e.state)
+	stdout, err := e.script.InvokeAndWait(ctx, "list", e)
 	if err != nil {
 		return nil, err
 	}
-
 	var decodedEntries []decodedExternalPluginEntry
 	if err := json.Unmarshal(stdout, &decodedEntries); err != nil {
 		activity.Record(
 			ctx,
 			"could not decode the entries from stdout\nreceived:\n%v\nexpected something like:\n%v",
 			strings.TrimSpace(string(stdout)),
-			"[{\"name\":\"<name_of_first_entry>\",\"supported_actions\":[\"list\"]},{\"name\":\"<name_of_second_entry>\",\"supported_actions\":[\"list\"]}]",
+			"[{\"name\":\"<name_of_first_entry>\",\"wmethods\":[\"list\"]},{\"name\":\"<name_of_second_entry>\",\"wmethods\":[\"list\"]}]",
 		)
 
 		return nil, fmt.Errorf("could not decode the entries from stdout: %v", err)
 	}
-
 	entries := make([]Entry, len(decodedEntries))
 	for i, decodedExternalPluginEntry := range decodedEntries {
 		entry, err := decodedExternalPluginEntry.toExternalPluginEntry()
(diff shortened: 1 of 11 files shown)


ekinanp:
I'm going to doc updates in a separate PR, because I'd like to write a general plugin tutorial doc that will include some of the stuff discussed in the present external plugin docs. That lets us make the latter more external-plugin focused instead of general plugin philosophy + external-plugin focused.

website/static/docs/external_plugins/examples/sshfs.sh
```diff
@@ -145,7 +157,10 @@ function print_children {
   # The code here is equivalent to ls'ing the directory, then invoking
   # `test-d` and stat on each entry to obtain the following information:
   #   <is_dir> <sizeAttr> <atime> <mtime> <ctime> <mode> <path>
-  stat_output=`vm_exec ${vm} "find ${dir} -mindepth 1 -maxdepth 1 | xargs -n1 -r -I {} bash -c '(test -d \\$@; echo -n \"\\$? \") && stat -c \"%s %X %Y %Z %f %n\" \\$@' _ {}"`
+  #
+  # TODO: Handle TTY strangeness in the returned output. For now, it is
```
ekinanp: This was causing missing chunks of output from the printed JSON. I.e. things looked something like:
```
[
","methods":["list"],"attributes":{"atime":1558322711,"mtime":1558133886,"ctime":1558133886,"mode":16749,"size":4096}},
","methods":["list"],"attributes":{"atime":1558133854,"mtime":1558133854,"ctime":1558133854,"mode":16749,"size":0}},
","methods":["list"],"attributes":{"atime":1558322715,"mtime":1558133858,"ctime":1558133858,"mode":16877,"size":3120}},
","methods":["list"],"attributes":{"atime":1558246965,"mtime":1558133869,"ctime":1558133869,"mode":16877,"size":840}},
","methods":["list"],"attributes":{"atime":1558245893,"mtime":1558133854,"ctime":1558133854,"mode":16749,"size":0}},
","methods":["list"],"attributes":{"atime":1558330785,"mtime":1558136154,"ctime":1558136154,"mode":16744,"size":4096}},
","methods":["list"],"attributes":{"atime":1558323061,"mtime":1558133868,"ctime":1558133868,"mode":16877,"size":8192}},
","methods":["list"],"attributes":{"atime":1558133859,"mtime":1558323061,"ctime":1558323061,"mode":17407,"size":4096}},
","methods":["list"],"attributes":{"atime":1558311781,"mtime":1558133856,"ctime":1558133858,"mode":16877,"size":4096}},
","methods":["list"],"attributes":{"atime":1558330786,"mtime":1547678697,"ctime":1547678697,"mode":16877,"size":4096}},
","methods":["list"],"attributes":{"atime":1558306860,"mtime":1547678697,"ctime":1547678697,"mode":57855,"size":7}},
","methods":["list"],"attributes":{"atime":1558306861,"mtime":1547678697,"ctime":1547678697,"mode":57855,"size":8}},
","methods":["list"],"attributes":{"atime":1558330784,"mtime":1547678697,"ctime":1547678697,"mode":57855,"size":7}},
","methods":["list"],"attributes":{"atime":1558306668,"mtime":1547678697,"ctime":1547678697,"mode":57855,"size":9}},
","methods":["list"],"attributes":{"atime":1432543056,"mtime":1432543056,"ctime":1547678697,"mode":16877,"size":6}},
","methods":["list"],"attributes":{"atime":1558246933,"mtime":1432543056,"ctime":1547678697,"mode":16877,"size":6}},
","methods":["list"],"attributes":{"atime":1558322709,"mtime":1547679200,"ctime":1547679200,"mode":16877,"size":17}},
","methods":["list"],"attributes":{"atime":1558322710,"mtime":1547679439,"ctime":1547679439,"mode":16877,"size":6}},
","methods":["list"],"attributes":{"atime":1558322710,"mtime":1432543056,"ctime":1547678697,"mode":16877,"size":6}},
","methods":["read","stream"],"attributes":{"atime":1547678985,"mtime":1547678985,"ctime":1547678985,"mode":33188,"size":0}}
]
```

when they should instead look something like
```
[
{"name":"dev","methods":["list"],"attributes":{"atime":1558322715,"mtime":1558133858,"ctime":1558133858,"mode":16877,"size":3120}},
{"name":"boot","methods":["list"],"attributes":{"atime":1558322711,"mtime":1558133886,"ctime":1558133886,"mode":16749,"size":4096}},
{"name":"proc","methods":["list"],"attributes":{"atime":1558133854,"mtime":1558133854,"ctime":1558133854,"mode":16749,"size":0}},
{"name":"run","methods":["list"],"attributes":{"atime":1558246965,"mtime":1558133869,"ctime":1558133869,"mode":16877,"size":840}},
{"name":"sys","methods":["list"],"attributes":{"atime":1558245893,"mtime":1558133854,"ctime":1558133854,"mode":16749,"size":0}},
{"name":"root","methods":["list"],"attributes":{"atime":1558330785,"mtime":1558136154,"ctime":1558136154,"mode":16744,"size":4096}},
{"name":"etc","methods":["list"],"attributes":{"atime":1558323061,"mtime":1558133868,"ctime":1558133868,"mode":16877,"size":8192}},
{"name":"tmp","methods":["list"],"attributes":{"atime":1558133859,"mtime":1558323061,"ctime":1558323061,"mode":17407,"size":4096}},
{"name":"var","methods":["list"],"attributes":{"atime":1558311781,"mtime":1558133856,"ctime":1558133858,"mode":16877,"size":4096}},
{"name":"usr","methods":["list"],"attributes":{"atime":1558330786,"mtime":1547678697,"ctime":1547678697,"mode":16877,"size":4096}},
{"name":"bin","methods":["list"],"attributes":{"atime":1558306860,"mtime":1547678697,"ctime":1547678697,"mode":57855,"size":7}},
{"name":"sbin","methods":["list"],"attributes":{"atime":1558306861,"mtime":1547678697,"ctime":1547678697,"mode":57855,"size":8}},
{"name":"lib","methods":["list"],"attributes":{"atime":1558330784,"mtime":1547678697,"ctime":1547678697,"mode":57855,"size":7}},
{"name":"lib64","methods":["list"],"attributes":{"atime":1558306668,"mtime":1547678697,"ctime":1547678697,"mode":57855,"size":9}},
{"name":"home","methods":["list"],"attributes":{"atime":1432543056,"mtime":1432543056,"ctime":1547678697,"mode":16877,"size":6}},
{"name":"mnt","methods":["list"],"attributes":{"atime":1558322709,"mtime":1547679200,"ctime":1547679200,"mode":16877,"size":17}},
{"name":"media","methods":["list"],"attributes":{"atime":1558246933,"mtime":1432543056,"ctime":1547678697,"mode":16877,"size":6}},
{"name":"opt","methods":["list"],"attributes":{"atime":1558322710,"mtime":1547679439,"ctime":1547679439,"mode":16877,"size":6}},
{"name":"srv","methods":["list"],"attributes":{"atime":1558322710,"mtime":1432543056,"ctime":1547678697,"mode":16877,"size":6}},
{"name":".autorelabel","methods":["read","stream"],"attributes":{"atime":1547678985,"mtime":1547678985,"ctime":1547678985,"mode":33188,"size":0}}
]
```

plugin/externalPluginEntry.go
```diff
@@ -217,121 +185,71 @@ func (e *externalPluginEntry) Stream(ctx context.Context) (io.ReadCloser, error)
...
 	}
 }
 
-// Exec executes a command on the given entry
 func (e *externalPluginEntry) Exec(ctx context.Context, cmd string, args []string, opts ExecOptions) (ExecCommand, error) {
-	// TODO: Figure out how to pass-in opts when we have entries
-	// besides Stdin. Could do something like
+	// Start the command.
+	//
+	// TODO: Handle other options. Go with
```
ekinanp: Does this seem like a good calling convention for Exec's options? Note that `<opts>` will be JSON serialized.

We could also do `<cmd> <args> <opts>` where `<args>` is a JSON array, but that could make things harder for Bash plugin authors (and also less flexible with line-by-line output formats due to the possibility of an arg including a newline).

plugin/externalPluginEntry_test.go
```diff
@@ -21,16 +23,37 @@ func (m *mockExternalPluginScript) Path() string {
 	return m.path
 }
 
-func (m *mockExternalPluginScript) InvokeAndWait(ctx context.Context, args ...string) ([]byte, error) {
-	retValues := m.Called(ctx, args)
+func (m *mockExternalPluginScript) InvokeAndWait(
```
ekinanp: This refactor mirrors the calling convention in the external plugin doc, which is why I wanted to do it.

plugin/externalPluginScript.go
```diff
@@ -4,16 +4,16 @@ import (
...
 
 	"github.com/puppetlabs/wash/activity"
+	"github.com/puppetlabs/wash/plugin/internal"
 )
 
 // externalPluginScript represents an external plugin's script
 type externalPluginScript interface {
 	Path() string
-	InvokeAndWait(ctx context.Context, args ...string) ([]byte, error)
+	InvokeAndWait(ctx context.Context, method string, entry *externalPluginEntry, args ...string) ([]byte, error)
```
ekinanp: This specification mirrors the calling convention in the external plugin doc, which is why I wanted to do it.

plugin/internal/command.go
```diff
@@ -0,0 +1,148 @@
+package internal
```
ekinanp: I don't want to export `internal.Command` because it was created to solve a specific use-case. If this makes sense, then I think we should move the core plugins over to a `plugin/core` folder so a newer Go user doesn't think this is a core plugin named `internal` (and it also makes sense to place them in their own directory).

plugin/internal/command.go
```diff
@@ -0,0 +1,148 @@
...
+	if ctx == nil {
+		panic("plugin.newCommand called with a nil context")
+	}
+	cmdObj := &Command{
+		Cmd:           exec.Command(cmd, args...),
+		ctx:           ctx,
+		pgid:          -1,
+		waitDoneCh:    make(chan struct{}),
+	}
+	cmdObj.SysProcAttr = &syscall.SysProcAttr{
```
ekinanp: The PGID code was ripped off https://medium.com/@felixge/killing-a-child-process-and-all-of-its-children-in-go-54079af94773

plugin/externalPluginRoot.go
```diff
@@ -24,34 +25,36 @@ func newExternalPluginRoot(script string) *externalPluginRoot {
 
 // Init initializes the external plugin root
 func (r *externalPluginRoot) Init() error {
-	ctx := context.Background()
-	script := r.script
-
-	stdout, err := r.script.InvokeAndWait(ctx, "init")
+	// Give external plugins about five-seconds to finish their
+	// initialization
+	ctx, cancelFunc := context.WithTimeout(context.Background(), 5 * time.Second)
```
ekinanp: Does this make sense? Should we go with a lower timeout?

plugin/externalPluginRoot.go
```diff
@@ -60,6 +63,8 @@ func (r *externalPluginRoot) Init() error {
 // We need this b/c *externalPluginEntry#List has a different receiver
 // (*externalPluginEntry) than *externalPluginRoot (i.e. b/c Go's
 // typechecker complains about it)
+//
+// TODO: Is this still an issue?
```
MikaelSmith: Dunno? How would you test it?

plugin/externalPluginRoot_test.go
```diff
@@ -36,14 +40,14 @@ func (suite *ExternalPluginRootTestSuite) TestInit() {
 	suite.Regexp(regexp.MustCompile("stdout"), err)
 
 	// Test that Init properly decodes the root from stdout
-	stdout := "{\"name\":\"foo\",\"supported_actions\":[\"list\"]}"
+	stdout := "{\"name\":\"foo\",\"wmethods\":[\"list\"]}"
```
MikaelSmith: How does this test pass? I don't see anything parsing `wmethods`.

plugin/externalPluginEntry.go
```diff
@@ -217,121 +185,71 @@ func (e *externalPluginEntry) Stream(ctx context.Context) (io.ReadCloser, error)
...
-			args...,
-		)...,
-	)
-
-	// Set-up stdin
-	if opts.Stdin != nil {
-		cmdObj.Stdin = opts.Stdin
-	}
-
+	// (opts is the JSON-serialization of Exec's options)
```
MikaelSmith: I don't see opts being used anywhere.

plugin/externalPluginRoot.go
MikaelSmith: No, this seems fine. Low timeouts can be finicky.

plugin/internal/command.go
MikaelSmith: I think the module description can handle that pretty well. Which we should add here.

plugin/internal/command.go
```diff
@@ -0,0 +1,148 @@
...
+	"syscall"
+	"time"
+
+	"github.com/puppetlabs/wash/activity"
+)
+
+// Command is a wrapper to exec.Cmd. It handles context-cancellation cleanup
+// and defines a String() method to make logging easier.
+type Command struct {
+	*exec.Cmd
```
MikaelSmith: It may be more confusing than helpful to make this wrap `exec.Cmd` rather than having it as a named property. You override some of the core methods, so not sure others would behave correctly.

website/static/docs/external_plugins/examples/sshfs.sh
```diff
@@ -48,51 +48,63 @@ function vm_exec {
...
+    # command is running. Setting up a tty ensures that the remote
+    # SSH command is killed when the calling process (our
+    # process) is killed. This avoids orphaned SSH processes.
+    #
+    # NOTE: We make TTY optional because it is one of the passed-in
+    # Exec options.
+    #
+    # NOTE: This type of plugin-specific cleanup is the plugin
+    # author's responsibility, not Wash's.
+    ssh -t -t root@"${vm}" "${cmd}"
```
MikaelSmith: You can join these as `ssh -tt`.

plugin/externalPluginEntry.go
ekinanp: Yeah, that is more idiomatic to how you'd think about an invoked command. It does require some opt-parsing logic though, but we could make that calling convention configurable.

plugin/internal/command.go
ekinanp: I can add a module description.

What's the reason for not moving the core plugins over to `plugin/core` ?

plugin/externalPluginEntry.go
```diff
@@ -80,56 +75,63 @@ func (e *externalPluginEntry) setCacheTTLs(ttls decodedCacheTTLs) {
...
 	}
-
 	var decodedEntries []decodedExternalPluginEntry
 	if err := json.Unmarshal(stdout, &decodedEntries); err != nil {
 		activity.Record(
 			ctx,
 			"could not decode the entries from stdout\nreceived:\n%v\nexpected something like:\n%v",
 			strings.TrimSpace(string(stdout)),
-			"[{\"name\":\"<name_of_first_entry>\",\"supported_actions\":[\"list\"]},{\"name\":\"<name_of_second_entry>\",\"supported_actions\":[\"list\"]}]",
+			"[{\"name\":\"<name_of_first_entry>\",\"wmethods\":[\"list\"]},{\"name\":\"<name_of_second_entry>\",\"wmethods\":[\"list\"]}]",
```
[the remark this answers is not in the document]
ekinanp: Ah yeah. I might as well add tests for these cases. Didn't at the time b/c I didn't want to deal with a long regex match.

plugin/externalPluginEntry.go
```diff
@@ -138,76 +140,42 @@ func (e *externalPluginEntry) Metadata(ctx context.Context) (JSONObject, error)
...
+		return nil, err
+	}
+	activity.Record(ctx, "Starting %v", cmd)
 	if err := cmd.Start(); err != nil {
-		activity.Record(ctx, "Closed command stdout: %v", stdoutR.Close())
-		activity.Record(ctx, "Closed command stderr: %v", stderrR.Close())
 		return nil, err
 	}
+	// "wait" will be used in Stream's error handlers. It will be wrapped
+	// in a "defer" call to ensure that cmd.Wait()'s called once we've read
```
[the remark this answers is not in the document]
ekinanp: My reason for doing it here is to enforce the idea that "If Stream outputs something to stderr, something went wrong w/ it so the process should have terminated. We can check the latter by calling Wait(). If something went wrong with the termination, or if the process was killed by the context-cancellation cleanup, we can log Wait's error." For stdout, I moved the `Wait()` over to `Close` because that's when we want to wait for termination and release the resources. So I think this code here is correct.

plugin/externalPluginRoot.go
ekinanp: Ah, I'll address this TODO here.

plugin/externalPluginRoot_test.go
ekinanp: Huh, that is weird. Not sure. I'll double check this.

plugin/externalPluginEntry.go
ekinanp: Was from an earlier iteration. I'll remove it.

plugin/internal/command.go
MikaelSmith: I don't think it'll actually improve clarity. Will https://godoc.org/github.com/puppetlabs/wash/plugin look any different? The table at https://godoc.org/github.com/puppetlabs/wash seems pretty straightforward to me, and we use internal other places so that seems like a clear pattern.

plugin/externalPluginEntry.go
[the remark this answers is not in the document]
MikaelSmith: Ok.

plugin/internal/command.go
ekinanp: According to https://travix.io/type-embedding-in-go-ba40dd4264df, Go's type embedding makes sure that embedded structs don't have access to anything in the embedding struct. Thus, any of `exec.Cmd`'s core methods should call only `exec.Cmd`'s core methods, meaning this embedding should be OK and should not affect anything in `exec.Cmd`.

plugin/internal/command.go
ekinanp: Hmm, think you're probably right. `internal` is a documented Go pattern for modular code you don't want to export.

Ok, I'll just add a module description.

plugin/internal/command.go
MikaelSmith: My point was if someone ran `Output` on our Command, it wouldn't use our Start/Wait methods and so probably wouldn't work as expected. So is it worth letting `exec.Cmd`'s `Output` method be exposed?

ekinanp: Ohh I see. Sorry about misunderstanding what you meant.

Their implementations are pretty short, so may as well expose them in case they're ever needed.

ekinanp: Hmm, maybe not. `Output` is a bit complicated because they use an internal io.Writer. We could copy it word for word, but I don't see a good reason to do that yet.

Guess I'll go with embedding a property, even if it means re-defining some exported fields as setters. Makes things more secure.

plugin/externalPluginEntry.go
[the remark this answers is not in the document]
ekinanp: Nevermind, there isn't a good way to test this since it is documentation (meaning the assertion is the example against itself), and the activity package isn't mocked. I'll just have to be more diligent in updating them.

I could make these internal variables that consist of serialized decoded external plugin entries (so that a change like `wmethods` would throw a compiler error), but that's not worth the complexity right now. May look into it later.

plugin/externalPluginRoot_test.go
ekinanp: Oh, forgot to address this. One second.

plugin/externalPluginEntry.go
```diff
@@ -274,4 +274,19 @@ func (s *stdoutStreamer) Read(p []byte) (int, error) {
...
+}
+
+func newStdoutDecodeErr(ctx context.Context, decodedThing string, reason error, stdout []byte, example string) error {
+	logMsg := fmt.Sprintf(
+		"could not decode %v from stdout\nreceived:\n%v\nexpected something like:\n%v",
+		decodedThing,
+		strings.TrimSpace(string(stdout)),
+		example,
+	)
+	if ctx == nil {
```
MikaelSmith: When will context be nil? And why are you handling this rather than letting `activity` handle it?

MikaelSmith: Ah, that's clear in the last commit.

ekinanp: That's removed in the last commit -- was a mistake

PR #269, opened 2019-05-20
Reorganize README and website, improve command help

### LINKED ISSUE #266: Document commands
Clear, Exec, Find, History, List, Meta, PS, Tail.

MikaelSmith:
Resolves #266.

### DIFF
--- a/cmd/clear.go
+++ b/cmd/clear.go
@@ -14,7 +14,10 @@ func clearCommand() *cobra.Command {
 	clearCmd := &cobra.Command{
 		Use:   "clear [<path>]",
 		Short: "Clears the cache at <path>, or the current directory if not specified",
-		Args:  cobra.MaximumNArgs(1),
+		Long: `Wash caches most operations. If the resource you're querying appears out-of-date, use this
+subcommand to reset the cache for resources at or contained within <path>. Defaults to the current
+directory if <path> is not specified.`,
+		Args: cobra.MaximumNArgs(1),
 	}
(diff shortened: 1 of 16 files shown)


cmd/clear.go
```diff
@@ -14,7 +14,10 @@ func clearCommand() *cobra.Command {
 	clearCmd := &cobra.Command{
 		Use:   "clear [<path>]",
```
ekinanp: Should we include the aliases created in https://github.com/puppetlabs/wash/blob/master/cmd/rootMain.go#L51-L59 in each of the command objects? That way, the help text prints those out too. It would also be nice if the command objects were the source of truth for their aliases on top of everything else.

website/content/docs/_index.md
```diff
@@ -2,8 +2,129 @@
...
+
+### wash clear
+
+Wash caches most operations. If the resource you're querying appears out-of-date, use this subcommand to reset the cache for resources at or contained within the specified path. Defaults to the current directory if a path is not specified.
+
+### wash exec
+
+For a Wash resource that implements the ability to execute a command, run the specified command and arguments. The results will be forwarded from the target on stdout, stderr, and exit code.
+
+### wash find
```
ekinanp: I'd hate updating `wash find` and having to remember to update this too. Is there a way we can automate some of this cruft via go generate or via something in Hugo? It'd be nice if whatever we do sources this stuff from the command objects.

PR #270, opened 2019-05-21
 Support ExecOptions in external plugin exec

ekinanp:

### DIFF
--- a/plugin/externalPluginEntry.go
+++ b/plugin/externalPluginEntry.go
@@ -229,8 +229,13 @@ func (e *externalPluginEntry) Stream(ctx context.Context) (io.ReadCloser, error)
 }
 
 func (e *externalPluginEntry) Exec(ctx context.Context, cmd string, args []string, opts ExecOptions) (ExecCommand, error) {
+	// Serialize opts to JSON
+	optsJSON, err := json.Marshal(opts)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal opts %v into JSON: %v", opts, err)
+	}
 	// Start the command.
-	cmdObj := e.script.NewInvocation(ctx, "exec", e, append([]string{cmd}, args...)...)
+	cmdObj := e.script.NewInvocation(ctx, "exec", e, append([]string{string(optsJSON), cmd}, args...)...)
 	execCmd := NewExecCommand(ctx)
 	cmdObj.SetStdout(execCmd.Stdout())
 	cmdObj.SetStderr(execCmd.Stderr())
--- a/website/static/docs/external_plugins/examples/sshfs.sh
+++ b/website/static/docs/external_plugins/examples/sshfs.sh
@@ -16,6 +16,11 @@
 # as the root user. For example, it assumes that something like
 # "ssh root@<vm> ls" works.
 #
+# Since Wash passes-in JSON objects to external plugin scripts,
+# this example also requires jq to handle the parsing. We will
+# eventually add multiple input formats, some of which are more
+# "bash" friendly.
+#
 # If you'd like to try this plugin out, then add the following to
 # your plugins.yaml file:
 #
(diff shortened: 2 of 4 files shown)


ekinanp:
Tested this with `wash ps`, which works:
```
enis:wash enis.inan$ ./wash ps mnt/sshfs/fub3cgsl9cxeqnq.delivery.puppetlabs.net/
NODE                                               PID     TIME       COMMAND
mnt/ss*s/fub3cgsl9cxeqnq.delivery.puppetlabs.net   1       00:00.00   /usr/lib/systemd/systemd --switched-root --system --deserialize 22
mnt/ss*s/fub3cgsl9cxeqnq.delivery.puppetlabs.net   2       00:00.00   (kthreadd)
mnt/ss*s/fub3cgsl9cxeqnq.delivery.puppetlabs.net   3       00:00.00   (ksoftirqd/0)
mnt/ss*s/fub3cgsl9cxeqnq.delivery.puppetlabs.net   5       00:00.00   (kworker/0:0H)
mnt/ss*s/fub3cgsl9cxeqnq.delivery.puppetlabs.net   7       00:00.00   (migration/0)
mnt/ss*s/fub3cgsl9cxeqnq.delivery.puppetlabs.net   8       00:00.00   (rcu_bh)
...
```

plugin/externalPluginEntry.go
```diff
@@ -229,8 +229,13 @@ func (e *externalPluginEntry) Stream(ctx context.Context) (io.ReadCloser, error)
...
 
 func (e *externalPluginEntry) Exec(ctx context.Context, cmd string, args []string, opts ExecOptions) (ExecCommand, error) {
+	// Serialize opts to JSON
+	optsJSON, err := json.Marshal(opts)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal opts %v into JSON: %v", opts, err)
+	}
 	// Start the command.
-	cmdObj := e.script.NewInvocation(ctx, "exec", e, append([]string{cmd}, args...)...)
+	cmdObj := e.script.NewInvocation(ctx, "exec", e, append([]string{string(optsJSON), cmd}, args...)...)
```
ekinanp: Calling convention's `<opts> <cmd> <args...>`

website/static/docs/external_plugins/examples/sshfs.sh
```diff
@@ -260,18 +260,27 @@ if [[ "${path}" == "" ]]; then
...
 
-    # exec'ing <cmd> <args> on a VM is equivalent to exec'ing it
-    # on the VM via. ssh (vm_exec)
+    # First we parse the provided Exec options. Wash guarantees that
+    # the passed-in options are valid JSON and that they are all present,
+    # so we don't need to do our own validation.
+    #
+    # NOTE: Only the "tty" option is relevant. We don't need to worry about
+    # "elevate" because we are already running our commands as root.
+    tty=`echo "${opts}" | jq .tty`
```
MikaelSmith: The dependency on jq is annoying, but it's so useful.

MikaelSmith: https://stackoverflow.com/questions/3919755/how-to-parse-query-string-from-a-bash-cgi-script is a cute alternative. If we only expect simple arguments, we could do that and let people split on `&` in other languages (which is pretty easy).

plugin/externalPluginEntry.go
```diff
@@ -229,8 +229,13 @@ func (e *externalPluginEntry) Stream(ctx context.Context) (io.ReadCloser, error)
 }
 
 func (e *externalPluginEntry) Exec(ctx context.Context, cmd string, args []string, opts ExecOptions) (ExecCommand, error) {
+	// Serialize opts to JSON
+	optsJSON, err := json.Marshal(opts)
```
MikaelSmith: Do we really want to pass `stdin`? Seems like we only want a subset of options.

MikaelSmith: Oh, looks like you dealt with that with some json markup.

website/static/docs/external_plugins/examples/sshfs.sh
ekinanp: Yeah, agreed. I think it's worth revisiting the SSHFS example once we start adding more bash-friendly output formats so that the jq dependency can be removed. I can add a TODO in the example to reflect that.

plugin/externalPluginEntry.go
ekinanp: `json:"-"` in https://github.com/ekinanp/wash/blob/update-ep-exec/plugin/types.go#L69 tells Go not to include that field in the serialization, so it won't be included. I'll add a comment about that though.

docs/docs/external_plugins/examples/sshfs.sh
```diff
@@ -16,13 +16,21 @@
...
+#
 # If you'd like to try this plugin out, then add the following to
 # your plugins.yaml file:
 #
 #   - name: 'sshfs'
 #     script: '<wash_parent>/wash/docs/external_plugins/examples/sshfs.sh'
 #
 # and start-up the Wash server
+#
+# NOTE: This example depends on
```
MikaelSmith: Incomplete sentence.

ekinanp: Thanks! Updated.

ekinanp:
Updated. I meant to update `website`. Going to run Hugo on the docs later once the tutorial is written and the external plugin docs have been updated.

PR #273, opened 2019-05-21
Fix flags, add subcommand aliases

MikaelSmith:


cmd/info.go
```diff
@@ -3,12 +3,10 @@ package cmd
...
 
-	"github.com/Benchkram/errz"
 	"github.com/puppetlabs/wash/api/client"
 	cmdutil "github.com/puppetlabs/wash/cmd/util"
 	"github.com/puppetlabs/wash/config"
 	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
 )
 
 func infoCommand() *cobra.Command {
```
ekinanp: Could we alias this to `winfo` ?

MikaelSmith: Hmm, in my case it's not builtin, it's just at /usr/bin/info. Which shell are you using?

ekinanp: Oh, I meant mine is also at `/usr/bin/info`. I'm using Bash.

If `info` isn't used that often, OK to keep it as-is

MikaelSmith: It seems to be an alternative to `man`. I've never used it.

ekinanp: May as well alias it then. `winfo` is still pretty short

MikaelSmith: The same argument can be made for `clear`. Should I change that back to `wclear` as well, since it doesn't do the same thing?

MikaelSmith: Guess that makes sense. Done.

ekinanp: Yeah, definitely for `clear`. I use clear a lot in my shell.

cmd/rootMain.go
```diff
@@ -53,6 +53,7 @@ func runShell(cachedir, mountpath string) exitCode {
 		"wexec":    "exec",
 		"find":     "find",
 		"help":     "help",
+		"winfo":    "info",
```
ekinanp: It'd be nice if this map were generated by the RootCommand so I don't have to remember to update this whenever I add a new subcommand, but it's not a blocker or that big of a deal.

MikaelSmith: I started doing that, then decided I didn't want to include all of them so left it out.

Issue #132, opened 2019-03-19
Figure out where wash config. should go. This includes the server config. (wash.yaml) + external plugins (plugins.yaml)


MikaelSmith:

jdwelch:
`~/.puppetlabs/wash`?

MikaelSmith:
As in `~/.puppetlabs/wash/wash.yml` and `~/.puppetlabs/wash/plugins.yaml`? I'm not sure plugins needs to be a separate file actually.

jdwelch:
Yup. Single file sounds good to me until there's some reason to make it more complex.

PR #274, opened 2019-05-21
Add Wash config file

ekinanp:
Resolves https://github.com/puppetlabs/wash/issues/132


cmd/internal/server/core.go
```diff
@@ -26,10 +24,10 @@ import (
 // Opts exposes additional configuration for server operation.
 type Opts struct {
 	CPUProfilePath      string
-	ExternalPluginsPath string
+	ExternalPlugins     []plugin.ExternalPluginSpec
 	LogFile             string
 	// LogLevel can be "warn", "info", "debug", or "trace".
-	LogLevel string
+	LogLevel            string
```
ekinanp: I think we should include the `Socket` as an option, or make it a parameter to `New`. That would remove the config dependency and make the `Server` object more generic.

MikaelSmith: That makes sense.

MikaelSmith:
I think we should also roll plugins.yaml into this, as noted in #132.

cmd/server.go
```diff
@@ -62,24 +72,48 @@ func serverMain(cmd *cobra.Command, args []string) exitCode {
...
+	if err != nil {
+		panic(err.Error())
+	}
+	if err := config.ReadFrom(configFile); err != nil {
+		return server.Opts{}, err
+	}
+
+	// Unmarshal the external plugins, if any are specified
+	var externalPlugins []plugin.ExternalPluginSpec
+	if externalPluginsRaw := viper.Get("external_plugins"); externalPluginsRaw != nil {
```
MikaelSmith: Why is this underscore, unlike all the other flags?

ekinanp: Oh, that's not a flag anymore. It's underscore b/c in the YAML, it'd be specified under an `external_plugins` key. Don't mind changing it to `external-plugins` though

MikaelSmith: Can you right your YAML file without quoting it that way?

Issue #275, opened 2019-05-22
Set option for meta primary to check full metadata in `wash find`


ekinanp:
Checking the full metadata is useful when the plugin API returns incomplete information via its List endpoint (like Docker, where `wash meta` outputs much more useful information about a container than what's in the `meta` attribute). However, this should be optional because it can be at least N requests, which can be expensive for e.g. an API like AWS.

Issue #277, opened 2019-05-22
Add daystart option to `wash find`


ekinanp:
In GNU find, this is a positional option that, when set, bases the comparison time for subsequent tests on the start of the current day instead of find's start time. Setting this would enable users to query against "today," which is useful when you want to find all things that were started/created today.

For the meta primary, the query would be `find <path> -daystart -m .time -{1d}` which translates to "within a day from today" => "today".

ekinanp:
Making this an option instead of a primary because expressions like `-m .time +1d -o -daystart -m .time -1d` can be hard to think about (former +1d is based on find's start time, while latter -1d is based on the current day).

PR #278, opened 2019-05-22
Add daystart + fullmeta options to `wash find`

ekinanp:
Resolves https://github.com/puppetlabs/wash/issues/275 and https://github.com/puppetlabs/wash/issues/277

### DIFF
--- a/cmd/internal/find/primary/meta.go
+++ b/cmd/internal/find/primary/meta.go
@@ -366,7 +366,7 @@ being compared. Then:
     of time values include stringified dates or unix seconds.
 
   * If Nu is specified, where Nu means N suffixed with a unit, then the difference "d"
-    between find's start time and v will be compared to N scaled as: 
+    between the reference time and v will be compared to N scaled as: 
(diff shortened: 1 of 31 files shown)


cmd/internal/find/primary/meta.go
```diff
@@ -67,15 +67,29 @@ var Meta = Parser.add(&Primary{
 })
 
 const metaDetailedDescription = `
-The meta primary constructs a predicate on the entry's meta attribute.
-It is a specialized filter, so you should only use it if you need to
+The meta primary constructs a predicate on the entry's metadata. By
+default, this is the meta attribute. If you'd like to construct the
+predicate on the entry's full metadata, then set the "fullmeta" option.
```
MikaelSmith: Is there a way to see the difference between data you'll be using for meta vs fullmeta?

MikaelSmith: I think we should ticket adding that as a new feature of the commands/API.

PR #282, opened 2019-05-23
Fix bug in `wash find` + add attribute flag to `wash meta`

ekinanp:

### DIFF
--- a/cmd/meta.go
+++ b/cmd/meta.go
@@ -17,12 +17,17 @@ func metaCommand() *cobra.Command {
 		RunE:  toRunE(metaMain),
 	}
 	metaCmd.Flags().StringP("output", "o", "json", "Set the output format (json or yaml)")
+	metaCmd.Flags().BoolP("attribute", "a", false, "Show the meta attribute instead of the full metadata")
 	return metaCmd
 }
 
 func metaMain(cmd *cobra.Command, args []string) exitCode {
 	path := args[0]
 	output, err := cmd.Flags().GetString("output")
+	if err != nil {
+		panic(err.Error())
+	}
+	showMetaAttr, err := cmd.Flags().GetBool("attribute")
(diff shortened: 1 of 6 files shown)


cmd/meta.go
```diff
@@ -17,6 +17,7 @@ func metaCommand() *cobra.Command {
 		RunE:  toRunE(metaMain),
 	}
 	metaCmd.Flags().StringP("output", "o", "json", "Set the output format (json or yaml)")
+	metaCmd.Flags().BoolP("attribute", "a", false, "Show the meta attribute instead of the full metadata")
```
MikaelSmith: I think this could use a little more description. I'd add a Long description to `meta` that talks about the two ways we fetch metadata. I think we should also add a small note about it where `find` talks about `-fullmeta`.

ekinanp:
Oops, prematurely pushed. I want to update `meta` primary description to tell people to use `meta` to construct their queries.

cmd/meta.go
```diff
@@ -12,11 +12,14 @@ import (
 func metaCommand() *cobra.Command {
 	metaCmd := &cobra.Command{
 		Use:   "meta <path>",
-		Short: "Prints the metadata of a resource",
+		Short: "Prints the entry's metadata",
+		Long:  `Prints the entry's metadata. By default, meta prints the full metadata as returned by the
```
MikaelSmith: I feel like this description still requires an understanding of plugin design. I'd say something like `Specify the --attribute flag to print the meta attribute - a reduced set of metadata returned when entries are enumerated, also viewable with the 'info' command - instead.`

ekinanp: That's much better. I did remove the reference to `info` because it doesn't clarify anything. May be worth removing the `info` command altogether since I originally introduced it to print the `meta` attribute.

PR #287, opened 2019-05-23
Rename "Group" => "Parent"

ekinanp:
Latter's more general and is consistent with how we've been thinking
about "Group" entries throughout the code.


ekinanp:
I'd like to update the swagger dogs + website, but `go get` is being very slow right now.

I also chose not to add a `plugin.NewGroup` helper because it turns out that makes things more complicated in the context of nested groups (like `resources/ec2`, `resources/s3`) -- it would require a bunch of nested `List` callbacks. So for now, I think it's OK for us to go with the simpler (if verbose) style of the existing structure.

Issue #302, opened 2019-05-29
Run multiple `wash` shells simultaneously


MikaelSmith:
Currently starting a new shell or `wash server` attempts to overwrites the previous socket. If one's running, it appears to fail to create a new one. That causes all open shells to start failing on API operations.

To make this work, we can either
- run multiple servers on temporary sockets. Commands could be configured to connect to a shell-specific one via an environment variable the shell sets.
- run a single server out-of-process, using file locking to ensure a single server, with the last shell to exit handling shutdown.

Running a single shared server is how wash is intended to be used, so let's take the 2nd approach.

ekinanp:
We'll probably want a way to re-use the same mountpoint when starting up a new shell since API paths are passed-in as absolute paths.

MikaelSmith:
I think this also needs to take into account prompting for input. If the daemon is running in a separate process, it needs to capture input and output from plugins so that it can forward that to the shell that initiated the request.

PR #303, opened 2019-05-29
Provide workaround for running multiple Wash shells

MikaelSmith:
Temporary fix for #302.

### DIFF
--- a/cmd/rootMain.go
+++ b/cmd/rootMain.go
@@ -8,7 +8,6 @@ import (
 	"path/filepath"
 
 	"github.com/mattn/go-isatty"
-	"github.com/puppetlabs/wash/cmd/internal/config"
 	"github.com/puppetlabs/wash/cmd/internal/server"
 	cmdutil "github.com/puppetlabs/wash/cmd/util"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


cmd/rootMain.go
```diff
@@ -124,12 +117,13 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
...
-	mountpath, err := ioutil.TempDir(cachedir, "mnt")
+	// Create a temporary run space for aliases and server files.
+	rundir, err := ioutil.TempDir(cachedir, "run")
 	if err != nil {
-		cmdutil.ErrPrintf("Unable to create temporary mountpoint in %v: %v\n", cachedir, err)
+		cmdutil.ErrPrintf("Error creating temporary run location in %v: %v\n", cachedir, err)
 		return exitCode{1}
 	}
-	defer os.RemoveAll(mountpath)
+	defer os.RemoveAll(rundir)
```
MikaelSmith: This seems to be interacting poorly with the mountpoint.

ekinanp: What do you mean?

MikaelSmith: I'm getting lots of "OSXFUSE Volume" icons on my desktop, which I think means they're not being unmounted correctly.

MikaelSmith:
That was my thinking. It was interfering with demos I'm writing, so I wanted a quick fix.

PR #304, opened 2019-05-29
 Support multiple paths in `wash find`

ekinanp:
Resolves https://github.com/puppetlabs/wash/issues/280


cmd/internal/find/main.go
```diff
@@ -67,15 +67,13 @@ func Main(cmd *cobra.Command, args []string) int {
...
 	conn := client.ForUNIXSocket(config.Socket)
 	walker := newWalker(result, conn)
+	exitCode := 0
 	for _, path := range result.Paths {
-		e, err := info(conn, path)
-		if err != nil {
-			cmdutil.ErrPrintf("%v\n", err)
-			continue
+		if !walker.Walk(path) {
+			exitCode = 1
```
ekinanp: I'm wondering if this might be a bit too strict, because unlike BSD/GNU find, `wash find` makes API requests across a network (which increases the likelihood of errors). Maybe it's best to make it consistent for now, and if users find it annoying, we can revisit the exit code stuff later. I vote for the latter option.

cmd/internal/find/walker.go
```diff
@@ -25,49 +25,63 @@ func newWalker(r parser.Result, conn client.Client) *walker {
 	}
 }
 
-func (w *walker) Walk(e types.Entry) {
-	w.walk(e, 0)
+// Walk returns true if the walk is successful (i.e. does not
+// have any errors), false otherwise.
+func (w *walker) Walk(path string) bool {
```
ekinanp: There's enough code here that it's worth having some unit tests. I'll do that after this PR.

cmd/internal/find/usage.go
```diff
@@ -36,7 +36,11 @@ func Usage() string {
 	u += "\n"
 	u += "NOTE: All entry attribute primaries return false if the given entry does not have\n"
 	u += "the specified attribute. For example, the -mtime primary will always return false\n"
-	u += "if the entry does not have an mtime attribute."
+	u += "if the entry does not have an mtime attribute.\n"
+	u += "\n"
+	u += "NOTE: find exits with status 0 if all entries are processed successfully, greater\n"
```
ekinanp: Copy-pasted, verbatim, from https://serverfault.com/a/225800

cmd/internal/find/main.go
MikaelSmith: Yeah, I'm not sure I'll like it. I definitely get access denied mixed in with my searches. It's currently easy to filter that out with `>/dev/null`. But it's also reasonable to error when that happens and I can `|| true`. So 🤷‍♂ .

PR #305, opened 2019-05-29
Fix FUSE, fallback to tail, and add '-c' argument for shell

MikaelSmith:
Fix an issue with OSXFUSE where the mountpoint doesn't correctly unmount introduced by #303. Also make FUSE server shutdown more deterministic to avoid shutdown races.

Defer to `/usr/bin/tail` for plain `tail` calls to make it more useful.

Add `-c` argument for Wash shell, primarily for debugging.

### DIFF
--- a/cmd/rootMain.go
+++ b/cmd/rootMain.go
@@ -117,6 +117,17 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
 		return exitCode{1}
 	}
 
+	// Mountpath is not cleaned up correctly if removed as part of deleting rundir, so it's placed
+	// in a separate location. The server has reported that it's completely done by the time we
+	// delete rundir, so I'm not sure why it doesn't clean up correctly. Alternatively, adding a
+	// 10ms sleep after srv.Stop() seemed to let it successfully unmount (with OSXFUSE).
+	mountpath, err := ioutil.TempDir(cachedir, "mnt")
+	if err != nil {
+		cmdutil.ErrPrintf("Unable to create temporary mountpoint in %v: %v\n", cachedir, err)
+		return exitCode{1}
+	}
+	defer os.RemoveAll(mountpath)
+
 	// Create a temporary run space for aliases and server files.
 	rundir, err := ioutil.TempDir(cachedir, "run")
 	if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/rootMain.go
```diff
@@ -117,6 +117,15 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
 		return exitCode{1}
 	}
 
+	// Mountpath is not cleaned up correctly if removed as part of deleting rundir, so it's placed
+	// in a separate location.
+	mountpath, err := ioutil.TempDir(cachedir, "mnt")
```
ekinanp: Hmm, why'd you decide to leave https://github.com/puppetlabs/wash/blob/master/fuse/core.go#L156 alone (assuming that was the issue)? Just curious.

cmd/tail.go
```diff
@@ -120,8 +121,19 @@ func tailMain(cmd *cobra.Command, args []string) exitCode {
 	}
 
 	if !follow {
-		cmdutil.ErrPrintf("Please use -f, other operations are not yet supported\n")
-		return exitCode{1}
+		// Defer to `/usr/bin/tail`
```
ekinanp: Should this be documented somewhere?

cmd/rootMain.go
MikaelSmith: What do you mean "leave it alone"? I made some changes there, but it still didn't work. I couldn't find any leftover goroutines that might be finishing up there, so confused what's going on. May be an actual thread that the fuse library or libfuse is running that I can't find.

cmd/tail.go
MikaelSmith: Yeah, definitely. I'll add it to the long help.

cmd/rootMain.go
ekinanp: Sorry, that comment was made before I saw your next commit.

fuse/core.go
```diff
@@ -164,17 +164,12 @@ func ServeFuseFS(filesys *plugin.Registry, mountpoint string) (chan<- context.Co
 		return nil, nil, err
 	}
 
-	// Start the FUSE server
+	// Start the FUSE server. We use the fuseExitedCh to catch externally triggered unmounts.
+	// If we're explicitly asked to shutdown the server, we want to wait until both Unmount and
+	// Serve have exited before signaling completion.
 	fuseServerStoppedCh := make(chan struct{})
```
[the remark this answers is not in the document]
MikaelSmith: Done.

cmd/rootMain.go
```diff
@@ -14,6 +15,9 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// rootCommandFlag is associated with the `-c` option of the root command, set in root.go.
+var rootCommandFlag string
```
ekinanp: Oh, was thinking in `root.go` b/c my first question was "Where is this defined?" when I saw it there. Guess a comment should go in both places, but I won't block the PR on it.

MikaelSmith: Oh, ok.

PR #306, opened 2019-05-30
Provide IsInteractive test for plugins

MikaelSmith:
Provide an IsInteractive function to test whether Wash is running in an
interactive mode so plugins can decide whether to prompt for input. This
prevents blocking when running a script via wash.


plugin/helpers.go
```diff
@@ -77,4 +79,19 @@ func TrackTime(start time.Time, name string) {
...
+
+// InitInteractive is used by Wash commands to set option-specific overrides. Only sets
+// interactivity to true if it already was and 'init' is also true.
+func InitInteractive(init bool) {
+	isInteractive = init && isInteractive
+}
+
+// IsInteractive returns true if Wash is running as an interactive session. If false, please don't
+// prompt for input on stdin.
+func IsInteractive() bool {
```
ekinanp: How about `plugin.Prompt(msg) (string, error)` ? That encapsulates all of this interactivity stuff, which lets us say "Use plugin.Prompt whenever you want to prompt something from the user."

ekinanp:
I think we should spend some time on properly handling authentication so that the MFA stuff can be generalized to external plugins. That way, users who are concerned about exposing sensitive stuff through Wash can still use it via the external plugin interface knowing full well that Wash has an API for handling authentication.

plugin/helpers.go
MikaelSmith: Hmm, could add that. I still want the IsInteractive call for cmd to avoid duplicating it.

PR #308, opened 2019-05-30
Add unit tests for `wash find`'s walker + main function

ekinanp:

### DIFF
--- a/cmd/util/client.go
+++ b/cmd/util/client.go
@@ -1,0 +1,12 @@
+package cmdutil
+
+import (
+	"github.com/puppetlabs/wash/api/client"
+	"github.com/puppetlabs/wash/cmd/internal/config"
+)
+
+// NewClient returns a new Wash client for the given subcommand.
+// Tests can set NewClient to a stub that returns a mock client.
+var NewClient = func() client.Client {
+	return client.ForUNIXSocket(config.Socket)
+}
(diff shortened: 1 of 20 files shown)


cmd/util/format.go
```diff
@@ -3,15 +3,50 @@ package cmdutil
...
 )
 
+// Stdout represents Stdout
+var Stdout io.Writer = os.Stdout
+
+// Stderr represents Stderr
+var Stderr io.Writer = os.Stderr
+
+// ColoredStderr represents a color supporting writer for Stderr
+var ColoredStderr io.Writer = color.Error
```
ekinanp: Reason for this is so the tests don't have to depend on the `color.Error` package

cmd/util/client.go
```diff
@@ -0,0 +1,12 @@
+package cmdutil
+
+import (
+	"github.com/puppetlabs/wash/api/client"
+	"github.com/puppetlabs/wash/cmd/internal/config"
+)
+
+// NewClient returns a new Wash client for the given subcommand.
+// Tests can set NewClient to a stub that returns a mock client.
+var NewClient = func() client.Client {
```
ekinanp: Once we move the `api/fs` stuff over to the client, this can become `NewClient(withLocalFSSupport bool)`.

Issue #289, opened 2019-05-23
Support plugin-specific config in Wash config


ekinanp:
Examples of plugin-specific config include MFA token durations for the AWS plugin; valid SSH keys for an SSHFS plugin, etc. Ideally, each plugin's config would look something like
```
<plugin_name>:
    <hash describing plugin config>
```

For our SSHFS/AWS examples, we'd have something like:
```
aws:
    mfa_duration: <time>
sshfs:
    valid_keys:
      - key1
      - key2
...
```

Wash would then pass the plugin-specific config along to the appropriate plugin.

This issue is an open-ended design-question. Some things to consider:

* The design should work with external plugins.

* The design should make it easy for plugin authors to access this config. For example, it should be easy for an AWS plugin author to be able to access EC2-specific config when manipulating EC2 instances; S3-specific config when manipulating S3 buckets/objects, etc.

timidri:
@ekinanp This issue is closed, but what is its replacement? Just curious where the plugin-specific config is on the roadmap.

ekinanp:
@timidri Plugin-specific config was merged in https://github.com/puppetlabs/wash/pull/314

Issue #288, opened 2019-05-23
Support enabling/disabling specific plugins in Wash config


ekinanp:
We could start by adding a single `plugins` key that lists all the plugins you want Wash to load. If the `plugins` key is not provided, then it is reasonable to default to "all" plugins.

Note that the plugins are loaded by the [Server object](https://github.com/puppetlabs/wash/blob/master/cmd/internal/server/core.go), while Wash's config is parsed in https://github.com/puppetlabs/wash/blob/master/cmd/server.go#L87-L119.

The config is documented in https://github.com/puppetlabs/wash/blob/master/website/content/docs/_index.md#config

MikaelSmith:
Should this be expected to disable external plugins as well, or just the core plugins?

ekinanp:
Not sure. Think it's OK to go with core plugins for now, since external plugins are already optional.

PR #314, opened 2019-06-04
Support plugin-specific config and limiting AWS profiles

MikaelSmith:
Support plugin-specific config (#289) and use it to allow configuring which AWS profiles are available (#317).

### DIFF
--- a/cmd/internal/server/core.go
+++ b/cmd/internal/server/core.go
@@ -13,20 +13,18 @@ import (
 	"github.com/puppetlabs/wash/api"
 	"github.com/puppetlabs/wash/fuse"
 	"github.com/puppetlabs/wash/plugin"
-	"github.com/puppetlabs/wash/plugin/aws"
-	"github.com/puppetlabs/wash/plugin/docker"
-	"github.com/puppetlabs/wash/plugin/kubernetes"
 
 	log "github.com/sirupsen/logrus"
 )
 
 // Opts exposes additional configuration for server operation.
 type Opts struct {
-	CPUProfilePath      string
-	ExternalPlugins     []plugin.ExternalPluginSpec
-	LogFile             string
+	CPUProfilePath  string
+	ExternalPlugins []plugin.ExternalPluginSpec
+	LogFile         string
 	// LogLevel can be "warn", "info", "debug", or "trace".
-	LogLevel            string
+	LogLevel     string
+	PluginConfig map[string]map[string]interface{}
 }
 
 type controlChannels struct {
@@ -42,6 +40,7 @@ type Server struct {
 	logFH      *os.File
 	api        controlChannels
 	fuse       controlChannels
+	plugins    map[string]plugin.Root
 }
 
 var levelMap = map[string]log.Level{
@@ -51,9 +50,9 @@ var levelMap = map[string]log.Level{
 	"trace": log.TraceLevel,
 }
 
-// New creates a new Server.
-func New(mountpoint string, socket string, opts Opts) *Server {
-	return &Server{mountpoint: mountpoint, socket: socket, opts: opts}
+// New creates a new Server. Accepts a list of core plugins to load.
+func New(mountpoint string, socket string, plugins map[string]plugin.Root, opts Opts) *Server {
+	return &Server{mountpoint: mountpoint, socket: socket, plugins: plugins, opts: opts}
 }
 
 // Start starts the server. It returns once the server is ready.
@@ -79,7 +78,7 @@ func (s *Server) Start() error {
 	}
 
 	registry := plugin.NewRegistry()
-	loadInternalPlugins(registry)
+	s.loadInternalPlugins(registry)
 	if len(s.opts.ExternalPlugins) > 0 {
 		loadExternalPlugins(registry, s.opts.ExternalPlugins)
 	}
@@ -166,19 +165,19 @@ func (s *Server) Stop() {
 	s.shutdown()
 }
 
-func loadPlugin(registry *plugin.Registry, name string, root plugin.Root) {
+func (s *Server) loadPlugin(registry *plugin.Registry, name string, root plugin.Root) {
 	log.Infof("Loading %v", name)
-	if err := registry.RegisterPlugin(root); err != nil {
+	if err := registry.RegisterPlugin(root, s.opts.PluginConfig[name]); err != nil {
 		// %+v is a convention used by some errors to print additional context such as a stack trace
 		log.Warnf("%v failed to load: %+v", name, err)
 	}
 }
 
-func loadInternalPlugins(registry *plugin.Registry) {
+func (s *Server) loadInternalPlugins(registry *plugin.Registry) {
 	log.Debug("Loading internal plugins")
-	loadPlugin(registry, "aws", &aws.Root{})
-	loadPlugin(registry, "docker", &docker.Root{})
-	loadPlugin(registry, "kubernetes", &kubernetes.Root{})
+	for name, root := range s.plugins {
+		s.loadPlugin(registry, name, root)
+	}
 	log.Debug("Finished loading internal plugins")
 }
 
--- a/cmd/internal/server/core.go
+++ b/cmd/internal/server/core.go
@@ -80,7 +80,7 @@ func (s *Server) Start() error {
 	registry := plugin.NewRegistry()
 	s.loadInternalPlugins(registry)
 	if len(s.opts.ExternalPlugins) > 0 {
-		loadExternalPlugins(registry, s.opts.ExternalPlugins)
+		s.loadExternalPlugins(registry, s.opts.ExternalPlugins)
 	}
 	if len(registry.Plugins()) == 0 {
 		return fmt.Errorf("No plugins loaded")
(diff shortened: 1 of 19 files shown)


MikaelSmith:
This needs a better way of handling config. It lets you restrict which profiles are loaded so you don't have to enter MFA tokens a bunch or mount profiles don't want to use.

MikaelSmith:
What I would like this to look like is
```
aws:
  profiles:
    - eso
```

where each plugin has a top-level key automatically associated with it, and the contents of that key are passed to the plugin `Init()` function.

plugin/types.go
```diff
@@ -58,7 +58,7 @@ type Parent interface {
 // Root represents the plugin root
 type Root interface {
 	Parent
-	Init() error
+	Init([]string) error
```
ekinanp: I think this should be `Init(map[string]interface{}) error` to keep it generic.

MikaelSmith: Definitely. This was a super-quick hack to test out fixing the issue with someone at the Contributor Summit. I'm halfway through reworking it to be more generic.

ekinanp: Hmm, this may be tough with external plugins b/c their config isn't persisted after the `init` call.

MikaelSmith: The idea would be that they pass what they care about along as part of `state`.

MikaelSmith:
I'll move to review once I've added external plugin support.

plugin/types.go
MikaelSmith: Chicken-and-egg problem though: init is called before we know the name of the plugin. So these seem like they need to be two different calls.

MikaelSmith: I guess it could be an environment variable we set after calling `init`. Just hard if you want to use that data during the `init` call.

ekinanp: Think two different calls make more sense. Something like `<plugin_script> root` then `<plugin_script> init <root_path> <root_state> <config>` ? 

I wonder if we should just do `<plugin_script>` and require it to return the root? So if no args are passed, people know that means they need to return the root JSON. Then we maintain symmetry by telling people that `<plugin_script> <method>` will always be invoked on an entry.

Two separate calls are also nice because we can then say "we'll only invoke `init` if your root object implements it". Then people writing simple plugins don't have to worry about it.

MikaelSmith: I'm really tempted to derive the name from the file name. It would make this quite a bit simpler.

ekinanp: Yeah, I think that might be easier. I don't like having "name" being referenced twice (once from the file and once from the root JSON outputted by `init`), but it's less things we have to tell people and it does seem like naming the script as `<plugin_name>.<ext>` is idiomatic (e.g. the `sshfs.sh` example does this).

plugin/externalPluginRoot.go
```diff
@@ -42,18 +48,22 @@ func (r *externalPluginRoot) Init() error {
...
 			err,
 			stdout,
-			"{\"name\":\"plugin_name\",\"methods\":[\"list\"]}",
+			"{}",
 		)
 	}
+
+	// Fill in required fields with data we already know.
+	decodedRoot.Name = r.Name()
+	decodedRoot.Methods = []string{"list"}
```
MikaelSmith: This would need to be changed to not overwrite Methods if it was supplied when we implement the data-driven hierarchy described in https://github.com/puppetlabs/wash/projects/9.

ekinanp:
May as well mention something about adding plugin config in the PR title

cmd/internal/server/core.go
```diff
@@ -13,20 +13,18 @@ import (
...
-	CPUProfilePath      string
-	ExternalPlugins     []plugin.ExternalPluginSpec
-	LogFile             string
+	CPUProfilePath  string
+	ExternalPlugins []plugin.ExternalPluginSpec
+	LogFile         string
 	// LogLevel can be "warn", "info", "debug", or "trace".
-	LogLevel            string
+	LogLevel string
+	Config   map[string]map[string]interface{}
```
ekinanp: `PluginConfig` reflects how this is being used.

cmd/internal/server/core.go
```diff
@@ -52,8 +51,8 @@ var levelMap = map[string]log.Level{
 }
 
 // New creates a new Server.
-func New(mountpoint string, socket string, opts Opts) *Server {
-	return &Server{mountpoint: mountpoint, socket: socket, opts: opts}
+func New(mountpoint string, socket string, plugins map[string]plugin.Root, opts Opts) *Server {
```
ekinanp: Is this meant to facilitate configuring which internal plugins Wash should load?

plugin/aws/root.go
```diff
@@ -60,10 +61,25 @@ func exists(path string) error {
 }
 
 // Init for root
-func (r *Root) Init() error {
+func (r *Root) Init(cfg map[string]interface{}) error {
 	r.EntryBase = plugin.NewEntry("aws")
 	r.SetTTLOf(plugin.ListOp, 1*time.Minute)
 
+	if profsI, ok := cfg["profiles"]; ok {
```
ekinanp: Why not have a `Config` struct that you can marshal this into?

plugin/externalPluginRoot.go
ekinanp: Bit worried that we're assuming too much too soon re: list being the only operation roots can support. Think we should keep `init` the way it is for now, and maybe add some validation to make sure that the returned name matches what we got?

plugin/types.go
```diff
@@ -55,10 +55,10 @@ type Parent interface {
 	List(context.Context) ([]Entry, error)
 }
 
-// Root represents the plugin root
+// Root represents the plugin root. The Init function is passed a configuration map.
```
ekinanp: Nit: "configuration map representing plugin-specific config"

plugin/externalPluginRoot.go
MikaelSmith: I was erring away from handling backwards-compatibility, but I guess in this case it's simple enough to warn about potential issues.

cmd/internal/server/core.go
MikaelSmith: It could be used that way. Primarily I was moving the definition of what plugins are available elsewhere, and didn't want others to rely on it as a global.

plugin/aws/root.go
MikaelSmith: Didn't think of it.

cmd/internal/server/core.go
ekinanp: Ah ok. Was going to suggest an `InternalPlugins`/`CorePlugins` option to keep it symmetric with `ExternalPlugins`. Bit weird to see something generic like `plugins` in the params and then have an `ExternalPlugins` option.

plugin/aws/root.go
MikaelSmith: Also it gets messy unless I remarshal to JSON and unmarshal again.

ekinanp: Hm, maybe we could extend `Root` with a `ConfigStruct` method to return a base config object so we can handle the marshalling for them? Then plugin authors just have to type-assert the passed-in struct.

So something like `Init(config interface{})`, where `config` is their specified struct.

MikaelSmith: I think it's not much more complicated than marshal/unmarshal, and provides clearer errors.

ekinanp: Either approach is OK with me.

plugin/externalPluginRoot.go
ekinanp: Let's do that then. Would rather keep `init`'s output symmetric with a single entry JSON in `list`'s output.

cmd/internal/server/core.go
MikaelSmith: We may want to move towards that symmetry when implementing #288.

ekinanp: Sounds good.

MikaelSmith:
Updated.

plugin/aws/root.go
MikaelSmith: The ConfigStruct concept moves the conversion around but we'd still have to handle it. Because we're not originally unmarshaling JSON, we're calling `viper.GetStringMap()`. So an improvement I'm tempted to defer for now.

timidri:
> Since Wash is dynamic, I'm worried that putting the config in `Init` might not be enough. For example, if you'd like to have some S3 or EC2 specific config, it'd be annoying as a plugin author to have to pass around a config struct to your entries. But I'm OK with worrying about that later (if users request it), so having the config in `Init` is simple enough.

@ekinanp I have actually hit exactly this issue. I am developing a plugin to interact with a Puppet Enterprise install - querying PuppetDB about its nodes, facts, resources etc. It would be great if I can put credentials for the PE instances into the wash plugin config, but if it's only passed to the init method, I will need to return it in state for every single entry, which makes it kinda hard to use. Could the config info be passed to every method instead?

MikaelSmith:
I was also considering environment variable. But feel free to open an issue; ideally propose where in the call signature the config should be passed in.

timidri:
@MikaelSmith Created issue: https://github.com/puppetlabs/wash/issues/349

Issue #291, opened 2019-05-23
Way for plugins to self-document their hierarchy


ekinanp:
Allows wash to expose a schema of the hierarchy to document what the plugin provides. Use it to optimize `wash find`. This is also useful in understanding how plugins are structured when developing and using them.

- [x] Add schema description to core plugins (#321)
- [x] Implement `stree` (#325)
- [x] Optimize `-action` with `wash find` (#333)
- [x] Extend schema to include metadata content (#351)
- [x] Optimize `-meta` with `wash find`
- [x] Support external plugin schemas (#364)
- [x] Update documentation (including adding notes about the `stree` command) (#365)

ekinanp:
The goal here is something like
```
$ schema aws
aws
    profile
        resources
            ec2
                instances
                    instance
                        metadata.json
                        consoleLog
            s3
                bucket
                    object
                    objectPrefix
                        object
                        objectPrefix    
```

i.e., given an arbitrary entry's path, the `schema` command would list the schema of the FS rooted at that entry. For example, if given an EC2 instance, then `schema` would list something like:
```
$ schema aws/demo/resources/ec2/instances/foo
aws/demo/resources/ec2/instances/foo
    metadata.json
    consoleLog
```

ekinanp:
Some TODOs that need to be finished after https://github.com/puppetlabs/wash/pull/321 is merged.

- [ ] Implement `stree`
- [ ] Optimize some parts of `wash find`
- [ ] Support external plugin schemas
- [ ] Update documentation

MikaelSmith:
We could include a note that stree can be useful for visualizing how Wash sees your plugin as you write it.

PR #321, opened 2019-06-10
Add plugin.EntrySchema + schema endpoint

ekinanp:


plugin/aws/ec2Dir.go
```diff
@@ -17,12 +17,18 @@ type ec2Dir struct {
 	entries []plugin.Entry
 }
 
-func newEC2Dir(session *session.Session) *ec2Dir {
+func ec2DirTemplate() *ec2Dir {
```
ekinanp: These `Template` functions are here because Go doesn't support (simple) meta-programming. In Ruby, this would look something like
```
class EC2Dir < Wash::Entry
  singleton "ec2"
  parent_of EC2InstancesDir

  def list
  end

  ...
end
```

where `Wash::Entry` would have a `schema` method that would check for these class-level tags and construct the schema accordingly (that's also how external plugins would support `schema` -- the entries would include an implemented "schema" method. We'll need to handle the case when entries don't implement `schema` to make this easy for Bash, or maybe Bash plugins can't access the schema functionality).

If/when we add attributes to this list, we'd likely have `ec2Dir.SupportedAttributes("ctime", "mtime")` which would translate to class-level `ctime`, `mtime` tags in Ruby, so e.g. something like
```
class EC2Dir < Wash::Entry
  attributes :ctime, :mtime
end
```
(which would then be `@ctime` and `@mtime` fields in the class).

ekinanp:
Here's an example of the `schema` endpoint's response:
```
{
  "type": "plugin.Registry",
  "short_type": "",
  "singleton": false,
  "actions": [
    "list"
  ],
  "children": [
    {
      "type": "aws.Root",
      "short_type": "aws",
      "singleton": true,
      "actions": [
        "list"
      ],
      "children": [
        {
          "type": "aws.profile",
          "short_type": "profile",
          "singleton": false,
          "actions": [
            "list"
          ],
          "children": [
            {
              "type": "aws.resourcesDir",
              "short_type": "resources",
              "singleton": true,
              "actions": [
                "list"
              ],
              "children": [
                {
                  "type": "aws.s3Dir",
                  "short_type": "s3",
                  "singleton": true,
                  "actions": [
                    "list"
                  ],
                  "children": [
                    {
                      "type": "aws.s3Bucket",
                      "short_type": "bucket",
                      "singleton": false,
                      "actions": [
                        "list"
                      ],
                      "children": [
                        {
                          "type": "aws.s3ObjectPrefix",
                          "short_type": "prefix",
                          "singleton": false,
                          "actions": [
                            "list"
                          ],
                          "children": [
                            {
                              "type": "aws.s3ObjectPrefix",
                              "short_type": "prefix",
                              "singleton": false,
                              "actions": [
                                "list"
                              ],
                              "children": null
                            },
                            {
                              "type": "aws.s3Object",
                              "short_type": "object",
                              "singleton": false,
                              "actions": [
                                "read",
                                "stream"
                              ],
                              "children": null
                            }
                          ]
                        },
                        {
                          "type": "aws.s3Object",
                          "short_type": "object",
                          "singleton": false,
                          "actions": [
                            "read",
                            "stream"
                          ],
                          "children": null
                        }
                      ]
                    }
                  ]
                },
                {
                  "type": "aws.ec2Dir",
                  "short_type": "ec2",
                  "singleton": true,
                  "actions": [
                    "list"
                  ],
                  "children": [
                    {
                      "type": "aws.ec2InstancesDir",
                      "short_type": "instances",
                      "singleton": true,
                      "actions": [
                        "list"
                      ],
                      "children": [
                        {
                          "type": "aws.ec2Instance",
                          "short_type": "instance",
                          "singleton": false,
                          "actions": [
                            "list",
                            "exec"
                          ],
                          "children": [
                            {
                              "type": "aws.ec2InstanceConsoleOutput",
                              "short_type": "console.out",
                              "singleton": false,
                              "actions": [
                                "read"
                              ],
                              "children": null
                            },
                            {
                              "type": "aws.ec2InstanceMetadataJSON",
                              "short_type": "metadata.json",
                              "singleton": true,
                              "actions": [
                                "read"
                              ],
                              "children": null
                            },
                            {
                              "type": "volume.FS",
                              "short_type": "fs",
                              "singleton": false,
                              "actions": [
                                "list"
                              ],
                              "children": [
                                {
                                  "type": "volume.dir",
                                  "short_type": "dir",
                                  "singleton": false,
                                  "actions": [
                                    "list"
                                  ],
                                  "children": [
                                    {
                                      "type": "volume.dir",
                                      "short_type": "dir",
                                      "singleton": false,
                                      "actions": [
                                        "list"
                                      ],
                                      "children": null
                                    },
                                    {
                                      "type": "volume.file",
                                      "short_type": "file",
                                      "singleton": false,
                                      "actions": [
                                        "read",
                                        "stream"
                                      ],
                                      "children": null
                                    }
                                  ]
                                },
                                {
                                  "type": "volume.file",
                                  "short_type": "file",
                                  "singleton": false,
                                  "actions": [
                                    "read",
                                    "stream"
                                  ],
                                  "children": null
                                }
                              ]
                            }
                          ]
                        }
                      ]
                    }
                  ]
                }
              ]
            }
          ]
        }
      ]
    },
    {
      "type": "docker.Root",
      "short_type": "docker",
      "singleton": true,
      "actions": [
        "list"
      ],
      "children": [
        {
          "type": "docker.containersDir",
          "short_type": "containers",
          "singleton": true,
          "actions": [
            "list"
          ],
          "children": [
            {
              "type": "docker.container",
              "short_type": "container",
              "singleton": false,
              "actions": [
                "list",
                "exec"
              ],
              "children": [
                {
                  "type": "docker.containerLogFile",
                  "short_type": "log",
                  "singleton": true,
                  "actions": [
                    "read",
                    "stream"
                  ],
                  "children": null
                },
                {
                  "type": "docker.containerMetadata",
                  "short_type": "metadata.json",
                  "singleton": true,
                  "actions": [
                    "read"
                  ],
                  "children": null
                },
                {
                  "type": "volume.FS",
                  "short_type": "fs",
                  "singleton": false,
                  "actions": [
                    "list"
                  ],
                  "children": [
                    {
                      "type": "volume.dir",
                      "short_type": "dir",
                      "singleton": false,
                      "actions": [
                        "list"
                      ],
                      "children": [
                        {
                          "type": "volume.dir",
                          "short_type": "dir",
                          "singleton": false,
                          "actions": [
                            "list"
                          ],
                          "children": null
                        },
                        {
                          "type": "volume.file",
                          "short_type": "file",
                          "singleton": false,
                          "actions": [
                            "read",
                            "stream"
                          ],
                          "children": null
                        }
                      ]
                    },
                    {
                      "type": "volume.file",
                      "short_type": "file",
                      "singleton": false,
                      "actions": [
                        "read",
                        "stream"
                      ],
                      "children": null
                    }
                  ]
                }
              ]
            }
          ]
        },
        {
          "type": "docker.volumesDir",
          "short_type": "volumes",
          "singleton": true,
          "actions": [
            "list"
          ],
          "children": [
            {
              "type": "docker.volume",
              "short_type": "volume",
              "singleton": false,
              "actions": [
                "list"
              ],
              "children": [
                {
                  "type": "volume.dir",
                  "short_type": "dir",
                  "singleton": false,
                  "actions": [
                    "list"
                  ],
                  "children": [
                    {
                      "type": "volume.dir",
                      "short_type": "dir",
                      "singleton": false,
                      "actions": [
                        "list"
                      ],
                      "children": null
                    },
                    {
                      "type": "volume.file",
                      "short_type": "file",
                      "singleton": false,
                      "actions": [
                        "read",
                        "stream"
                      ],
                      "children": null
                    }
                  ]
                },
                {
                  "type": "volume.file",
                  "short_type": "file",
                  "singleton": false,
                  "actions": [
                    "read",
                    "stream"
                  ],
                  "children": null
                }
              ]
            }
          ]
        }
      ]
    },
    {
      "type": "kubernetes.Root",
      "short_type": "kubernetes",
      "singleton": true,
      "actions": [
        "list"
      ],
      "children": [
        {
          "type": "kubernetes.k8context",
          "short_type": "context",
          "singleton": false,
          "actions": [
            "list"
          ],
          "children": [
            {
              "type": "kubernetes.namespace",
              "short_type": "namespace",
              "singleton": false,
              "actions": [
                "list"
              ],
              "children": [
                {
                  "type": "kubernetes.podsDir",
                  "short_type": "pods",
                  "singleton": true,
                  "actions": [
                    "list"
                  ],
                  "children": [
                    {
                      "type": "kubernetes.pod",
                      "short_type": "pod",
                      "singleton": false,
                      "actions": [
                        "read",
                        "stream",
                        "exec"
                      ],
                      "children": null
                    }
                  ]
                },
                {
                  "type": "kubernetes.pvcsDir",
                  "short_type": "persistentvolumeclaims",
                  "singleton": true,
                  "actions": [
                    "list"
                  ],
                  "children": [
                    {
                      "type": "kubernetes.pvc",
                      "short_type": "persistentvolumeclaim",
                      "singleton": false,
                      "actions": [
                        "list"
                      ],
                      "children": [
                        {
                          "type": "volume.dir",
                          "short_type": "dir",
                          "singleton": false,
                          "actions": [
                            "list"
                          ],
                          "children": [
                            {
                              "type": "volume.dir",
                              "short_type": "dir",
                              "singleton": false,
                              "actions": [
                                "list"
                              ],
                              "children": null
                            },
                            {
                              "type": "volume.file",
                              "short_type": "file",
                              "singleton": false,
                              "actions": [
                                "read",
                                "stream"
                              ],
                              "children": null
                            }
                          ]
                        },
                        {
                          "type": "volume.file",
                          "short_type": "file",
                          "singleton": false,
                          "actions": [
                            "read",
                            "stream"
                          ],
                          "children": null
                        }
                      ]
                    }
                  ]
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

plugin/schema.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"reflect"
+	"strings"
+)
+
+// EntrySchema represents an entry's schema.
+type EntrySchema struct {
+	Type      string        `json:"type"`
+	ShortType string        `json:"short_type"`
+	Singleton bool          `json:"singleton"`
```
ekinanp: Singleton is important for `stree`. Non-singleton entries are displayed as `[<type/short_type>]`, while singleton entries are displayed as `<type/short_type>`. This way, people won't get confused into thinking there's more than one "resources" directory, or "metadata.json" file.

`Type` is meant to be the raw (language-specific) class-type. Don't really know when/if that will be used by clients (here it's used to handle cycles when constructing entry schemas). `ShortType` is important so that you avoid printing something like `ec2Dir/ec2InstancesDir/ec2instance/ec2InstanceMetadataJSON` which doesn't add any clarity.

(This is also one nice thing about the `*Template()` functions because if you're re-using a prior entry's template, you can override its short type to something else. Doing this with struct tags is difficult).

plugin/types.go
```diff
@@ -46,12 +46,18 @@ type Entry interface {
...
+	// This is used by the plugin registry to mark plugin roots as singleton
+	// objects. This way, plugin authors don't have to do this themselves.
+	markSingleton()
 }
 
 // Parent is an entry with children. It will be represented as a directory in the Wash
 // filesystem.
 type Parent interface {
 	Entry
+	ChildSchemas() []EntrySchema
```
ekinanp: Reason for `ChildSchemas` is to force core plugin authors (at compile-time) to provide this info as I don't see a good reason why people shouldn't document their plugins while writing them.

plugin/aws/ec2Dir.go
ekinanp: Think we could include the cache TTLs in the schema, since they're also class-level params. Would be useful if someone's curious about how long things are cached in Wash.

ekinanp: We could also include meta schemas here to validate meta primary queries. Idea is to have a method that takes in an empty object/struct, marshal it (for the purpose of obtaining its fields), and work from there. Doing this could also remove the (annoying) behavior of defaulting maxdepth to 1, and having to set it to -1 for meta to recurse.

plugin/schema.go
ekinanp: Am open to a better name than `ShortType`.

plugin/entryBase.go
```diff
@@ -39,18 +39,17 @@ type EntryBase struct {
...
 }
 
 // NewEntry creates a new entry
-func NewEntry(name string) EntryBase {
-	if name == "" {
-		panic("plugin.NewEntry: received an empty name")
-	}
-
+//
+// TODO: Update docs with the changes to NewEntry
```
MikaelSmith: TODOs like this make it really hard to review a work by individual commits. If you're going to resolve the TODO in a single body of work, I'd rather it amend the commit where the TODO was introduced.

Also, there's no explanation about why you're fundamentally changing the interaction with plugin.EntryBase. Name feels like a fundamental part of an entry, so why is it no longer required?

plugin/schema.go
MikaelSmith: This mostly seems like it should be inline documentation.

plugin/schema.go
```diff
@@ -0,0 +1,82 @@
...
+	Actions   []string      `json:"actions"`
+	Children  []EntrySchema `json:"children"`
+	entry     Entry
+}
+
+// ChildSchemas is a helper that's used to implement Parent#ChildSchemas
+func ChildSchemas(childTemplates ...Entry) []EntrySchema {
```
MikaelSmith: I'm not sure Entry is the right thing to include here if we have to change NewEntry to satisfy it. I'd be fine with a narrower interface.

api/schema.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/puppetlabs/wash/plugin"
+)
+
+var schemaHandler handler = func(w http.ResponseWriter, r *http.Request) *errorResponse {
```
MikaelSmith: This should have annotations for API docs.

plugin/entryBase.go
MikaelSmith: I haven't reviewed all the changes to specific plugins yet because this question feels fairly fundamental to the PR.

plugin/schema.go
MikaelSmith: short type seems more akin to a description. What's the expectation for how it's used?

plugin/entryBase.go
ekinanp: Main reason for the change is because Go's metaprogramming sucks (see my comment in https://github.com/puppetlabs/wash/pull/321#discussion_r292224305). TLDR is that the schema stuff is equivalent to class-level annotations in other languages, but Go has no way of doing that easily and in a way that isn't error-prone/fragile. So, the somewhat "hacky" way for Wash to figure out what an Entry looks like is to receive a "stub"/"template" object containing all the class-level configuration. That "stub/template" object is also the basis for creating specific instances of those objects (e.g. like EC2 instances, containers, S3 buckets, etc.) -- that's why the plugin-specific changes have things like `containerTemplate()`, `ec2InstanceTemplate()` (recommend reviewing the Docker changes).

`NewEntry` established an initial EntryBase template that you can then extend with your plugin-specific config like caching. So, it should really be `NewEntryTemplate()`. I can change its name to reflect that.

Biggest fundamental change for core plugin design is you have to specify entry schemas, hence why I made `ChildSchemas` part of the Parent interface. I also wanted a way to make it clear, at the type-level, that `ChildSchemas` returned schemas, which is why I didn't name it `ChildTemplates` (that's why `plugin.ChildSchemas` is there so you can at least do everything in one-two lines of code while getting that type-level safety). So the pattern for core plugin authors now is (1) Set-up a template object for your entry that you will then pass-into the parent's ChildSchema method; (2) Extend that template-object with instance-specific info like the name, attributes, etc.). Hence why there's `containerTemplate()`, then `newContainer()` in e.g. the Docker plugin.

MikaelSmith: Could we have separate `NewEntry` and `NewEntryTemplate` functions, where `NewEntry` still takes the name as an argument? That would help distinguish between different types of use.

plugin/schema.go
ekinanp: Maybe `singleton` could be inline, but then a Wash user would have to figure that out themselves.

The intended use for `singleton` and `short_type`. For example
```
stree mnt/docker
mnt/docker
  containers
    [container]
      metadata.json
      log
      fs
        [dir]
          [dir]
          [file]
        [file]
  volumes
    [volume]
      [dir]
        [dir]
        [file]
      [file]
```

(`short_type` would be what's printed as "file"/"dir"/"volumes"/"container"; singleton entries don't have the `[]` around them).

plugin/aws/ec2InstanceConsoleOutput.go
```diff
@@ -19,16 +19,25 @@ type ec2InstanceConsoleOutput struct {
...
 }
 
-func newEC2InstanceConsoleOutput(ctx context.Context, inst *ec2Instance, latest bool) (*ec2InstanceConsoleOutput, error) {
+func ec2InstanceConsoleOutputTemplate() *ec2InstanceConsoleOutput {
 	cl := &ec2InstanceConsoleOutput{
-		inst:   inst,
-		latest: latest,
+		EntryBase: plugin.NewEntry(),
 	}
+	cl.SetShortType("console.out")
```
MikaelSmith: Why isn't this a singleton?

plugin/entryBase.go
ekinanp: Hmm, maybe. It'd be weird explaining why `newContainer()` uses a different EntryBase object than `containerTemplate()` though (when it shouldn't, `containerTemplate`'s EntryBase is the right EntryBase object).

What's wrong with `SetName`?

ekinanp: I do think having a `NewEntryTemplate`/`EntryTemplate` method makes sense, just not sure that we should have `NewEntry` as well.

api/schema.go
ekinanp: I can add that once we agree on what the `EntrySchema` object should look like.

plugin/entryBase.go
MikaelSmith: It feels like conflating two concepts: the template, vs a named entry. And I don't care for the extra method call for something that's going to be used almost every time. An Entry has a name, while you're re-using it for something that's entry-like (a template of an entry).

MikaelSmith: In my mind, `NewEntryTemplate` creates an a skeleton of the entry type. When you name it, it becomes an instance, and I feel like `NewEntry` represents that concept of a named entry.

plugin/schema.go
ekinanp: Think this was addressed in https://github.com/puppetlabs/wash/pull/321#discussion_r292613928

I guess I could do a type-alias `type EntryTemplate = Entry`, but that's an extra thing people have to remember. Then again, it might make the method documentation clearer (and explicitly indicate that it's separate from `plugin.Schema`).

plugin/aws/ec2InstanceConsoleOutput.go
```diff
@@ -19,16 +19,25 @@ type ec2InstanceConsoleOutput struct {
...
+}
+
+func newEC2InstanceConsoleOutput(ctx context.Context, inst *ec2Instance, latest bool) (*ec2InstanceConsoleOutput, error) {
+	cl := ec2InstanceConsoleOutputTemplate()
+	cl.inst = inst
+	cl.latest = latest
 
 	if cl.latest {
-		cl.EntryBase = plugin.NewEntry("console-latest.out")
+		cl.EntryBase = plugin.NewEntryBase()
```
MikaelSmith: Why does this line exist? We've already initialized it in ec2InstanceConsoleOutputTemplate.

ekinanp: Yeah, that was in-flight work. Should be removed now.

api/schema.go
ekinanp: Nevermind, I'll add all of that once `stree` + some `wash find` optimizations are finished.

plugin/docker/containersDir.go
```diff
@@ -0,0 +1,48 @@
+package docker
```
MikaelSmith: This kind of rename I like to break out into a separate commit. But not worth redoing now.

MikaelSmith:
Supports #291.

PR #325, opened 2019-06-12
Add the stree command

ekinanp:

### DIFF
--- a/api/types/schema.go
+++ b/api/types/schema.go
@@ -1,0 +1,7 @@
+package apitypes
+
+import "github.com/puppetlabs/wash/plugin"
+
+// EntrySchema describes an entry's schema, which
+// is what's returned by the /fs/schema endpoint.
+type EntrySchema = plugin.EntrySchema
--- a/cmd/stree.go
+++ b/cmd/stree.go
@@ -1,0 +1,61 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/xlab/treeprint"
+
+	apitypes "github.com/puppetlabs/wash/api/types"
+	cmdutil "github.com/puppetlabs/wash/cmd/util"
+	"github.com/spf13/cobra"
+)
+
+func streeCommand() *cobra.Command {
+	streeCmd := &cobra.Command{
+		Use:   "stree [<path>...]",
+		Short: "Displays the entry's stree (schema-tree)",
+		Long: `Displays the entry's stree (schema-tree), which is a high-level overview of the entry's
+hierarchy. Non-singleton types are bracketed with "[]".`,
+		RunE: toRunE(streeMain),
+	}
+	return streeCmd
+}
+
+func streeMain(cmd *cobra.Command, args []string) exitCode {
+	paths := args
+	if len(paths) == 0 {
+		paths = []string{"."}
+	}
+	conn := cmdutil.NewClient()
+	for _, path := range paths {
+		schema, err := conn.Schema(path)
+		if err != nil {
+			cmdutil.ErrPrintf("%v\n", err)
+			return exitCode{1}
+		}
+		stree := treeprint.New()
+		fill(stree, schema)
+		stree.SetValue(path)
+		cmdutil.Print(stree.String())
+	}
+	return exitCode{0}
+}
+
+func fill(stree treeprint.Tree, schema apitypes.EntrySchema) treeprint.Tree {
+	value := schema.Label
+	if len(value) == 0 {
+		value = schema.Type
+	}
+	if !schema.Singleton {
+		value = fmt.Sprintf("[%v]", value)
+	}
+	stree.SetValue(value)
+	for _, child := range schema.Children {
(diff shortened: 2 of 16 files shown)


ekinanp:
Example output:
```
➜  mnt616562680 stree
.
├── aws
│   └── [profile]
│       └── resources
│           ├── s3
│           │   └── [bucket]
│           │       ├── [prefix]
│           │       │   ├── [prefix]
│           │       │   └── [object]
│           │       └── [object]
│           └── ec2
│               └── instances
│                   └── [instance]
│                       ├── [console.out]
│                       ├── metadata.json
│                       └── fs
│                           ├── [dir]
│                           │   ├── [dir]
│                           │   └── [file]
│                           └── [file]
├── docker
│   ├── containers
│   │   └── [container]
│   │       ├── log
│   │       ├── metadata.json
│   │       └── fs
│   │           ├── [dir]
│   │           │   ├── [dir]
│   │           │   └── [file]
│   │           └── [file]
│   └── volumes
│       └── [volume]
│           ├── [dir]
│           │   ├── [dir]
│           │   └── [file]
│           └── [file]
└── kubernetes
    └── [context]
        └── [namespace]
            ├── pods
            │   └── [pod]
            └── persistentvolumeclaims
                └── [persistentvolumeclaim]
                    ├── [dir]
                    │   ├── [dir]
                    │   └── [file]
                    └── [file]
```

With multiple files:
```
➜  mnt633914967 stree aws docker
aws
└── [profile]
    └── resources
        ├── s3
        │   └── [bucket]
        │       ├── [prefix]
        │       │   ├── [prefix]
        │       │   └── [object]
        │       └── [object]
        └── ec2
            └── instances
                └── [instance]
                    ├── [console.out]
                    ├── metadata.json
                    └── fs
                        ├── [dir]
                        │   ├── [dir]
                        │   └── [file]
                        └── [file]
docker
├── containers
│   └── [container]
│       ├── log
│       ├── metadata.json
│       └── fs
│           ├── [dir]
│           │   ├── [dir]
│           │   └── [file]
│           └── [file]
└── volumes
    └── [volume]
        ├── [dir]
        │   ├── [dir]
        │   └── [file]
        └── [file]
```

cmd/stree.go
```diff
@@ -0,0 +1,61 @@
...
+	apitypes "github.com/puppetlabs/wash/api/types"
+	cmdutil "github.com/puppetlabs/wash/cmd/util"
+	"github.com/spf13/cobra"
+)
+
+func streeCommand() *cobra.Command {
+	streeCmd := &cobra.Command{
+		Use:   "stree [<path>...]",
+		Short: "Displays the entry's stree (schema-tree)",
+		Long: `Displays the entry's stree (schema-tree), which is a high-level overview of the entry's
```
ekinanp: Think this could be more complete, but not sure how to word this better.

cmd/stree.go
```diff
@@ -0,0 +1,61 @@
...
+		fill(stree, schema)
+		stree.SetValue(path)
+		cmdutil.Print(stree.String())
+	}
+	return exitCode{0}
+}
+
+func fill(stree treeprint.Tree, schema apitypes.EntrySchema) treeprint.Tree {
+	value := schema.Label
+	if len(value) == 0 {
```
ekinanp: Think we should just require `Label` (i.e. panic if it is not provided). I see no good reason to default to printing Type.

api/types/schema.go
```diff
@@ -0,0 +1,7 @@
+package apitypes
+
+import "github.com/puppetlabs/wash/plugin"
+
+// EntrySchema describes an entry's schema, which
+// is what's returned by the /fs/schema endpoint.
+type EntrySchema = plugin.EntrySchema
```
ekinanp: I kind of want to adopt this pattern for `plugin.EntryAttributes` + the actions. It'd be nice if the cmd's did not depend on the `plugin` package, but I can see the redundancy behind that (esp. w.r.t the actions) so am happy to remove this if that's preferred.

cmd/stree.go
```diff
@@ -0,0 +1,61 @@
...
+		value = schema.Type
+	}
+	if !schema.Singleton {
+		value = fmt.Sprintf("[%v]", value)
```
ekinanp: We could add some more info here (e.g. the entry's supported actions), but not sure if that'd be useful or if we should just make that an option.

api/types/schema.go
MikaelSmith: I'm ok with that pattern. I think I've used it sparingly elsewhere.

ekinanp: Ok. I'll put up a small cleanup PR after this to adopt this pattern where it makes sense.

cmd/stree.go
MikaelSmith: I'd definitely be interested in entry's supported actions. Making it an additional option feels consistent with other tools in the ecosystem.

ekinanp: Ok, I'll file an issue.

MikaelSmith:
Supports #291.

PR #327, opened 2019-06-12
Panic if EntrySchema#Label is not provided

ekinanp:
The Label's useful for stree. We previously defaulted to printing the
fully qualified type if Label was not provided, but there was no good
reason to do this.

### DIFF
--- a/cmd/stree.go
+++ b/cmd/stree.go
@@ -43,9 +43,6 @@ func streeMain(cmd *cobra.Command, args []string) exitCode {
 
 func fill(stree treeprint.Tree, schema apitypes.EntrySchema) treeprint.Tree {
 	value := schema.Label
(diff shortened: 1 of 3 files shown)


ekinanp:
I chose not to do the apitypes cleanup here b/c that required some newer files (`apitypes/action.go`, `apitypes.jsonObject.go`) that I don't think are necessary right now.

One thing I did notice is we use the pattern of `entry.Supports(ListAction) => List if yes` a lot, which makes me wonder if we should include an extra `IsParent`/`Parent` field in apitypes.Entry ? Kind of thinking it'd be useful to separate parents from files and that having to check list is a bit of a low-level detail, but we can also do that via list so dunno if it's necessary.

MikaelSmith:
Supports #291.

PR #328, opened 2019-06-12
Add option to configure enabled core plugins

### LINKED ISSUE #288: Support enabling/disabling specific plugins in Wash config

MikaelSmith:
Resolves #288.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -107,10 +107,22 @@ func serverOptsFor(cmd *cobra.Command) (map[string]plugin.Root, server.Opts, err
 		return nil, server.Opts{}, fmt.Errorf("failed to unmarshal the external-plugins key: %v", err)
 	}
 
-	// Copy internalPlugins so we don't mutate it.
+	// Load internal plugins that are not specifically excluded.
+	enabledPlugins := viper.GetStringSlice("plugins")
 	plugins := make(map[string]plugin.Root)
-	for name, plug := range internalPlugins {
-		plugins[name] = plug
+	if len(enabledPlugins) > 0 {
+		for _, name := range enabledPlugins {
+			if plug, ok := internalPlugins[name]; ok {
+				plugins[name] = plug
+			} else {
+				log.Warnf("Requested unknown plugin %s", name)
+			}
+		}
+	} else {
+		// Copy internalPlugins so we don't mutate it.
+		for name, plug := range internalPlugins {
+			plugins[name] = plug
+		}
 	}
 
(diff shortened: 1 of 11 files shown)


cmd/server.go
```diff
@@ -90,38 +91,78 @@ func bindServerArgs(cmd *cobra.Command, args []string) {
...
+	return plugins, server.Opts{
+		CPUProfilePath: viper.GetString("cpuprofile"),
+		LogFile:        viper.GetString("logfile"),
+		LogLevel:       viper.GetString("loglevel"),
+		PluginConfig:   config,
 	}, nil
 }
+
+// externalPluginSpec represents an external plugin's specification.
+type externalPluginSpec struct {
```
ekinanp: Hmm, I still think this should be in plugin. Otherwise if we add other options to this (like output format), that will be extra arguments to `NewExternalPluginRoot` that may eventually get translated to their own separate struct and thus result in redundant code.

MikaelSmith: I think of externalPluginSpec is solely a serialization format. I don't think it has much to do with the actual plugin implementation.

I can kind of see your point though. NewExternalPluginRoot should probably take an ExternalPluginSpec, not a string (or two strings) as is currently implemented.

ekinanp: I think of it as what clients need to know about external plugins in order to use them (it being a serialization format is just a nice consequence of that). That reason is why I liked the addition of `load` in the PR

plugin/externalPluginRoot.go
```diff
@@ -12,8 +12,8 @@ type externalPluginRoot struct {
 	*externalPluginEntry
 }
 
-// NewExternalPluginRoot returns a new plugin root given the plugin script
-func NewExternalPluginRoot(name, script string) Root {
+// newExternalPluginRoot returns a new plugin root given the plugin script
+func newExternalPluginRoot(name, script string) *externalPluginRoot {
```
ekinanp: I think we can inline this into `Load`

MikaelSmith: Done.

PR #332, opened 2019-06-15
Fix a regression introduced with volume improvements

MikaelSmith:
PR #330 introduced a regression where, when the cache expired, listing a volume could repopulate the cache with different data and break navigation. Fix so that VolumeList calls for a particular cache key always uses the same path.

### DIFF
--- a/volume/dir.go
+++ b/volume/dir.go
@@ -11,8 +11,14 @@ import (
 type dir struct {
 	plugin.EntryBase
 	impl Interface
-	key  plugin.Entry
-	path string
+	// subtreeRoot represents the location we started searching for this subtree, which is used both as
+	// a cache key for that search and to be able to repeat the search. We must always query the same
+	// path when running VolumeList for its related key. If we didn't, we might start with a cache at
+	// '/', then later refill the same cache entry with a hierarchy starting at '/foo'. If we used that
+	// new cache data to try and list '/', we'd only get back a directory containing 'foo' and omit any
+	// other files in '/' because they wouldn't be in the cache at the time.
+	subtreeRoot *dir
+	path        string
 }
 
 func dirBase() *dir {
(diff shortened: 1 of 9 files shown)


volume/dir.go
```diff
@@ -7,11 +7,18 @@ import (
 	"github.com/puppetlabs/wash/plugin"
 )
 
+// keystone represents an entry that can be used as a key, as well as the volume path associated
+// with that key. We must always query the same path when running VolumeList for its related key.
```
ekinanp: Still not quite sure I understand the issue. Some examples of where things could go wrong would be useful.

MikaelSmith: I'll work on a better comment with an example.

MikaelSmith: I wrote up what I think would happen. I haven't found a combination of commands that actually make it happen, but I think my reasoning is sound.

ekinanp: Oh I think I see the issue now (the comment helped). Entries rooted at the same node (e.g. `/`, or `/1/2/.../<maxdepth>` or `/<maxdepth+1>/.../<maxdepth+maxdepth>`) should enumerate the root whenever the cache runs out but instead they enumerate themselves.

I think this would be clearer if we changed `key plugin.Entry` => `root *dir` (or a better name than `root`, maybe `subtreeRoot` or `enumeratingParent`?). That way, `volume.List` takes in `root.path` instead of `key.string` and it's clear that the `root` is the cache key b/c it is where we're enumerating from.

MikaelSmith: Oh right, don't need a new type for it.

MikaelSmith: It's not quite that simple. The root we start from isn't a `*dir`, so I need to special-case to use impl for that, and it gets kind of messy.

MikaelSmith: Actually I don't see a way to clean that up without doing what I have here. I need to attach a path to the root entry, which means making something that wraps a plugin.Entry. But we don't want to implement `dir` that way because we need to set things on an EntryBase for it.

MikaelSmith: Updated.

PR #335, opened 2019-06-17
Bug fix setting ec2 User. Extend Identity type

eboutili:
Fixes a bug where setting User in .ssh/config wasn't working. 

Also extends the Identity type in transport/ssh.go. Currently, when using 'wash exec' on an ec2 instance, the hostname is derived automatically because it can be fetched from the instance metadata. Similarly, with these changes, the user often isn't required to define IdentityFile or User in ~/.ssh/config. The new fields in the Identity type are:

- IdentityFile. The ssh key name is fetched from metadata and appended to make an educated guess that the user's private key is ~/.ssh/keyname.pem

- FallbackUser, currently hardcoded to be the default user for Amazon Linux AMI instances, "ec2-user". (A further enhancement is planned to scrape the default user name from console output -- eg "centos" or "ubuntu" -- and use that for FallbackUser.)

Signed-off-by: Eric Boutilier <eric.boutilier@puppet.com>

### DIFF
--- a/plugin/aws/ec2Instance.go
+++ b/plugin/aws/ec2Instance.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"sync"
 	"time"
+	"os"
+	"path/filepath"
 
 	awsSDK "github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -219,7 +221,11 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
 
 	var identityfile string
 	if name, ok := meta["KeyName"]; ok {
-		identityfile = ("~/.ssh/" + name.(string) + ".pem")
+		homedir, err := os.UserHomeDir()
+		if err != nil {
+			return nil, err
+		}
+		identityfile = (filepath.Join(homedir, ".ssh", (name.(string) + ".pem")))
 	} else {
 		return nil, fmt.Errorf("No key name found for %v", inst)
 	}
--- a/plugin/aws/ec2Instance.go
+++ b/plugin/aws/ec2Instance.go
@@ -217,7 +217,13 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
 		return nil, fmt.Errorf("No public interface found for %v", inst)
 	}
 
+	var identityfile string
+	if name, ok := meta["KeyName"]; ok {
+		identityfile = ("~/.ssh/" + name.(string) + ".pem")
+	} else {
+		return nil, fmt.Errorf("No key name found for %v", inst)
+	}
 	// Use the default user for Amazon AMIs. See above for ideas on making this more general. Can be
 	// overridden in ~/.ssh/config.
-	return transport.ExecSSH(ctx, transport.Identity{Host: hostname, User: "ec2-user"}, append([]string{cmd}, args...), opts)
+	return transport.ExecSSH(ctx, transport.Identity{Host: hostname, FallbackUser: "ec2-user", IdentityFile: identityfile}, append([]string{cmd}, args...), opts)
 }
--- a/plugin/aws/ec2Instance.go
+++ b/plugin/aws/ec2Instance.go
@@ -12,6 +12,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	ec2Client "github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/puppetlabs/wash/activity"
 	"github.com/puppetlabs/wash/plugin"
 	"github.com/puppetlabs/wash/transport"
 	"github.com/puppetlabs/wash/volume"
@@ -207,7 +208,7 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
 		return nil, err
 	}
 
-	// TODO: scrape default user and authorized keys from console output. Probably only works for Amazon AMIs.
+	// TODO: scrape default user from console output.
 	// See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/connection-prereqs.html#connection-prereqs-fingerprint
 	// for some helpful defaults we could add.
 	var hostname string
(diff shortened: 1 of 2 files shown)


MikaelSmith:
Thanks for putting this together!

transport/ssh.go
```diff
@@ -47,7 +47,7 @@ func getHostKeyCallback() (ssh.HostKeyCallback, error) {
 	return knownhosts.New(filepath.Join(homedir, ".ssh", "known_hosts"))
 }
 
-func sshConnect(host, port, user string, strictHostKeyChecking bool) (*ssh.Client, error) {
+func sshConnect(host, port, user string, identityfile string, strictHostKeyChecking bool) (*ssh.Client, error) {
```
MikaelSmith: identityfile appears to be unused. Was there trivial work to use it, or is that something you're still exploring?

plugin/aws/ec2Instance.go
```diff
@@ -217,7 +217,13 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
 		return nil, fmt.Errorf("No public interface found for %v", inst)
 	}
 
+	var identityfile string
+	if name, ok := meta["KeyName"]; ok {
+		identityfile = ("~/.ssh/" + name.(string) + ".pem")
```
MikaelSmith: We should fallback to ssh-agent if the file doesn't exist.

transport/ssh.go
eboutili: Yikes, yep, thank you. Your other comment is totally related to this also, and part of the same issue really. I figured out the fix last night I think, so will be back with that change soon. Thanks for the _really_ fast review/feedback. What I should do with this (I think) is add some conditional logic for when ```sshConfig``` is getting defined. The Auth field should now be ```[]ssh.AuthMethod{ssh.PublicKeys(signer),}``` if a key name is found and ```[]ssh.AuthMethod{agent}``` if it isn't. 

For other contributors who might be reading this, my mess-up was I thought ```exec``` was connecting successfully because my code was making it see ```~/.ssh/<keyname>.pem``` But it was actually just getting the key from my running ssh-agent. IOW I was being tricked by a flawed testing cycle. You don't want to have ```AddKeysToAgent``` set to ```yes``` globally  in your default ssh config file when testing  ssh code!

This also uncovered another bug related to this. When assembling a path string, the ```+``` operator is not a good idea -- use ```filepath.Join()``` instead, so I'll fix that too.

MikaelSmith: The Auth field takes an array, and I believe it will try all the options. If you find a valid public key, we should still pass the agent as a fallback method, i.e. `[]ssh.AuthMethod{ssh.PublicKeys(signer), agent}`.

eboutili: Makes sense, will do. Thanks!

eboutili: Here's how I approached that in my fork (not committed yet though). Which I think achieves what you said(?) though surely with a more rookie coding approach/style.
```//NEW
		authmethod := []ssh.AuthMethod{agent}
		key, err := ioutil.ReadFile(identityfile)
		if err != nil {
			log.Fatalf("unable to read private key: %v", err)
		} else {
			signer, err := ssh.ParsePrivateKey(key)
			if err != nil {
		    fmt.Errorf("unable to parse private key: %v", err)
			}
			authmethod = []ssh.AuthMethod{ssh.PublicKeys(signer),}
		}

		sshConfig := &ssh.ClientConfig{
			User:            user,
			Auth:            authmethod,
			HostKeyCallback: hostKeyCallback,
		}
//NEW
```

eboutili: I see a but in the above. Editing...

MikaelSmith: Going to make some fixes
```
	var authmethod []ssh.AuthMethod
	if key, err := ioutil.ReadFile(identityfile); err != nil {
		activity.Record("Unable to read private key, falling back to SSH agent: %v", err)
	} else {
		if signer, err := ssh.ParsePrivateKey(key); err != nil {
			activity.Record("uUnable to parse private key, falling back to SSH agent: %v", err)
		} else {
			authmethod = append(authmethod, ssh.PublicKeys(signer))
		}
	}
	// Append agent now so that it comes last in case we find another method to try.
	authmethod = append(authmethod, agent)

	sshConfig := &ssh.ClientConfig{
		User:            user,
		Auth:            authmethod,
		HostKeyCallback: hostKeyCallback,
	}
```

MikaelSmith: Another note: you always need to explicitly return. Calling `fmt.Errorf` just creates an error object, it doesn't do anything with it. So those lines are effectively no-ops.

eboutili: Cool, thank you!

eboutili: I implemented the changes and tested. Ready when you are for another review. 
Note that I punted on using the activity.Record function. It wasn't working here, and it turned out to be harder for me to figure out than I expected. Happy to change that, but hoping it's something simple I'm just missing. Could you explain? No rush of course.

plugin/aws/ec2Instance.go
eboutili: Done.

plugin/aws/ec2Instance.go
```diff
@@ -217,7 +219,17 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
...
 
+	var identityfile string
+	if name, ok := meta["KeyName"]; ok {
+		homedir, err := os.UserHomeDir()
+		if err != nil {
+			return nil, err
+		}
+		identityfile = (filepath.Join(homedir, ".ssh", (name.(string) + ".pem")))
+	} else {
+		return nil, fmt.Errorf("No key name found for %v", inst)
```
MikaelSmith: I still think this shouldn't fail if we can't find a key. The ssh-agent may have it.

transport/ssh.go
```diff
@@ -64,9 +65,23 @@ func sshConnect(host, port, user string, strictHostKeyChecking bool) (*ssh.Clien
 			}
 		}
 
+		var authmethod []ssh.AuthMethod
+		if key, err := ioutil.ReadFile(identityfile); err != nil {
+			return nil, fmt.Errorf("Unable to read private key, falling back to SSH agent: %v", err)
+			// activity.Record(ctx, "Unable to read private key, falling back to SSH agent: %v", err)
```
MikaelSmith: The missing piece is that you need the `context.Context` (often given the variable name `ctx`) object passed to whatever method calls `sshConnect`.

transport/ssh.go
MikaelSmith: https://godoc.org/github.com/puppetlabs/wash/activity describes `activity.Record` a bit. I think what's missing is that Wash initializes a Journal ID on the contexts it passes to plugin methods.

plugin/aws/ec2Instance.go
eboutili: Ah, good catch. I think I can knock this out pretty quickly.

transport/ssh.go
eboutili: ... should be able to do this one fast too. Thank you.

transport/ssh.go
eboutili: Sounds good.

plugin/aws/ec2Instance.go
eboutili: Ready for review.

transport/ssh.go
eboutili: Ready for review

plugin/aws/ec2Instance.go
```diff
@@ -217,7 +220,20 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
 		return nil, fmt.Errorf("No public interface found for %v", inst)
 	}
 
+	var identityfile string
+	if keyname, ok := meta["KeyName"]; ok {
+		if err != nil {
```
MikaelSmith: Oh, `err` should never be nil here. We don't set it after checking on line 207.

MikaelSmith: So you can drop this if/else block.

PR #336, opened 2019-06-17
Update help to match environment when running in wash shell

### LINKED ISSUE #329: Ctrl + C'ing a command kills wash
**Describe the bug**
Killing the command wash is currently executing kicks me out of the wash shell all together.

**To Reproduce**
1. deploy a single telegraf container to k8s (I am using a deployment)
2. cd into `kubernetes/<cluster>/<namespace>/pods`
3. Run `exec ./<pod name> ls /bin/`
4. See error `./<pod name>: line 1: 2019-06-12T20:37:00Z: command not found` repeated over and over again
5. Press Ctrl + c to kill the exec
6. exec stops but so does the entire wash shell session

**Expected behavior**
Only the exec should be killed, not the entire shell.

**Environment (please complete the following information):**
 - OS: macOS 10.14.5
 - Version/Build number: wash version 0.4.0
 - Cloud target: k8s pod in GKE

**Additional context**
My local shell is zsh and I am using iTerm2.

MikaelSmith:
When running in the `wash` shell, we alias several commands so they
don't override common shell builtins. However the `help` text wasn't
updated to reflect this, which can lead to unexpected errors like trying
to run `exec` and having it overwrite the shell process rather than
invoke a command on the target.

Update help output for aliased commands when run from within the shell
to refer to only command names that are setup in the shell.

Fixes #329.

### DIFF
--- a/cmd/internal/config/core.go
+++ b/cmd/internal/config/core.go
@@ -15,12 +15,14 @@ import (
 
 // Contains all the keys for Wash's shared config
 const (
-	SocketKey = "socket"
+	SocketKey   = "socket"
+	EmbeddedKey = "embedded"
 )
 
 // Socket is the path to the Wash server's UNIX
 // socket
 var Socket string
+var Embedded bool
 
 // Init initializes the config package. It loads Wash's defaults and
 // sets up viper
@@ -48,6 +50,7 @@ func Init() error {
 
 	// Load the shared config
 	Socket = viper.GetString(SocketKey)
+	Embedded = viper.GetBool(EmbeddedKey)
 
 	return nil
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -35,9 +35,6 @@ func toRunE(main commandMain) runE {
 
 func rootCommand() *cobra.Command {
 	rootCmd := &cobra.Command{
-		// TODO: Set this to "" when we're ready to ship so that
-		// when we alias our custom commands, someone typing in
-		// e.g. `meta --help` will not see `wash meta` in the usage
 		Use:    "wash [<script>]",
 		PreRun: bindServerArgs,
 		RunE:   toRunE(rootMain),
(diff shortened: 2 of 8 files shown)


cmd/internal/config/core.go
```diff
@@ -48,6 +50,7 @@ func Init() error {
 
 	// Load the shared config
 	Socket = viper.GetString(SocketKey)
+	Embedded = viper.GetBool(EmbeddedKey)
```
ekinanp: I think `config.Embedded` will also be set if it is set in `wash.yaml` right? Is that intentional?

cmd/root.go
```diff
@@ -52,12 +49,44 @@ then starts your system shell with shortcuts configured for wash subcommands.`,
...
-	addServerArgs(rootCmd, "warn")
-	// rootCommandFlag is used in rootMain.go.
-	rootCmd.Flags().StringVarP(&rootCommandFlag, "command", "c", "", "Run the supplied string and exit")
+
+	if config.Embedded {
+		rootCmd.Use = ""
+		rootCmd.PreRun = nil
+		rootCmd.Long = ""
+		// Augment the usage template to minimize usage when set to empty.
+		rootCmd.SetUsageTemplate(`Usage:{{if (and .Runnable (ne .Use ""))}}
```
ekinanp: Was this copy-pasted from Cobra? Just curious.

cmd/root.go
```diff
@@ -52,12 +49,44 @@ then starts your system shell with shortcuts configured for wash subcommands.`,
...
+Global Flags:
+{{.InheritedFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasHelpSubCommands}}
+
+Additional help topics:{{range .Commands}}{{if .IsAdditionalHelpTopicCommand}}
+  {{rpad .CommandPath .CommandPathPadding}} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableSubCommands}}
+
+Use "{{.CommandPath}} [command] --help" for more information about a command.{{end}}
+`)
+	} else {
+		addServerArgs(rootCmd, "warn")
```
ekinanp: Why not include `wash server` when `config.Embedded` is set?

cmd/rootMain.go
```diff
@@ -23,7 +23,7 @@ func writeAlias(path, subcommand string) error {
 	if err != nil {
 		return err
 	}
-	_, err = f.WriteString("#!/bin/sh\nexec wash " + subcommand + " \"$@\"")
+	_, err = f.WriteString("#!/bin/sh\nWASH_EMBEDDED=1 exec wash " + subcommand + " \"$@\"")
```
ekinanp: What does `WASH_EMBEDDED` mean? Is it equivalent to `IN_WASH_SHELL` ?

cmd/internal/config/core.go
MikaelSmith: It's not a scenario I'm concerned about. And I don't think it will be, this is all evaluated before the config file is loaded.

cmd/root.go
MikaelSmith: Yeah, then tweaked slightly.

cmd/root.go
MikaelSmith: Including it when there's already a running wash daemon seems confusing.

cmd/rootMain.go
MikaelSmith: Yes.

cmd/internal/config/core.go
ekinanp: Ok. I asked b/c I was wondering if it would make sense to read directly from the environment variable (and also do the same for the socket). The latter may be worthwhile, b/c then we don't have to say "Don't put your socket overrides in wash.yaml" in our documentation b/c the only way it's read is from the environment.

That doesn't have to be a part of this PR though.

cmd/root.go
ekinanp: Ok. May be nice to extract it into a variable b/c it is a bit long.

cmd/root.go
ekinanp: Yeah, I guess breaking the symmetry is OK there.

cmd/rootMain.go
ekinanp: Ok. A small comment explaining what it's for could help (so someone doesn't have to read through the code and figure out themselves that we're only doing this for more user-friendly help when running the `wash shell` vs without). Could maybe go in the usage template?

MikaelSmith: I'm not sure it needs to be presented to people using Wash. I don't think they draw a connection between `wash help` and `help` within the `wash` shell.

ekinanp: Oh, the comment was for maintainers, not users.

PR #337, opened 2019-06-18
Fix find, ps, version when in a wash shell

MikaelSmith:


cmd/root.go
```diff
@@ -129,3 +129,11 @@ Additional help topics:{{range .Commands}}{{if .IsAdditionalHelpTopicCommand}}
 
 Use "{{.CommandPath}} [command] --help" for more information about a command.{{end}}
 `
+
+// Return use name and alias based on whether we're embedded in a wash shell.
+func commandNames(name string) (string, []string) {
```
ekinanp: Wonder if there's a more descriptive name that captures the "Use this method if your command conflicts with a (common) shell command/built-in" intent of this method.

MikaelSmith: Probably. generateShellAlias?

ekinanp: That works.

PR #340, opened 2019-06-19
Update core plugin docs to provide more complete explanations

MikaelSmith:
Updated to provide useful context for working on existing plugins as
well as how to start a new one.

### DIFF
--- a/website/content/docs/core_plugins/_index.md
+++ b/website/content/docs/core_plugins/_index.md
@@ -2,15 +2,29 @@
 title= "Core Plugins"
 +++
 
-## How to create a new core plugin
+## Working on core plugins
 
 The [plugin] package defines a set of interfaces that a plugin can implement to enable specific behaviors.
+- [Parent](https://godoc.org/github.com/puppetlabs/wash/plugin#Parent) can list its children and is presented as a directory in the Wash filesystem.
+- [Readable](https://godoc.org/github.com/puppetlabs/wash/plugin#Readable) can retrieve the entire contents of a file and makes it available to read via standard system calls to the filesystem.
+- [Streamable](https://godoc.org/github.com/puppetlabs/wash/plugin#Streamable) can provide a stream of updates on an entry. That may be events, log entries, or new writes to a file.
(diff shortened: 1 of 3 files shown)


website/content/docs/core_plugins/_index.md
```diff
@@ -2,15 +2,29 @@
...
+
+Each entry must implement [plugin.Entry](https://godoc.org/github.com/puppetlabs/wash/plugin#Entry), which is a sealed interface that can only be satisfied by [plugin.NewEntryBase](https://godoc.org/github.com/puppetlabs/wash/plugin#NewEntryBase).
+
+Each entry that implements the `Parent` interface must provide a schema for its children. This is commonly done by giving each child entry a function that constructs a canonical version of the entry object where only the `EntryBase` field is initialized. Instances of the entry can then build on that canonical version by giving it a name and initializing other fields. If there will only ever be one instance of the entry type - such as a named directory that's a container for a specific type of thing like EC2 instances - then use the [IsSingleton()](https://godoc.org/github.com/puppetlabs/wash/plugin#EntryBase.IsSingleton) method when constructing the schema.
 
-### Starting a plugin
+Use [activity.Record](https://godoc.org/github.com/puppetlabs/wash/activity) for all plugin-related logging. Each plugin method that Wash calls is passed a `context.Context` object that is initialized for use with `activity.Record`.
```
eboutili: The lightbulb for me came on when I read the above in combination with a tip you wrote in my PR today, so I'd suggest adding that tip here as well, namely: "Wash initializes a Journal ID on the contexts it passes to plugin methods."

eboutili: Well, I guess you're saying the same thing, so maybe no change is necessary. I think what made the difference for me was that what you wrote in my PR helped me close the loop on the relationships between Journal ID, context.Context, and ctx relative methods and the things that call them.

MikaelSmith: Added the connection to Journal ID. Does that work?

eboutili: Yup, works for me.

PR #343, opened 2019-06-20
Add metadata schemas to plugin.EntrySchema

ekinanp:


plugin/entryBase.go
```diff
@@ -215,3 +256,29 @@ func (e *EntryBase) SetTestID(id string) {
 func notRunningTests() bool {
 	return flag.Lookup("test.v") == nil
 }
+
+var aliases = func() map[reflect.Type]*jsonschema.Type {
```
ekinanp: jsonschema's a pretty good library, but there are some bugs. Since I want to get something out quick, I'm having Wash point to my fork of it containing hacky fixes for those bugs that work for our use-cases. The fork is at https://github.com/ekinanp/jsonschema. I've filed issues at https://github.com/alecthomas/jsonschema/issues for the things I had to fix in that fork.

ekinanp: I wanted to use JSON schemas because they're a standard. `wash find` doesn't need the full-blown schema -- it only needs to know whether a given path (e.g. like `.tags[?].key`) is valid, so only the `properties` key for objects or the `items` keys for arrays are necessary.

The schemaP for the meta primary will consist of constructing a "mock" JSON object and then validating that object against the provided schema, returning true if validation succeeded, false otherwise. Thus, I'll be munging the JSON schema in `wash find` to remove things like `RequiredProperties` (since I only care that the properties exist), deleting `type` entries for the primitive types, and upcasing "property" names so that the matching key stuff works. Note that my reason for removing `type` for primitive types is b/c I don't think strict type-level validation is necessary for the meta primary, especially when we allow multiple types (string, int) for e.g. the time predicate. I think just validating that what the user types in leads to an existent metadata entry is more than enough.

On a side note, I'm wondering if I should change `JSONSchema` => `MetadataSchema` and just do the `wash find`-specific munging here so that the API can return it. Although that would require me to comment "This is for `wash find` only", which indicates that it belongs in `wash find`. I guess returning the schema as-is means we could re-use it if we ever need to for other things.

plugin/docker/container.go
```diff
@@ -21,16 +21,21 @@ type container struct {
 	client *client.Client
 }
 
-func containerBase() *container {
+func containerBase(forInstance bool) *container {
```
ekinanp: Not really a fan of this pattern. Am wondering if changing `forInstance` to `forChildSchemas` would make things clearer. Then again, maybe it's worth changing these to `containerTemplate` so that I can add the `NewEntryTemplate`/`NewEntry`. Note that we still need this to return a `container` object in order for Wash to get its type ID, its child schemas, etc.

I don't want to do a massive restructuring in this PR though, so any changes to what's here will be done once the initial metadata schema support's been added to `wash find`. Main reason is to get something working.

plugin/docker/container-log.go
```diff
@@ -18,7 +18,7 @@ type containerLogFile struct {
 	client        *client.Client
 }
 
-func containerLogFileBase() *containerLogFile {
+func containerLogFileBase(forInstance bool) *containerLogFile {
```
ekinanp: There is one subtlety here. If an entry's attributes are set but `meta` is not provided, `meta` defaults to returning a hash of those attributes (to be consistent with the "attributes are a subset of the metadata" idea). The current code does not work with those entries because there isn't any attributes schema yet so I can't construct a suitable default `meta` schema. Guess that could be a TODO or something to look at next, although I don't think it will be as valuable since lots of things support `ctime`, `mtime`, etc.

plugin/entryBase.go
ekinanp: I'm also wondering if `aliases` should be plugin-specific, maybe something provided by the plugin roots? Kube's `Time` type + `resource.Quantity` types are fairly common.

MikaelSmith: We should work towards making those fixes in the upstream library. It'll be more maintainable in the long run.

plugin/entryBase.go
```diff
@@ -181,6 +195,33 @@ func (e *EntryBase) IsSingleton() *EntryBase {
 	return e
 }
 
+// SetMetaAttributeSchema sets the meta attribute's schema. obj is an empty struct
+// that will be marshalled into a JSON schema. SetMetaSchema will panic
```
MikaelSmith: `SetMetaSchema` -> `SetMetaAttributeSchema`

plugin/entryBase.go
MikaelSmith: Aliases should definitely be something plugin-specific. Why would K8's time and quantity types be used outside of the kubernetes plugin?

plugin/entryBase.go
```diff
@@ -3,7 +3,13 @@ package plugin
 import (
 	"context"
 	"flag"
+	"fmt"
+	"reflect"
 	"time"
+
+	"github.com/ekinanp/jsonschema"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
```
MikaelSmith: We should definitely avoid pulling k8s in as a dependency of the plugin module. That's going to cause real problems if we eventually split individual plugins out as separate artifacts to be more modular.

plugin/entryBase.go
ekinanp: Ok. What do we do if the upstream library is not responsive? Do we write our own? Note that this was the only "good" library I could find that generated JSON schemas from Go structs.

plugin/entryBase.go
```diff
@@ -181,6 +195,33 @@ func (e *EntryBase) IsSingleton() *EntryBase {
...
+	}
+	e.metaAttributeSchema = s
+	return e
+}
+
+// SetMetadataSchema sets e#Metadata's schema. obj is an empty struct
+// that will be marshalled into a JSON schema. SetMetadataSchema will
+// panic if obj is not a struct.
+//
+// NOTE: Only use SetMetadataSchema if you're overriding e#Metadata.
```
MikaelSmith: Can you include something about why both options exist? Probably in the commit message.

plugin/docker/container-log.go
MikaelSmith: `containerLogFileBase` is a convention, not an interface you have to conform to. If you're not going to use `forInstance` in this case, why add it?

plugin/docker/container.go
MikaelSmith: I personally think having the argument name reflect what it does, rather than the intended use case (because those can change or their may be multiple use cases) makes more sense.

MikaelSmith: I would lean towards separating the schema from instance creation be something the compiler can enforce. Having `MetadataSchema` be a method that `plugin.Entry` types have to implement, and then only calling it when interacting with `ChildSchemas`, might work.

plugin/aws/ec2Instance.go
```diff
@@ -40,14 +40,17 @@ const (
...
-func ec2InstanceBase() *ec2Instance {
+func ec2InstanceBase(forInstance bool) *ec2Instance {
 	ec2Instance := &ec2Instance{
 		EntryBase: plugin.NewEntryBase(),
 	}
-	ec2Instance.
-		SetLabel("instance").
-		SetTTLOf(plugin.ListOp, 30*time.Second).
-		DisableCachingFor(plugin.MetadataOp)
+	ec2Instance.DisableCachingFor(plugin.MetadataOp)
```
MikaelSmith: I don't think we implement `Metadata` any more, so this might be unnecessary.

plugin/entryBase.go
MikaelSmith: I guess we keep the fork if it's unresponsive. But it's worth having more than hacks, with testing, if we're going to rely on the library. And once that's done, it's easy to make an upstream PR.

ekinanp: Ok. I can file an issue to implement a proper fix for the library once this is merged.

ekinanp:
I'm working on a refactor PR to clean-up some of the EntrySchema stuff. That will make this easier to implement, so am closing this PR until that one's finished.

PR #351, opened 2019-06-24
Add metadata schemas to plugin.EntrySchema

ekinanp:
Metadata schemas will be used in `wash find` to optimize the meta
primary (and to make it less annoying to use).

plugin.Root now supports a WrappedTypes method. This is needed by the
Kubernetes plugin because the Kubernetes library creates wrappers around
basic Go types like time.Time. In the general case, WrappedTypes will be
implemented by core plugins whose SDKs create wrappers around primitive types

### DIFF
--- a/plugin/cache.go
+++ b/plugin/cache.go
@@ -117,12 +117,12 @@ func CachedOp(ctx context.Context, opName string, entry Entry, ttl time.Duration
 // DuplicateCNameErr represents a duplicate cname error, which
 // occurs when at least two children have the same cname.
 type DuplicateCNameErr struct {
-	ParentID                        string
-	FirstChildName                  string
-	FirstChildSlashReplacer         rune
-	SecondChildName                 string
-	SecondChildSlashReplacer        rune
-	CName                           string
+	ParentID                 string
+	FirstChildName           string
+	FirstChildSlashReplacer  rune
+	SecondChildName          string
+	SecondChildSlashReplacer rune
+	CName                    string
 }
 
 func (c DuplicateCNameErr) Error() string {
@@ -163,12 +163,12 @@ func CachedList(ctx context.Context, p Parent) (map[string]Entry, error) {
 
 			if duplicateEntry, ok := searchedEntries[cname]; ok {
 				return nil, DuplicateCNameErr{
-					ParentID:                        p.id(),
-					FirstChildName:                  duplicateEntry.name(),
-					FirstChildSlashReplacer:         duplicateEntry.slashReplacer(),
-					SecondChildName:                 entry.name(),
-					SecondChildSlashReplacer:        entry.slashReplacer(),
-					CName:                           cname,
+					ParentID:                 p.id(),
+					FirstChildName:           duplicateEntry.name(),
+					FirstChildSlashReplacer:  duplicateEntry.slashReplacer(),
+					SecondChildName:          entry.name(),
+					SecondChildSlashReplacer: entry.slashReplacer(),
+					CName:                    cname,
 				}
 			}
 			searchedEntries[cname] = entry
@@ -177,6 +177,8 @@ func CachedList(ctx context.Context, p Parent) (map[string]Entry, error) {
 			// where the context doesn't include the parent's ID.
 			id := strings.TrimRight(p.id(), "/") + "/" + cname
 			entry.setID(id)
+
+			passAlongWrappedTypes(p, entry)
 		}
 
 		return searchedEntries, nil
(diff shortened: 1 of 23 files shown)


ekinanp:
The changes to the individual plugins seemed small enough that I figured I might as well put everything in a single commit. I can split this out if that's preferred.

plugin/types.go
```diff
@@ -80,11 +83,27 @@ type Parent interface {
...
+// also don't count. Only promoted types ("type Time time.Time") and wrapper structs
+// ("struct Time { t time.Time }") count.
+//
+// NOTE: Without WrappedTypes, the underlying JSON schema library will treat promoted types and
+// wrapper structs as JSON objects, which would be incorrect. This, unfortunately, is a
+// limitation of Go's reflect package.
 type Root interface {
 	Parent
 	Init(map[string]interface{}) error
+	WrappedTypes() SchemaMap
```
ekinanp: Another alternative I was considering was making this optional. E.g. something like
```
type HasWrappedTypes struct {
    Root
    WrappedTypes() SchemaMap
}
```

That way, plugin authors whose SDKs don't have potentially conflicting wrapped types won't have to implement this. However, that is also an extra thing we'll need to document, which is why I like having `WrappedTypes()` in Root. Although returning nil can be redundant, it's easy enough (and done in only one place -- the root) that we may as well require plugin authors to implement it.

I'm also not sure if WrappedTypes is a good name, so open to other alternatives.

plugin/jsonSchema.go
```diff
@@ -0,0 +1,39 @@
+package plugin
+
+import (
+	"github.com/ekinanp/jsonschema"
+)
+
+// JSONSchema represents a JSON schema
+type JSONSchema = jsonschema.Schema
```
ekinanp: The type alias and the schemas here prevent plugin authors from having to learn about the `jsonschema` library. That's why this file's here (and why it is useful).

Note that Alec did get back to me on one of my jsonschema issues (https://github.com/alecthomas/jsonschema/issues/41) so may end up filing an issue to properly fix the `jsonschema` library and merge those fixes upstream once some of our other in-priority work's finished.

plugin/cache.go
```diff
@@ -177,6 +177,8 @@ func CachedList(ctx context.Context, p Parent) (map[string]Entry, error) {
 			// where the context doesn't include the parent's ID.
 			id := strings.TrimRight(p.id(), "/") + "/" + cname
 			entry.setID(id)
+
+			passAlongWrappedTypes(p, entry)
```
ekinanp: I'm going to add unit tests for the EntrySchema stuff once the metadata schema work is finished, so am holding off on updating CachedList's unit tests until that point.

Note that I did test this with curl + `/fs/schema?path=$(realpath mnt)` -- the wrapped types are working OK.

plugin/types.go
MikaelSmith: We need to document it either way. I lean towards making it optional, we've already added lots of required functions that make it more complicated to start.

plugin/entrySchema.go
```diff
@@ -191,17 +221,85 @@ func (s *EntrySchema) fillChildren(visited map[string]bool) {
...
+// the helper is because /fs/schema uses repeated calls to CachedList when
+// fetching the entry, so we need to pass-along the wrapped types when
+// searching for it. However, Parent#ChildSchemas uses empty Entry objects
+// that do not go through CachedList (by definition). Thus, the entry found
+// by /fs/schema needs to pass its wrapped types along to the children to
+// determine their metadata schemas. This is done in s.fill().
+func passAlongWrappedTypes(p Parent, child Entry) {
+	var wrappedTypes SchemaMap
+	// This check is equivalent to "_, ok := child.(Root); ok".
+	// However the latter's more expensive due to the extra
```
MikaelSmith: Have you benchmarked whether this is true?

plugin/types.go
ekinanp: Ok. Any suggestions on the name?

plugin/entrySchema.go
MikaelSmith: Looking at the assembly produced, it seems like the same amount of work.

MikaelSmith: Also seems simpler if `Root` just overrode the implementation of `wrappedTypes` to call `WrappedTypes` (if it exists).

plugin/types.go
MikaelSmith: What you proposed seems fine.

plugin/entrySchema.go
MikaelSmith: Nevermind, forgot Root was an interface. What's here now looks good.

MikaelSmith:
Supports #291.

PR #352, opened 2019-06-24
Data driven hierarchy

### LINKED ISSUE #345: Provide data-driven hierarchy for external plugins
Add an ability to provide multiple levels of hierarchy at once when it's all known ahead-of-time (so we don't have to invoke the plugin multiple times).

Add a new format for `methods` that allows you to specify static hierarchy for the `list` method and static content for the `read` method.

An example of what this might look like (in pseudo-yaml) is
```
{
  name: name,
  methods: {
    list: [
      { name: name1, methods: { read: value_str1 },
      { name: name2, methods: { read: value_str2 }
    ]
  ]
}
```

If the listed items specify non-static methods (i.e. `{"read": null}` or `["list", "exec"]`) the plugin script should be invoked on them using whatever state was provided with the item.

MikaelSmith:
Adds a new format for `methods` that allows you to specify static hierarchy for the `list` method and static content for the `read method`.

Includes documentation, and adds a table-of-contents to the external plugins docs.

Resolves #345.


website/content/docs/external_plugins/_index.md
```diff
@@ -88,6 +100,46 @@ We see from `cache_ttls` that the result of `some_entry`'s `list` method will be
 
 `init` adopts the standard error conventions described in the [Errors](#errors) section.
 
+### static methods
+
+If your plugin knows some of the hierarchy and wants to describe it now rather than relying on more calls to `list` or `read` methods, it can do so by describing that hierarchy in the `methods` property of any attribute.
```
ekinanp: I was thinking that the `methods` hash would be the default format, but now I'm starting to agree that defaulting to an array would be better b/c enumerating your hierarchy is an uncommon scenario and an array makes things easy for Bash plugin authors (especially if we start supporting multiple output formats). Plus, the explanation of a "null" result is a bit wordy, not very clear [e.g. what does "just like providing the method name in an array" mean?], doesn't explain why `stream`/`exec` should always return nil, and requires me to juggle two things at once (that methods can be a hash, and that methods can be an array). I also could not come up with any better documentation than what you have here, so that suggests to me that maybe the `methods` hash isn't the right approach now that we're defaulting to an array.

So, as an alternative, what do you think of something like
```
[
"list",
["read", <result_of_read>],
"exec",
...
]
```
Explaining this format would be easy and would not require an extra section. We would say something like

> You return an array of the entry's supported methods. If you're able to prefetch a method's result, then you can include the result as a tuple of `(<method>, <result>)`. Prefetching's a useful way to avoid unnecessary plugin script invocations, etc. etc. etc.

The `prefetch => use tuple-format` type of thinking would also stop us from having to explain why `stream`/`exec` (and other "dynamic" methods) would always return nil in the `methods` hash (because external plugin authors can figure out for themselves that it is impossible to prefetch `exec`/`stream`'s/dynamic method's results from their semantics.).

I don't think the validation for the tuple-extension would be too hard. Plus internally, we'd still keep the methods hash in externalPluginEntry because that is the sanest (and easiest) way to represent the data.

ekinanp:
One thing that's a bit tricky about this is how do we refresh the data? Guessing we could try doing what the volume package does, which is to enumerate the subtree root by re-invoking list on it.

MikaelSmith:
Good point on refreshing. I think I can do something about that.

MikaelSmith:
I haven't applied this to Volume yet. I'm exploring that, but I think it'll be a big enough change that I don't want to hold this review up.

plugin/externalPluginEntry.go
```diff
@@ -96,34 +147,55 @@ func (e *externalPluginEntry) Schema() *EntrySchema {
...
-		return nil, newStdoutDecodeErr(
-			ctx,
-			"the entries",
-			err,
-			stdout,
-			"[{\"name\":\"entry1\",\"methods\":[\"list\"]},{\"name\":\"entry2\",\"methods\":[\"list\"]}]",
-		)
+	if impl, ok := e.methods["list"]; ok && impl != nil {
+		// Entry statically implements list. Construct new entries based on that rather than invoking the script.
+		bits, err := json.Marshal(impl)
```
ekinanp: Guessing early validation of list's result didn't work out?

fuse/core.go
```diff
@@ -55,12 +54,12 @@ var uid, gid = getIDs()
 
 type fuseNode struct {
 	ftype             string
-	parent            plugin.Parent
+	parent            *dir
 	entry             plugin.Entry
```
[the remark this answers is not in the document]
MikaelSmith: I still need to store name and id. I guess I could do that directly.

plugin/externalPluginEntry.go
MikaelSmith: That'd be a lot of extra validation. I didn't explore it.

ekinanp: That's fine, was just curious.

website/content/docs/external_plugins/_index.md
```diff
@@ -141,6 +152,34 @@ Below is an example of valid `list` output:
 ]
 ```
 
+If you're able to pre-fetch a method's result as part of the `list` method, then you can include the result as a tuple of `[<method>, <result>]` in the `methods` array. Pre-fetching is a useful way to avoid unnecessary plugin script invocations. Below is an example that includes pre-fetched method results. Note that when `read` content is provided in this manner, the size of that content will be automatically populated in `attributes`.
```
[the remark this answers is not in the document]
MikaelSmith: What I had before was
```
If your plugin knows some of the hierarchy and wants to describe it now rather than relying on more calls to `list` or `read` methods, it can do so by describing that hierarchy in the `methods` property of any attribute.

Instead of providing a list of methods to `methods`, provide an object. The keys of that object are supported methods on the entry. If the key maps to `null`, then the method is supported and will invoke your script again for that method (just like providing the method name in an array). The following alternative values may be provided
- for the `list` method, an object corresponding to what would be returned by invoking the `list` method on the script
- for the `read` method, a string containing the entry's contents

Below is an example JSON object declaring a static directory that contains known files and content, but may also support streaming new updates dynamically (by invoking the `stream` method on the script).
```

Not sure what you were referring to.

fuse/core.go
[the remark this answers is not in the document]
MikaelSmith: There's several reasons that doesn't work. Using it to determine whether it's a directory, fallback in case this is a plugin root.

website/content/docs/external_plugins/_index.md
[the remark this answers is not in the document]
ekinanp: Hmm, looks like the example here corresponds to the `static directory` case so may be worth mentioning that.

Was thinking a `Prefetch Scenarios` section could be useful so we could also describe Corey's use-case with the hash, but maybe that's too early.

plugin/externalPluginEntry.go
MikaelSmith: Not really. You'd want to try it out by actually listing things, so you'd run into an error pretty quickly.

website/content/docs/external_plugins/_index.md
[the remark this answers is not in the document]
MikaelSmith: I added more explanation of the example. Not ready to add scenarios.

website/content/docs/external_plugins/_index.md
```diff
@@ -141,6 +152,36 @@ Below is an example of valid `list` output:
...
 ```
 
+If you're able to pre-fetch a method's result as part of the `list` method, then you can include the result as a tuple of `[<method>, <result>]` in the `methods` array. Pre-fetching is a useful way to avoid unnecessary plugin script invocations.
+
+Below is an example that includes pre-fetched method results for a static directory that contains known files and content, but may also support streaming new updates dynamically (by invoking the `stream` method on the script). Note that when `read` content is provided in this manner, the size of that content will be automatically populated in `attributes`.
+
+```json
+[
+  {
+    "name": "mydir",
```
ekinanp: I'd maybe add a note that's something like "Notice how `list`'s result matches what would have been returned by `plugin_script list mydir`" so the symmetry's more apparent.

plugin/entryBase.go
```diff
@@ -128,6 +133,15 @@ func (e *EntryBase) SetAttributes(attr EntryAttributes) *EntryBase {
 	return e
 }
 
+// Prefetched tags an entry as having prefetched data that is populated
```
ekinanp: How about something like:

"Prefetched marks the entry as a prefetched entry. A prefetched entry is an entry that was fetched as part of a batch operation that fetched multiple levels of hierarchy at once. Volume directories and files are good examples of prefetched entries (see the volume package for more details)."

Reason for the reword is b/c I don't think it's worth including the "use this if you want your entry's data to be updated" -- the updating/refreshing is a part of a prefetched entry's semantics.

plugin/externalPluginEntry_test.go
```diff
@@ -69,19 +69,61 @@ func (suite *ExternalPluginEntryTestSuite) TestDecodeExternalPluginEntryRequired
...
+		suite.False(entry.isPrefetched())
+
+		methods := entry.supportedMethods()
+		suite.Equal(2, len(methods))
+		suite.Contains(methods, "list")
+		suite.Contains(methods, "stream")
+	}
+}
+
+func (suite *ExternalPluginEntryTestSuite) TestDecodeExternalPluginEntryWithImpls() {
```
ekinanp: Nit: `Impls` => `MethodResults` or something similar.

fuse/core.go
```diff
@@ -111,38 +110,51 @@ func (f *fuseNode) applyAttr(a *fuse.Attr, attr *plugin.EntryAttributes) {
...
-			activity.Record(ctx, "FUSE: Attr errored %v, %v", f, err)
-			return err
-		}
-		attr = plugin.Attributes(updatedEntry)
-		// NOTE: We could set f.entry to updatedEntry, but doing so would require
-		// a separate mutex which may hinder performance. Since updating f.entry
-		// is not strictly necessary for the other FUSE operations, we choose to
-		// leave it alone.
+	// FUSE caches nodes for a long time, meaning there's a chance
+	// that f's attributes are outdated. CachedList returns the entry's
```
ekinanp: Should this comment be updated now that we're no longer directly referencing cached list?

fuse/core.go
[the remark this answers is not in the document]
ekinanp: Ah that sucks. Could we use ftype to store all of that when the fuseNode's created?

plugin/entryBase.go
ekinanp: Plus, it's also easier to tell people "If you fetched this thing via a batched operation that fetched multiple levels of hierarchy, then use this helper to mark it as prefetched." Maybe that sentence alone is good enough.

ekinanp: Kind of wondering if `Prefetch` is a good name here. By definition, prefetch means you're getting data you'll know you need soon, but that you don't know for sure you'll need. In our case, that's not true for the first level in the hierarchy since the prefetch was trigged by a list call to the parent (meaning that we know for sure we'll need the first level, instead of "we might need the first level"). That's why I'm wondering if something like `BulkFetched`/`BatchFetched` might be better? It would be consistent with how we're thinking about this stuff and how we intend for it to be used (e.g. the external plugin need arose b/c a user wanted a way to pass-in batch-fetched data to Wash).

I could also be talking nonsense here so if I am, please feel free to ignore this comment.

MikaelSmith: I think the name is fine. If someone is confused about the documentation, we'll reconsider.

fuse/core.go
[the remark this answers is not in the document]
MikaelSmith: I'll work on it after lunch.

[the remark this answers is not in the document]
MikaelSmith: Too much of the system is built on having an entry. It may not always be what we want, but often it is. I could only get rid of the entry reference if I moved to always recreating the entire history because I changed to `parent *dir`, so `dir` needs to have an entry.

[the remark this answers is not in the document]
MikaelSmith: I do have a few small tweaks to add though.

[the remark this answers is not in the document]
ekinanp: Ok, I don't have an issue with the current changes too. Thanks for digging into that.

Issue #348, opened 2019-06-21
Scrape user name from the ec2 console output


eboutili:
If User isn't defined in ```~/.ssh/config```, have ```exec``` try to connect with the user name that the instance was provisioned with.

PR #353, opened 2019-06-26
Resolve Issue #348: Scrape user name from the ec2 console output

eboutili:
Scan console output for the user name that the ec2 instance was provisioned with. if none is found, set it to a best guess, "ec2-user".
 
A change to the method cachedConsoleOutput was also needed... Makes it a function of ec2Instance instead of ec2InstanceConsoleOutput.

### DIFF
--- a/plugin/aws/ec2Instance.go
+++ b/plugin/aws/ec2Instance.go
@@ -69,7 +69,13 @@ func newEC2Instance(ctx context.Context, inst *ec2Client.Instance, session *sess
 }
 
 func (inst *ec2Instance) cachedConsoleOutput(ctx context.Context, latest bool) (consoleOutput, error) {
-	output, err := plugin.CachedOp(ctx, "ConsoleOutput", inst, 30*time.Second, func() (interface{}, error) {
+	var opname string
+	if latest {
+		opname = "ConsoleOutputLatest"
+	} else {
+		opname = "ConsoleOutput"
+	}
+	output, err := plugin.CachedOp(ctx, opname, inst, 30*time.Second, func() (interface{}, error) {
 		request := &ec2Client.GetConsoleOutputInput{
 			InstanceId: awsSDK.String(inst.id),
 		}
--- a/plugin/aws/ec2Instance.go
+++ b/plugin/aws/ec2Instance.go
@@ -132,9 +132,13 @@ func (inst *ec2Instance) List(ctx context.Context) ([]plugin.Entry, error) {
 
 	consoleOutput, err := newEC2InstanceConsoleOutput(ctx, inst, false)
 	if err != nil {
-		return nil, err
+			return nil, err
 	}
 	entries = append(entries, consoleOutput)
+
+	// output, err := consoleOutput.cachedConsoleOutput(ctx)
+	output, err := inst.cachedConsoleOutput(ctx)
+	fmt.Println(string(output.content))
 
 	if inst.hasLatestConsoleOutput {
 		if latestConsoleOutput == nil {
--- a/plugin/aws/ec2Instance.go
+++ b/plugin/aws/ec2Instance.go
@@ -136,11 +136,6 @@ func (inst *ec2Instance) List(ctx context.Context) ([]plugin.Entry, error) {
 			return nil, err
 	}
 	entries = append(entries, consoleOutput)
-
-	// output, err := consoleOutput.cachedConsoleOutput(ctx)
-	output, err := inst.cachedConsoleOutput(ctx)
-	fmt.Println(string(output.content))
-
 	if inst.hasLatestConsoleOutput {
 		if latestConsoleOutput == nil {
 			latestConsoleOutput, err = newEC2InstanceConsoleOutput(ctx, inst, true)
@@ -209,21 +204,6 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
 	if err != nil {
 		return nil, err
 	}
-
-	// TODO: scrape default user from console output.
-	// See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/connection-prereqs.html#connection-prereqs-fingerprint
-	// for some helpful defaults we could add.
-
-	var fallbackuser string
-	re := regexp.MustCompile(`\WAuthorized keys from .home.*authorized_keys for user ([^+]*)+`)
-	output, err := (inst.cachedConsoleOutput(ctx))
-	outputstream := (string(output.content))
-	match := re.FindStringSubmatch(outputstream)
-	if match != nil {
-		fallbackuser = (match[1])
-		fmt.Println(fallbackuser)
-	}
-
 	var hostname string
 	if name, ok := meta["PublicDnsName"]; ok {
 		hostname = name.(string)
@@ -232,7 +212,6 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
 	} else {
 		return nil, fmt.Errorf("No public interface found for %v", inst)
 	}
-
 	var identityfile string
 	if keyname, ok := meta["KeyName"]; ok {
 		if homedir, err := os.UserHomeDir(); err != nil {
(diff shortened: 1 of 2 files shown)


plugin/aws/ec2InstanceConsoleOutput.go
```diff
@@ -50,16 +50,16 @@ type consoleOutput struct {
 	content []byte
 }
 
-func (cl *ec2InstanceConsoleOutput) cachedConsoleOutput(ctx context.Context) (consoleOutput, error) {
-	output, err := plugin.CachedOp(ctx, "ConsoleOutput", cl, 30*time.Second, func() (interface{}, error) {
+func (inst *ec2Instance) cachedConsoleOutput(ctx context.Context) (consoleOutput, error) {
```
MikaelSmith: I'd suggest moving this to `ec2Instance.go`, since that covers functions related to that type.

plugin/aws/ec2InstanceConsoleOutput.go
```diff
@@ -50,16 +50,16 @@ type consoleOutput struct {
...
+	output, err := plugin.CachedOp(ctx, "ConsoleOutput", inst, 30*time.Second, func() (interface{}, error) {
 		request := &ec2Client.GetConsoleOutputInput{
-			InstanceId: awsSDK.String(cl.inst.id),
+			InstanceId: awsSDK.String(inst.id),
 		}
-		if cl.latest {
-			request.Latest = awsSDK.Bool(cl.latest)
+		if inst.hasLatestConsoleOutput {
```
MikaelSmith: I think this needs to be an argument you pass in. `newEC2InstanceConsoleOutput` is used to generate two different files, to cover `console.out` and `console-latest.out`. From your use in `ec2Instance` we can choose which one, but for use in `ec2InstanceConsoleOutput`'s `Open` function we need to use whatever it was created with (`cl.latest`).

plugin/aws/ec2Instance.go
```diff
@@ -225,7 +222,24 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
 			identityfile = (filepath.Join(homedir, ".ssh", (keyname.(string) + ".pem")))
 		}
 	}
-	// Use the default user for Amazon AMIs. See above for ideas on making this more general. Can be
-	// overridden in ~/.ssh/config.
-	return transport.ExecSSH(ctx, transport.Identity{Host: hostname, FallbackUser: "ec2-user", IdentityFile: identityfile}, append([]string{cmd}, args...), opts)
+	var fallbackuser string
```
MikaelSmith: I'd personally appreciate a newline to separate getting the fallbackuser from getting identityfile.

plugin/aws/ec2Instance.go
```diff
@@ -225,7 +222,24 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
...
+	// Scan console output for user name instance was provisioned with. Set to ec2-user if not found
+	re := regexp.MustCompile(`\WAuthorized keys from .home.*authorized_keys for user ([^+]*)+`)
+	output, err := (inst.cachedConsoleOutput(ctx))
+	if err != nil {
+		activity.Record(ctx, "Cannot get cached console output", err)
```
MikaelSmith: Travis CI is highlighting that these calls to `Record` both need to include formatting characters. So add `: %v` to the end of the string.

plugin/aws/ec2InstanceConsoleOutput.go
eboutili: Done

plugin/aws/ec2Instance.go
eboutili: Done

plugin/aws/ec2Instance.go
eboutili: Fixed

plugin/aws/ec2InstanceConsoleOutput.go
MikaelSmith: You need to essentially preserve the old behavior. Which means add a 2nd argument to `cachedConsoleOutput` that is `latest`. Then when you call it from `ec2Instance` pass `inst.hasLatestConsoleOutput`, when you call it from `ec2InstanceConsoleOutput` pass `cl.latest`.

eboutili: Ah, thank you. I just couldn't make that leap. 

In the meantime, I got to thinking maybe the cachedConsoloeOutput function could get at `latest` if I simply point to the struct that contains it like this:
cl := &ec2InstanceConsoleOutput{}
That's based on the assumption that value for `latest` will always be what it should be at that point. It seems to work. But then I saw you post and suspect maybe this approach isn't correct (e.g. I'm making wrong assumptions). If not though, is there an advantage to doing it one way or the other?

MikaelSmith: I think you're missing what it was doing before. https://github.com/puppetlabs/wash/blob/0.4.0/plugin/aws/ec2Instance.go#L117-L130 creates 2 versions of ec2InstanceConsoleOutput. One passes in `latest=false`, the other - if `hasLatestConsoleOutput == true` - passes in `latest=true`. So we have versions that request with and without console latest.

@ekinanp I don't actually remember why we want the `!latest` version when `hasLatestConsoleOutput == true`. Do you remember why you included it?

ekinanp: That might be me mis-interpreting the API docs. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-console.html implies that they could be two different outputs. If I'm wrong and they're the same output. then I don't think we need to include latest.

MikaelSmith:
Do you mind if I squash all the commits when I merge?

plugin/aws/ec2Instance.go
```diff
@@ -66,6 +68,38 @@ func newEC2Instance(ctx context.Context, inst *ec2Client.Instance, session *sess
 	return ec2Instance
 }
 
+func (inst *ec2Instance) cachedConsoleOutput(ctx context.Context, latest bool) (consoleOutput, error) {
+	output, err := plugin.CachedOp(ctx, "ConsoleOutput", inst, 30*time.Second, func() (interface{}, error) {
```
MikaelSmith: Oh arg, `latest` now means two calls to `CachedOp` may have different inputs. Can you use different strings when latest is true vs false? Something like `ConsoleOutput` vs `ConsoleOutputLatest`?

MikaelSmith: This is probably more complicated than it needs to be now. I can try to do some extra cleanup later.

plugin/aws/ec2InstanceConsoleOutput.go
eboutili: I went ahead and pushed the changes just now per Michael's earlier guidance. That is, in case we want to merge still keeping the two-versions interpretation. Happy to work on the task of making it one version if that's the way to go... either post-merge or just continuing to update this PR in-place.

plugin/aws/ec2Instance.go
eboutili: Sure, I think I know what you mean. No worries, I'll have a look in a bit.

eboutili: Totally fine if you squash commits btw.

eboutili: Whoops, just saw your last post. Yes, that sounds good.

MikaelSmith: Can you still push the change for using different cache prefixes: ConsoleOutput and ConsoleOutputLatest?

eboutili: Well, this line of code...

```output, err := plugin.CachedOp(ctx, "ConsoleOutput", inst, 30*time.Second, func() (interface{}, error) { ... ```

involves (I think) advanced concepts that I don't have a deep enough understanding of yet., therefore it might best for you to provide the code needed in this case. Or... is it just a matter of doing something like this?

``` 
func (inst *ec2Instance) cachedConsoleOutput(ctx context.Context, latest bool) (consoleOutput, error) {
  var opname string
  if latest {
    opname = "ConsoleOutputLatest"
  } else {
    opname = "ConsoleOutput"
  }
  output, err := plugin.CachedOp(ctx, opname, inst, 30*time.Second, func() (interface{}, error) { ... 
```

MikaelSmith: Just what you proposed is what I had in mind.

PR #357, opened 2019-06-29
Fix De'Morgan's law bug in the meta primary

ekinanp:

### DIFF
--- a/cmd/internal/find/primary/meta/predicateExpression_test.go
+++ b/cmd/internal/find/primary/meta/predicateExpression_test.go
@@ -11,7 +11,7 @@ import (
 // The predicates are tested separately in their *Predicate.go files, so they
 // will not be tested here. Instead, the tests here serve as "integration tests" for
 // the parsePredicateExpression function. They're meant to test parser errors, each of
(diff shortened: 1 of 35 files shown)


cmd/internal/find/primary/meta/predicateExpression_test.go
```diff
@@ -187,8 +187,17 @@ func (s *PredicateExpressionTestSuite) TestPredicateExpressions() {
 	s.RNTC("-true -a -empty -primary", "-primary", []interface{}{})
 }
 
+func (s *PredicateExpressionTestSuite) TestDeMorgansLaw() {
+	// Parsed as "NOT(3) AND NOT(6)"
+	s.RTC("! ( 3 -o 6 ) -primary", "-primary", float64(5))
```
MikaelSmith: I'm not sure I follow how you arrive at 5. Are these binary operations?

MikaelSmith: Oh duh. Just matching that it's not a 3 and not a 6. Was overthinking this.

Issue #370, opened 2019-07-03
Docker Volume read/stream don't work


MikaelSmith:
Docker volume operations on a file - read and stream - don't work right.

If read is cancelled in the middle of querying logs, the container used to read the file is not properly cleaned up. The ContainerRemove and ContainerKill operations return errors that the context was cancelled.

If stream is cancelled at any point, the same thing happens. So when you Ctrl-C to end your `tail -f` session, it leaves a running container behind.

Finally, `tail -f` just doesn't work. It halts waiting on the Stream() call to finish.

PR #372, opened 2019-07-03
Add wash validate command

### LINKED ISSUE #350: Provide tooling around validating plugin input/output
We document the call format and expected output for external plugins. It would help to have tools for validating plugin output and that make it easy to test different invocations of your script.

Add a `wash validate <plugin>` command that can be run standalone. It should walk the tree (if there’s a schema use it to avoid descending too far); aim for a depth-first traversal up to the point that we've exercised the entire schema. Report back any failures and how to reproduce them.

MikaelSmith:
Explores a plugin to help identify issues in its construction. Runs a breadth-first search through the hierarchy, exercising implemented methods on the entries, and reports any errors that occur. If testing an external plugin, provides the commands used to invoke it.

Resolves #350.

- [x] Use TypeID rather than criteria if available
- [x] Handle context cancellation on Ctrl-C
- [x] Use Schema to limit search depth (TypeID handles selecting representative types)


MikaelSmith:
I haven't added the part to take advantage of a schema yet if present.

MikaelSmith:
This helped identify the bugs in https://github.com/puppetlabs/wash/issues/370.

cmd/validate.go
```diff
@@ -0,0 +1,272 @@
...
+
+		if all {
+			for _, entry := range entries {
+				// Make a local copy for the lambda to capture.
+				entry := entry
+				wp.Submit(func() { processEntry(ctx, pw, wp, entry, all, errs) })
+			}
+		} else {
+			// Group children by ones that look "similar", and select one from each group to test.
+			groups := make(map[criteria][]plugin.Entry)
```
ekinanp: Do you think we should require external plugins to return a TypeID? That would eliminate the "group on criteria" stuff.

I was originally thinking a TypeID would be required if the external plugin supports entry schemas, but with `validate`, I'm starting to think we should just require it. It's also less things people need to juggle ("Oh I have to include a TypeID whenever I'm using schemas. I'm not using schema stuff so no need for me to do that." vs. "Oh, I gotta include a TypeID. Ok, let me make something up"). Although "make something up" could be an "easier said than done" thing. So dunno what's best here.

ekinanp:
It'd be nice to see a sample invocation here (if it's not too verbose).

cmd/validate.go
```diff
@@ -0,0 +1,272 @@
...
+
+func processEntry(ctx context.Context, pw progress.Writer, wp pool, e plugin.Entry, all bool, errs chan<- error) {
+	defer wp.Done()
+	name := plugin.ID(e)
+	crit := newCriteria(e)
+	tracker := progress.Tracker{Message: fmt.Sprintf("Testing %s %s", crit, name), Total: 4}
+	pw.AppendTracker(&tracker)
+
+	if plugin.ListAction().IsSupportedOn(e) {
+		entries, err := plugin.CachedList(ctx, e.(plugin.Parent))
```
ekinanp: Yeah, I guess starting a whole server up would be overkill. At the same time, breaking consistency with the other commands + seeing API-specific stuff like ErrorObj move over to plugin is also kind of weird. Not using a server also means we'll need to handle context-cancellation related cleanup (like on Ctrl-C). I don't know how hard the latter would be though.

How horrible would a "light-weight" server (just the API instead of FUSE) be?

cmd/validate.go
```diff
@@ -0,0 +1,272 @@
...
+	crit.list = plugin.ListAction().IsSupportedOn(entry)
+	crit.read = plugin.ReadAction().IsSupportedOn(entry)
+	crit.stream = plugin.StreamAction().IsSupportedOn(entry)
+	crit.exec = plugin.ExecAction().IsSupportedOn(entry)
+	return crit
+}
+
+func (c criteria) String() string {
+	s := []byte("----")
+	if c.list {
```
ekinanp: Think a legend in the help text might be useful.

plugin/externalPluginScript.go
```diff
@@ -12,10 +12,31 @@ import (
 // externalPluginScript represents an external plugin's script
 type externalPluginScript interface {
 	Path() string
-	InvokeAndWait(ctx context.Context, method string, entry *externalPluginEntry, args ...string) ([]byte, error)
+	InvokeAndWait(ctx context.Context, method string, entry *externalPluginEntry, args ...string) ([]byte, invokeContext, error)
 	NewInvocation(ctx context.Context, method string, entry *externalPluginEntry, args ...string) *internal.Command
 }
 
+type invokeContext struct {
```
ekinanp: I think this should be `invocation` and command/stdout/stderr should be their raw types (so `internal.Command` and `[]byte` for the streams). Then the type signatures for the `Invoke*` methods would change to:
```
([]byte, invocation, error)
invocation
```
which reads a bit better.

ekinanp: I guess the raw types isn't too important, especially for the streams. For the command object, I was thinking that munging it is easy enough we may as well return it as-is so that it's easier to access its other methods if we ever need to do so in the future.

MikaelSmith:
For some very simple examples
```
$ wash validate -a demo
Testing l--- /demo ... done! [3 in 136ms]
Testing -r-- /demo/a ... done! [3 in 0s]
Testing -r-- /demo/b ... done! [3 in 0s]
Looks good!
$ wash validate -a ~/static.sh
Testing l--- /static ... done! [3 in 3ms]
Testing l--- /static/dir ... done! [3 in 3ms]
Testing -r-- /static/dir/a ... done! [3 in 0s]
Testing -r-- /static/dir/b ... done! [3 in 0s]
Looks good!
```

cmd/validate.go
```diff
@@ -0,0 +1,272 @@
...
+		return exitCode{1}
+	}
+
+	cmdutil.Println("Looks good!")
+	return exitCode{0}
+}
+
+type pool struct {
+	wp *workerpool.WorkerPool
+	wg *sync.WaitGroup
```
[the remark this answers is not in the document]
MikaelSmith: Comment at line 117 explains it.
> We use a worker pool to limit work-in-progress, and a wait group to know when all queued work is complete.

cmd/validate.go
MikaelSmith: TypeID would be useful, but not sure we should require it.

cmd/validate.go
MikaelSmith: "API-specific" stuff like ErrorObj was moved over so I had a way to return structured information from externalPluginEntry, not anything related to using a server. I also don't see `validate` as something you should use from within the wash shell. I started with an API server, but it seemed like overkill.

I'll add the context cancellation.

MikaelSmith:
TODO:
[ ] Use TypeID rather than criteria if available
[ ] Handle context cancellation on Ctrl-C
[ ] Use Schema to limit search depth (TypeID handles selecting representative types)

MikaelSmith:
Addressed everything except the items in TODO in the description.

cmd/validate.go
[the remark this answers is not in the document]
ekinanp: Oh ok. I was asking because the wording in [StopWait's docs](https://godoc.org/github.com/gammazero/workerpool#WorkerPool.StopWait) suggests that  `StopWait` already does that, but I'm not at all familiar with the library.

cmd/validate.go
ekinanp: Ok, that sounds good.

cmd/validate.go
ekinanp: Ok, that's fair.

plugin/types.go
```diff
@@ -191,3 +192,24 @@ type Readable interface {
 	Entry
 	Open(context.Context) (SizedReader, error)
 }
+
+// ErrorFields represents the fields of an ErrorObj
+type ErrorFields = map[string]interface{}
+
+// ErrorObj represents a structured error object. If a plugin returns this type, it will be
+// passed back as part of an API response.
+type ErrorObj struct {
```
ekinanp: Would a `NewErrorObj(Kind, Msg, Fields)` helper make sense? I'm asking b/c that would centralize the `puppetlabs.wash` prefix in all of our Kind fields.

cmd/validate.go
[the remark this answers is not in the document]
MikaelSmith: I need to allow new tasks to be submitted. I don't know how many tasks there will be, so don't know when to stop without having something that keeps track of all the potential tasks in the system.

plugin/types.go
MikaelSmith: I'm not sure. The reason kind are constants are so you can match against them. We'd have to have to use a similar builder on the command side, which seems more complicated.

cmd/validate.go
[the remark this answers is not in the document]
ekinanp: Ah ok. So the waitGroup's there b/c the queue's dynamic, while `StopWait()` assumes a static queue, right? If so, a comment along those lines would be appreciated

cmd/validate.go
```diff
@@ -0,0 +1,275 @@
...
+		for err := range errs {
+			erred++
+			cmdutil.ErrPrintf("%v\n", err)
+		}
+	}()
+
+	// We use a worker pool to limit work-in-progress, and a wait group to know when all queued work
+	// is complete. Put the plugin on the worker pool. Listing the registry should never error.
+	wp := newPool(parallel)
+	entries, err := plugin.CachedList(ctx, registry)
```
ekinanp: A comment explaining why we're listing the registry instead of passing-in the root to processEntry would be nice. Guessing it's so the ID/other cached-list specific stuff is set?

plugin/types.go
ekinanp: Yeah, it's probably not that big of a deal anyways.

cmd/validate.go
MikaelSmith: Yes. I'll add that.

.travis.yml
```diff
@@ -12,6 +12,13 @@ env:
 jobs:
   include:
   - name: Test Go stable
+  - name: Test Docker plugin
+    before_script:
+    - go install
+    - docker pull busybox
+    - docker-compose -f examples/swarm/docker-compose.yml up -d
+    script:
+    - wash validate docker
```
MikaelSmith: I have seen a transient with this. Will dig into it a bit.

MikaelSmith: Ok, if it hits a symlink (like `/var/run` in the web server from this example) or a write-only file (like `/proc/sysrq-trigger`) it treats it as a file and tries to read/stream from it. So need to
- [ ] update write-only files to not support read/stream
- [ ] fix whatever's wrong with symlinks, we're likely ignoring part of the stat results or need to pass an argument to stat

cmd/validate.go
```diff
@@ -100,13 +103,21 @@ func validateMain(cmd *cobra.Command, args []string) exitCode {
...
 	pw.Style().Colors = progress.StyleColorsExample
 	go pw.Render()
 
+	// On Ctrl-C cancel the context to ensure plugin calls have a chance to cleanup.
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		<-sigCh
+		cancel()
```
ekinanp: Do we need to wait here to give stuff time to cleanup?

cmd/validate.go
```diff
@@ -249,8 +255,12 @@ func processEntry(ctx context.Context, pw progress.Writer, wp cmdutil.Pool, e pl
 			// Group children by ones that look "similar", and select one from each group to test.
 			groups := make(map[criteria][]plugin.Entry)
 			for _, entry := range entries {
-				crit := newCriteria(entry)
-				groups[crit] = append(groups[crit], entry)
+				ccrit := newCriteria(entry)
+				// If we have a TypeID, only explore children if they are different from the parent.
+				// This prevents simple recursion like volume directories containing more dirs.
+				if ccrit.typeID != "" && ccrit != crit {
```
ekinanp: Yes, I think this is a good heuristic. Only thing I'd add is to maybe print the TypeID in the console? That way, we can see what kind of thing is being tested.

volume/fs.go
```diff
@@ -44,7 +45,15 @@ func (d *FS) List(ctx context.Context) ([]plugin.Entry, error) {
 	return List(ctx, d)
 }
 
-var errNonZero = fmt.Errorf("Exec exited non-zero")
+type nonZeroError struct {
```
ekinanp: Hmm, why make a structured error object here, but use `plugin.ErrorObject` for the external plugin invoke error?

.travis.yml
MikaelSmith: I think I'm going to pull out the .travis.yml changes and include them when I fix #376.

volume/fs.go
MikaelSmith: This was a simpler case, I didn't think I'd need more formatting. The error kind is different too. It could make sense to use ErrorObject, but I don't have a case for doing something non-local.

cmd/validate.go
ekinanp: Oh also, do you think it's worth mentioning this heuristic in the help text (since it is not strictly correct?) I lean towards "No until a user files an issue about it."

MikaelSmith: Same.

MikaelSmith: I'll work TypeID into the output if present.

cmd/validate.go
MikaelSmith: Nope, we capture the signal. Execution halts because the context is cancelled, and we exit via normal control flow. So in most cases it gets
```
...^C
Testing -rs- /docker/containers/swarm_web_1/fs/proc/cpuinfo ... done! [3 in 364ms]
Testing l--- /docker/volumes/swarm_redis ... done! [3 in 1.177s]
25.00% [####..............] [1 in 1.866158s] ... Testing -rs- /docker/volumes/swarm_redis/appendonly.aof
Error validating Stream on /docker/volumes/swarm_redis/appendonly.aof, shutting down after Ctrl-C: context canceled
50.00% [#########.........] [2 in 1.966337s] ... Testing -rs- /docker/volumes/swarm_redis/appendonly.aof
Found 1 errors.
```

ekinanp: Ah cool, that makes sense.

volume/fs.go
MikaelSmith: Eventually I think I'd like errors to be more complete, so we can do stuff like highlight what part of JSON output produced an error. But we can tackle that separately.

MikaelSmith: I removed the ErrorObj changes. Only difference in output is that the URL help text is at the bottom rather than between error message and detailed output.

cmd/validate.go
MikaelSmith: Done.

PR #373, opened 2019-07-08
 Don't report outer expression operators as parse errors

### LINKED ISSUE #361: Meta primary expression parser reports outer expression operators as parse errors
Running
```
➜  mnt158800440 find \! \( -meta '.tags[?]' .key -exists -a -meta '.foo' -exists \)
find: -meta: ): no beginning '('
```

This should parse successfully, with the last `)` interpreted as being  part of the outer expression. However here, `)` is interpreted as being part of the inner expression

ekinanp:
Previously, something like "( -m -exists )" would return a parse error
because the ")" did not begin with a leading "(". This was because the
meta primary expression parser thought that ")" was part of its
expression instead of the outer expression.

This commit introduces the IncompleteOperatorError that the inner
expression parser can return. This error will pop back up to the outer
expression parser. The error's semantics are similar to the
UnknownTokenError's semantics.

Resolves #361


cmd/internal/find/parser/expression/parser.go
```diff
@@ -141,10 +194,11 @@ func (parser *parser) Parse(tokens []string) (predicate.Predicate, []string, err
 	// := operator's scoping rules. tks is used to avoid accidentally
 	// overwriting tokens.
 	//
-	// POST-LOOP INVARIANT: err == nil or err is an UnknownTokenError
+	// POST-LOOP INVARIANT: err == nil or err is an UnknownTokenError/IncompleteOperatorError
 	var p predicate.Predicate
 	var tks []string
 	var err error
+Loop:
```
MikaelSmith: Could this be a function that we return from instead? There's a lot going on in this function already.

ekinanp: Updated the PR. I wasn't sure if it added much clarity at first, but looking at it now, I think that it does.

PR #377, opened 2019-07-09
 Separate EntrySchema client code from plugin author code

### LINKED ISSUE #363: Fix /fs/schema's response to be more idiomatic w/ graph serialization
Right now, /fs/schema's response collapses cycles into a leaf node (to make it easier for stree to represent it). This representation, however, is not very idiomatic with graph serialization (which usually uses a hash of <vertex> => <edges/dependent vertices>). It is also difficult to work with and understand, especially when it comes to filling in the children + metadata schemas.

We should find a better representation to return for /fs/schema.

ekinanp:
This PR also resolves #363


ekinanp:
1400 of these lines are due to the AWS test fixture now including the metadata schema. Otherwise the changes here are about +400 lines, -600 lines.

plugin/entrySchema.go
```diff
@@ -67,25 +66,24 @@ type EntrySchema struct {
...
 	metaAttributeSchemaObj interface{}
 	metadataSchemaObj      interface{}
-	partialSchema          bool
 }
 
 // NewEntrySchema returns a new EntrySchema object with the specified label.
 //
 // NOTE: If your entry's a singleton, then the label should match the entry's
 // name, i.e. the name that's passed into plugin.NewEntry.
 func NewEntrySchema(e Entry, label string) *EntrySchema {
```
ekinanp: Note that the changes here mean we no longer need to return `EntrySchema` pointers. I chose not to do so here b/c it makes the builder pattern slightly more complicated. Specifically, plugin authors would need to do something like
```
s := plugin.NewEntrySchema(...)
return s.SetLabel(...).IsSingleton()
``` 
and if they ever want to re-use their schema after building it, then they'd need to remember to set the schema to the returned value. So something like
```
s := NewEntrySchema(e, "bar")
s.SetLabel("foo")
print(s.Label)
```
would return "bar". However, doing `s = s.SetLabel("foo")` would return "foo".

So if possible, I'd still like to return pointers. Pointers or no pointers, I think we'd still need to check TypeIDs + ChildSchemas(), esp. since the latter could still return nil (i.e. some validation will always be needed). So may as well return pointers if it's not too much of a performance hit.

api/types/entrySchema.go
```diff
@@ -0,0 +1,181 @@
...
+type EntrySchema struct {
+	plugin.EntrySchema
+	children map[string]*EntrySchema
+	// graph is an ordered map of <TypeID> => <EntrySchema>. We store it to make
+	// MarshalJSON's implementation easier.
+	//
+	// NOTE: The reason we don't synchronize children with graph is b/c entry
+	// schemas are immutable. Clients that want to mess with the state (e.g. like
+	// `wash find`) can do so at their own peril.
+	graph *linkedhashmap.Map
```
ekinanp: linkedhashmap marshals/unmarshals/iterates in insertion order, which is why I'm using it here.

api/types/entrySchema.go
```diff
@@ -0,0 +1,181 @@
...
+	if s.graph != nil {
+		return s.graph.ToJSON()
+	}
+	return json.Marshal(s.EntrySchema)
+}
+
+// UnmarshalJSON unmarshals the entry's schema from JSON.
+func (s *EntrySchema) UnmarshalJSON(data []byte) error {
+	rawGraph := linkedhashmap.New()
+	if err := rawGraph.FromJSON(data); err != nil {
```
MikaelSmith: Won't this error if `s.graph` was nil when you marshalled it?

api/types/entrySchema.go
```diff
@@ -0,0 +1,181 @@
...
+		return err
+	}
+	if rawGraph.Size() <= 0 {
+		msg := fmt.Sprintf("s.UnmarshalJSON: expected a non-empty JSON object but got %v instead", string(data))
```
MikaelSmith: I'm not sure why this is panic-worthy. Wouldn't normal error handling make sense on the client?

api/types/entrySchema.go
```diff
@@ -0,0 +1,181 @@
...
+		panic(msg)
+	}
+	s.graph = linkedhashmap.New()
+	s.children = make(map[string]*EntrySchema)
```
MikaelSmith: This isn't filled in. Should it be?

api/types/entrySchema.go
ekinanp: Not sure what you mean. If you're talking about data being the "null" string, then that's taken care of by making `EntrySchema` a pointer. This UnmarshalJSON's only called when data isn't "null".

api/types/entrySchema.go
MikaelSmith: If they need to be ordered, why not serialize as an array of objects? We can reconstruct the map from that easily .enough.

api/types/entrySchema.go
MikaelSmith: MarshalJSON has if s.graph != nil marshal the graph, else marshal s.EntrySchema. But UnmarshalJSON seems to assume that it marshaled a linkedhashmap no matter what. Is that because plugin.EntrySchema uses the same marshaling? If so why do we have two different types to store it in?

api/types/entrySchema.go
ekinanp: I don't like the array because it doesn't structurally enforce the "TypeIDs are unique" requirement of entry schemas, which is fundamental. It's also true that maps don't have to enforce insertion order either, but it is also not as uncommon to consider an insertion order when you're reasoning about map keys vs. considering uniqueness when you're reasoning about arrays. For example, I can't help but think about insertion order when I look at JSON objects, while when I think about array uniqueness, I tend to think about things as sets rather than arrays.

Also, other languages have very easy-to-use ordered map structures. For example, Ruby hashes already let you iterate in insertion order. Python has OrderedDict which is also very easy to use. Even Java has the LinkedHashMap class. It's just Go that sucks (mostly due to its lack of generics support), so if it is only one language that happens to make enforcing insertion order suck, and that level of suck is not too bad (which I don't think it is here), then I would prefer returning a response that's more consistent with how we think about schemas. In this case, that response is a JSON object with the extra caveat that the first key must be the invoked entry's schema.

api/types/entrySchema.go
ekinanp: Yeah, plugin.EntrySchema [creates a linkedhashmap and marshals that](https://github.com/ekinanp/wash/blob/schema-cleanup/plugin/entrySchema.go#L99-L105). 

Reason for the different types is b/c plugin.EntrySchema's meant to be used by plugin authors for constructing entry schemas. You cannot do more complicated schema operations like graph-traversal or schema-pruning (those would require more documentation in godocs for people to sift through). All the more complicated schema-stuff is handled in the apitypes.EntrySchema class (I didn't add any graph-traversal functions yet because I wasn't sure how to abstract out some of the wash find/stree traversal since those traversals required knowledge of the children's results and passing in multiple callbacks to a generic function seemed complicated).

api/types/entrySchema.go
ekinanp: Oh yeah, I guess I can return an error here. I'm too used to panics being a good way to enforce the type's invariants, but that can be a little too strict. Will update.

api/types/entrySchema.go
ekinanp: That's filled in at line 80. Reason I created the map here was to avoid a nil pointer exception since I'm creating it for the other cases at line 66. But I think I can clean that up a bit. Will update.

api/types/entrySchema.go
ekinanp: I think I might remove `SetLabel` per https://github.com/puppetlabs/wash/pull/377/commits/670d53ae10f56f541c383b9977f02d902d2ee288 since that's basically saying SetLabel doesn't properly work without some more code.

Issue #378, opened 2019-07-09
Wash does not correctly represent symlinks on a volume


MikaelSmith:
The volume module provides utilities for listing. However, for symlinks it represents them as files (even if it's a symlink to a directory). In the fix for #376, they'll be represented as the actual target rather than a symlink to the target.

We should correctly represent these as symlinks to other paths on the system.

PR #379, opened 2019-07-09
Fix file representation in volumes

### LINKED ISSUE #376: Some files not represented correctly with Volume module
Symlinks (like `/var/run` in the web server from the docker example) or write-only files (like `/proc/sysrq-trigger`) are treated as files. They claim to be readable/streamable, but trying to do so results in an error. In the case of the symlink, they also can't be listed (but we should be able to list a symlink to a directory).

Also some file types were not correctly represented. For example, character device would not be reflected in `ls` output.

MikaelSmith:
Fix symlinks so we get something usable. For now that means following symlinks for find/stat. Filed #378 to add representation of symlinks.

Fix mode serialization so we get the correct mode from the API. What we get from serializing an Entry now differs from what we get when requesting attributes/metadata; each uses a format that's appropriate for the use case (attributes/metadata gets a stringified representation similar to what `ls` outputs, Entries use the raw os.FileMode data). Find/stat now use a Tty when Wash does to better reflect the output a user would get when in a shell on the system. Mode is now included in `list` output.

Fix a few issues from `wash validate`:
- switch from TypeID to Label so we display something more human-readable
- wait for the progress writer to finish rendering before exiting
- skip trying to read things that we shouldn't, such as those without read permission or that would block or return infinite data

Adds Travis CI testing that exercises `wash validate` by using it to validate the Docker plugin using our Docker Swarm example.

Fixes #376.

### DIFF
--- a/plugin/entryAttributes.go
+++ b/plugin/entryAttributes.go
@@ -203,7 +203,9 @@ func (a *EntryAttributes) ToMap(includeMeta bool) map[string]interface{} {
 		mp["ctime"] = a.Ctime()
 	}
 	if a.HasMode() {
-		mp["mode"] = a.Mode()
+		// The mode string representation is the only portable representation. FileMode uses its own
+		// definitions for type bits, not those in http://man7.org/linux/man-pages/man7/inode.7.html.
+		mp["mode"] = a.Mode().String()
 	}
 	if a.HasSize() {
 		mp["size"] = a.Size()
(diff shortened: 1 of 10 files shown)


cmd/list.go
```diff
@@ -61,7 +62,9 @@ func formatListEntries(ls []apitypes.Entry) string {
 			name += "/"
 		}
 
-		table[i] = []string{name, mtimeStr, verbs}
+		mode := entry.Attributes.Mode().String()
```
ekinanp: Hmm, why print mode? It isn't a very common attribute, at least not nearly as common as mtime.

If we do decide to print the mode, then I think this should mark nonexistent modes as `unknown`. That way, we don't imply to users that mode-less things like containers/VMs have a mode.

plugin/entryAttributes.go
```diff
@@ -251,11 +258,13 @@ func (a *EntryAttributes) UnmarshalJSON(data []byte) error {
 		a.SetCtime(t)
 	}
 	if mode, ok := mp["mode"]; ok {
-		m, err := munge.ToFileMode(mode)
-		if err != nil {
-			return attrMungeError("mode", err)
+		// Without a type hint, all numbers are given as floats. This represents the uint32 type
```
ekinanp: Is this basically saying "Even though os.FileModes are uint32 types, json.Unmarshal unmarshals them as float64. That's ok, because float64 has sufficient precision to represent all uint32 types." ? If so, the current comment could be clearer because I had to think a little bit about what you meant here.

cmd/list.go
MikaelSmith: It was useful when I wasn't getting reliable data out. I'll pull it back out.

MikaelSmith:
Updated.

PR #380, opened 2019-07-10
Make setting size optional

### LINKED ISSUE #355: Size attribute must be returned when listing entries in an external script
If your external script `list` handler doesn't return a size, it defaults to `0`. The symptom of that is the `ls -l` output showing the zero-byte file. When you later try to `read` the file (eg, by using `cat`) then your script is invoked and returns content and the proper exit code, but the shell doesn't display anything on stdout.

The first problem is that this doesn't appear to be documented.

The second problem is that calculating an accurate size is often expensive, and it would be nice to not need it. (I can see that breaking a lot of assumptions in other tooling, so maybe that suggestion is not a good one.)

```
[~]$ wash --version
wash version 0.4.0
```

MikaelSmith:
Previously an entry that didn't set its size would default to 0. Many common CLI tools will ignore an empty file rather than try to open it. Update to use a default of 4k, and when cached content is available use that to give a more accurate size. This allows plugin writers to only set the size when they actually know it without downloading the content.

Resolves #355.

### DIFF
--- a/datastore/cache.go
+++ b/datastore/cache.go
@@ -17,6 +17,7 @@ import (
 // Cache is an interface for a cache.
 type Cache interface {
 	GetOrUpdate(category, key string, ttl time.Duration, resetTTLOnHit bool, generateValue func() (interface{}, error)) (interface{}, error)
+	Get(category, key string) (interface{}, error)
 	Flush()
 	Delete(matcher *regexp.Regexp) []string
 }
@@ -71,6 +72,25 @@ func (cache *MemCache) Limit(n int) *MemCache {
 	return cache
 }
 
+func formKey(category, key string) string {
+	return category + "::" + key
+}
+
+// Get retrieves the value stored at the given key. If not cached, returns (nil, nil).
+// Even if a nil value is cached, that's unlikely to be a useful value so we don't see
+// a reason to differentiate between absent and nil.
+func (cache *MemCache) Get(category, key string) (interface{}, error) {
+	key = formKey(category, key)
+	value, found := cache.instance.Get(key)
+	if found {
+		if err, ok := value.(error); ok {
+			return nil, err
+		}
+		return value, nil
+	}
+	return nil, nil
+}
+
 // GetOrUpdate attempts to retrieve the value stored at the given key.
 // If the value does not exist, then it generates the value using
 // the generateValue function and stores it with the specified ttl.
@@ -85,7 +105,7 @@ func (cache *MemCache) GetOrUpdate(category, key string, ttl time.Duration, rese
 
 	// From here on key is a composition of category and key so we can maintain
 	// a single cache.
-	key = category + "::" + key
+	key = formKey(category, key)
 	value, found := cache.instance.Get(key)
 	if found {
 		log.Tracef("Cache hit on %v", key)
--- a/website/content/docs/_index.md
+++ b/website/content/docs/_index.md
@@ -161,6 +161,8 @@ Wash entries can support the following actions:
 * `exec` - lets you execute a command against an entry
   - _e.g. run a shell command inside a container, or on an EC2 vm, or on a routerOS device, etc._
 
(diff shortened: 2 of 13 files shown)


website/content/docs/external_plugins/_index.md
```diff
@@ -71,7 +71,7 @@ You can include additional (optional) keys in the printed JSON object. These key
...
 * `cache_ttls`. This specifies how many seconds each method's result should be cached (`ttl` is short for time to live). Currently, Wash caches the result of `list`, `read`, and `metadata`.
-* `attributes`. This represents the entry's attributes (see the [`Attributes/Metadata`](/wash/docs#attributes-metadata) section). Time attributes are specified in Unix seconds. Octal modes must be prefixed with the `0` delimiter (e.g. like `0777`). Hexadecimal modes must be prefixed with the `0x` delimiter (e.g. like `0xabcd`).
+* `attributes`. This represents the entry's attributes (see the [`Attributes/Metadata`](/wash/docs#attributes-metadata) section). Time attributes are specified in Unix seconds. Octal modes must be prefixed with the `0` delimiter (e.g. like `0777`). Hexadecimal modes must be prefixed with the `0x` delimiter (e.g. like `0xabcd`). For entries that can be `read`, provide the size if you know it; otherwise Wash will provide a functional default and update the size when the entry has been `read`.
```
ekinanp: Since there's some core plugin stuff that also doesn't provide a size due to download being expensive, I think we should move this bit to the general "plugin concepts" doc. I'd also add a note that the caveat with not providing a size is that you cannot filter those entries on the size attribute using `wash find`.

datastore/cache.go
```diff
@@ -71,6 +73,23 @@ func (cache *MemCache) Limit(n int) *MemCache {
...
+func (cache *MemCache) Get(category, key string) (interface{}, error) {
+	key = formKey(category, key)
+	value, found := cache.instance.Get(key)
+	if found {
+		if err, ok := value.(error); ok {
+			return nil, err
+		}
+		return value, nil
+	}
+	return nil, KeyNotFound{key: key}
```
ekinanp: Nit: Maybe a comment along the lines of "We return an error here since nil could also be a cached value" would be useful? I had to spend a second or two thinking about why this was here.

website/content/docs/external_plugins/_index.md
ekinanp: Sorry, that you cannot always filter those entries on the size attribute ("not always" being when you've never read it or when a previous read expired).

datastore/cache.go
MikaelSmith: Hmm, should we not do the same coersion of cached errors? We could return them as values and let the caller deal with them, but I think that would make this particular use more complex.

ekinanp: Ah yeah we did cache errors. Ignore me then. I think what's here is OK.

MikaelSmith: I think I simplified it a bit, give it another look.

ekinanp: Yeah, I think what you have is OK. If people want to know whether a key exists in the cache, we could always add a `Exists` helper to the interface.

It sucks that we can't do something like `(value, err, ok)` due to Go-linting, but I guess our cache also caches errors so it's a bit different from a typical map (and that type-signature also reads weird).

website/content/docs/_index.md
```diff
@@ -161,6 +161,8 @@ Wash entries can support the following actions:
 * `exec` - lets you execute a command against an entry
   - _e.g. run a shell command inside a container, or on an EC2 vm, or on a routerOS device, etc._
 
+For entries that can be `read`, provide the size if you know it; otherwise Wash will provide a functional default and update the size when the entry has been `read`.
```
ekinanp: Nothing about `wash find` here?

MikaelSmith: Forgot.

PR #385, opened 2019-07-11
 Support external plugin schemas

### LINKED ISSUE #364: Support external plugin entry schemas

ekinanp:
Resolves #364

### DIFF
--- a/plugin/externalPluginEntry.go
+++ b/plugin/externalPluginEntry.go
@@ -7,9 +7,15 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"sort"
 	"strings"
 	"time"
 
+	log "github.com/sirupsen/logrus"
+
+	"github.com/getlantern/deepcopy"
+
+	"github.com/emirpasic/gods/maps/linkedhashmap"
 	"github.com/puppetlabs/wash/activity"
 	"github.com/puppetlabs/wash/plugin/internal"
 )
@@ -26,6 +32,7 @@ type decodedCacheTTLs struct {
 
 // decodedExternalPluginEntry describes a decoded serialized entry.
 type decodedExternalPluginEntry struct {
+	TypeID        string           `json:"type_id"`
 	Name          string           `json:"name"`
 	Methods       []interface{}    `json:"methods"`
 	SlashReplacer string           `json:"slash_replacer"`
@@ -58,17 +65,46 @@ func mungeToMethods(input []interface{}) (map[string]interface{}, error) {
 	return methods, nil
 }
 
-func (e decodedExternalPluginEntry) toExternalPluginEntry() (*externalPluginEntry, error) {
-	if e.Name == "" {
+func (e decodedExternalPluginEntry) toExternalPluginEntry(schemaKnown bool, isRoot bool) (*externalPluginEntry, error) {
+	if len(e.Name) <= 0 {
 		return nil, fmt.Errorf("the entry name must be provided")
 	}
-	if e.Methods == nil {
+	if len(e.Methods) <= 0 {
 		return nil, fmt.Errorf("the entry's methods must be provided")
+	}
+	if len(e.TypeID) <= 0 {
+		return nil, fmt.Errorf("the entry's type ID must be provided")
 	}
 
 	methods, err := mungeToMethods(e.Methods)
 	if err != nil {
 		return nil, err
+	}
+
+	// INVARIANT: If root implements schema, then schemaKnown == true (and vice versa).
+	// Idea here is that entry schemas also include their descendant's schema. So if the
+	// root implements schema, then the root's schema will include every entry's schema.
+	// Thus, it is reasonable for us to expect (and require) every entry to implement schema
+	// if the root implements it. It is also reasonable for us to expect (and require)
+	// every entry to _not_ implement schema if the root does not implement it.
+	if result, ok := methods["schema"]; ok {
+		if !schemaKnown && !isRoot {
+			return nil, fmt.Errorf("entry %v (%v) implements schema, but the plugin root doesn't", e.Name, e.TypeID)
+		}
+		// schemaKnown || isRoot
+		if !isRoot && result != nil {
+			return nil, fmt.Errorf(
+				"entry %v (%v) prefetched its schema. Only plugin roots support schema prefetching",
+				e.Name,
+				e.TypeID,
+			)
+		}
+		// schemaKnown || (isRoot || result == nil). In either case, the schema's
+		// known.
+		schemaKnown = true
+	} else if schemaKnown {
+		// ok == false here
+		return nil, fmt.Errorf("entry %v (%v) must implement schema", e.Name, e.TypeID)
 	}
 
 	// If read content is static, it's likely it's not coming from a source that separately provides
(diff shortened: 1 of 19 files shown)


ekinanp:
A lot of the changes here are test fixtures.

Here's an example of external plugin schemas in action:
```
➜  mnt249947213 stree
WARN /schema (root): the schema's supported methods (list, schema) don't match the instance's supported methods (schema, list)
.
├── docker
│   ├── containers
│   │   └── [container]
│   │       ├── log
│   │       ├── metadata.json
│   │       └── [fs]
│   │           ├── [file]
│   │           └── [dir]
│   │               ├── [dir]
│   │               └── [file]
│   └── volumes
│       └── [volume]
│           ├── [dir]
│           │   ├── [dir]
│           │   └── [file]
│           └── [file]
└── schema
    └── NodeA
        └── [NodeB]
```

```
➜  mnt136691213 stree schema/NodeA
schema/NodeA
└── [NodeB]
```

There's a small bug I'll need to fix (the warning message in the first sample shouldn't have been printed), but other than that think this is ready for a review.

plugin/externalPluginScript.go
```diff
@@ -62,9 +62,9 @@ func (s externalPluginScriptImpl) InvokeAndWait(
 		return inv, newInvokeError(err.Error(), inv)
 	}
 
-	activity.Record(ctx, "stdout: %v", inv.stdout)
+	activity.Record(ctx, "stdout: %v", inv.stdout.String())
```
ekinanp: Wondering if we should reduce the verbosity here. When I do something like
```
➜  mnt107015990 stree
INFO Invoking /Users/enis.inan/GitHub/wash/schema.rb schema /schema ''
INFO stdout: {
  "root": {
  	"label": "schema",
  	"methods": ["list","schema"],
  	"singleton": true,
  	"children": [
  		"A"
  	]
  },
  "A": {
  	"label": "NodeA",
  	"methods": ["list","schema"],
  	"singleton": true,
  	"children": [
  		"B"
  	]
  },
  "B": {
  	"label": "NodeB",
  	"methods": ["read","schema"]
  }
}
.
├── docker
│   ├── containers
│   │   └── [container]
│   │       ├── log
│   │       ├── metadata.json
│   │       └── [fs]
│   │           ├── [dir]
│   │           │   ├── [dir]
│   │           │   └── [file]
│   │           └── [file]
│   └── volumes
│       └── [volume]
│           ├── [dir]
│           │   ├── [dir]
│           │   └── [file]
│           └── [file]
└── schema
    └── NodeA
        └── [NodeB]
```

It's a bit annoying having to see stdout all the time.

MikaelSmith: It seems wrong that we’re trying to write to activity on the CLI side.

ekinanp: Yeah, that's a good point -- that output would add-up the more external plugins you have. Should we file an issue about it?

plugin/externalPluginEntry.go
```diff
@@ -141,12 +187,101 @@ func (e *externalPluginEntry) ChildSchemas() []*EntrySchema {
...
+	var graph *linkedhashmap.Map
+	if e.schemaGraphs != nil {
+		g, ok := e.schemaGraphs[e.typeID]
+		if !ok {
+			msg := fmt.Errorf(
+				"e.Schema(): entry schemas were prefetched, but no schema graph was provided for %v (%v)",
+				ID(e),
+				e.typeID,
+			)
+			panic(msg)
```
ekinanp: Thinking this might be better if it were a warning.

plugin/externalPluginEntry.go
```diff
@@ -141,12 +187,101 @@ func (e *externalPluginEntry) ChildSchemas() []*EntrySchema {
...
+		} else {
+			for i := range instanceMethods {
+				if instanceMethods[i] != schemaMethods[i] {
+					log.Warnf(mismatchMsg)
+					break
+				}
+			}
+		}
+	} else {
+		// Entry schemas were not prefetched, so we'll need to shell out. Even though entry
```
ekinanp: In the gem, it would look something like
```
Wash.exclude_schema
```
so the code would know not to include the schema in root.

plugin/externalPluginScript.go
MikaelSmith: I'm confused why this code is being exercised client-side in the first place.

plugin/externalPluginEntry.go
```diff
@@ -58,29 +65,60 @@ func mungeToMethods(input []interface{}) (map[string]interface{}, error) {
 	return methods, nil
 }
 
-func (e decodedExternalPluginEntry) toExternalPluginEntry() (*externalPluginEntry, error) {
-	if e.Name == "" {
+func (e decodedExternalPluginEntry) toExternalPluginEntry(schemaKnown bool, isRoot bool) (*externalPluginEntry, error) {
+	if len(e.Name) <= 0 {
```
MikaelSmith: Why change this? The comparison to empty string is considered ideomatic.

plugin/externalPluginScript.go
ekinanp: I think it might have been during the time the activity wrote to the debug logs for all journals, so I put it there to make debugging easier.

plugin/externalPluginEntry.go
ekinanp: Apparently this is a bit more idiomatic -- I saw this type of comparison being used in the Go source. If I'm wrong, happy to change it back.

plugin/externalPluginEntry.go
MikaelSmith: I'm ok with panic for now.

plugin/externalPluginEntry.go
MikaelSmith: I guess they both are: https://stackoverflow.com/a/18595217/2048059. Since it's consistent with the other tests here, the new code is fine.

plugin/externalPluginEntry.go
MikaelSmith: Hmm. You start using log messages here, which we've avoided in plugins. They should use activity.

That suggests either:
- Schema needs to support returning an error (this makes sense to me specifically for external plugins)
- activity needs multiple log levels, and for warn/error we also expose that at the shell level.

I think both would be useful.

MikaelSmith: Since we can operate without a schema, the least changes would be to add different levels to the activity log so we can have higher severity issues, and also log them appropriately.

ekinanp: I can support the first option in the PR. Will update.

plugin/externalPluginScript.go
MikaelSmith: Also, rather than calling `.String()` we could replace `%v` with `%s` (which should do the same thing).

plugin/externalPluginEntry.go
ekinanp: (Documenting what we talked about on slack): Nevermind on the error handling support, reason for the warnings is b/c those are during development. Ideally once the external plugins finished, they'd prefetch their schema at the root so things would never error.

plugin/externalPluginScript.go
ekinanp: Looks like it doesn't. Here's what I get when I use `%s`:
```
INFO stdout: {{
  "type_id": "root",
	"name": "schema",
	"methods": [
			"schema",
		"list"
	]
}
 %!s(int=0) %!s(bytes.readOp=0)}
```

vs `%v` + `String()`:
```
INFO stdout: {
  "type_id": "root",
	"name": "schema",
	"methods": [
			"schema",
		"list"
	]
}
```

ekinanp:
Updated to return an error instead of log a warning.

plugin/externalPluginScript.go
MikaelSmith: Huh, that's surprising.

PR #386, opened 2019-07-12
Don't require type ID + continue supporting empty methods array

ekinanp:


plugin/testdata/externalPluginSchema.json
```diff
@@ -1160,9 +1160,7 @@
     "aws.ec2InstanceMetadataJSON": {
         "label": "metadata.json",
         "singleton": true,
-        "methods": [
-            "read"
-        ],
+        "methods": [],
```
ekinanp: This ensures that there's a test case to check that empty method arrays in entry schema JSONs are also supported.

Issue #365, opened 2019-07-02
Update documentation with EntrySchema changes


ekinanp:
This includes:
* A brief overview of entry schemas in the plugin tutorials
* Updating wash find's help text to include the new schema work. At the very least, this includes updating the meta primary's docs.
* Updating the swagger docs with /fs/schema's response once that is finalized.

PR #387, opened 2019-07-15
 Update docs with the entry schema changes

ekinanp:
Resolves https://github.com/puppetlabs/wash/issues/365

### DIFF
--- a/api/types/entrySchema.go
+++ b/api/types/entrySchema.go
@@ -16,7 +16,7 @@ import (
 // swagger:response
 type EntrySchema struct {
 	plugin.EntrySchema
-	children map[string]*EntrySchema
+	children []*EntrySchema
 	// graph is an ordered map of <TypeID> => <EntrySchema>. We store it to make
 	// MarshalJSON's implementation easier.
 	//
@@ -79,10 +79,9 @@ func (s *EntrySchema) UnmarshalJSON(data []byte) error {
 	// Fill-in the children map
 	s.graph.Each(func(key interface{}, value interface{}) {
 		schema := value.(*EntrySchema)
-		schema.children = make(map[string]*EntrySchema)
 		for _, childTypeID := range schema.EntrySchema.Children {
 			rawChild, _ := s.graph.Get(childTypeID)
-			schema.children[childTypeID] = rawChild.(*EntrySchema)
+			schema.children = append(schema.children, rawChild.(*EntrySchema))
 		}
 	})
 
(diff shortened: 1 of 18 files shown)


api/types/entrySchema.go
```diff
@@ -12,6 +12,8 @@ import (
 
 // EntrySchema describes an entry's schema, which
 // is what's returned by the /fs/schema endpoint.
+//
+// swagger:response
```
ekinanp: I've been having trouble installing swagger on my machine and getting it to work. Also, not sure how to document this since the type is not a 1:1 mapping with the response (the latter is a `<type_id> => <schema>` map w/ first key being entry's type ID). So could use some help with this part.

MikaelSmith: I'll try. Go 1.12 is still only part-way through the Go Modules migration, so I keep the wash project in GOPATH and run swagger without go modules (meaning I have to separately `go get` all the dependencies). Hopefully 1.13 will fix this.

MikaelSmith:
Travis is saying the website wasn't regenerated for all your changes.

api/types/entrySchema.go
MikaelSmith: I'll plan to update the swagger docs as a separate PR.

website/content/docs/_index.md
```diff
@@ -218,4 +219,42 @@ To make `wash find`'s filtering less tedious and better performing, entries can
...
+
+These questions can be generalized to any Wash entry.
+
+Entry schemas are a useful way to document your plugin's hierarchy without having to maintain a README. Users can view your hierarchy via the `stree` command. For example, if you invoke `stree docker` in a Wash shell (try it!), you should see something like
+
+```
+docker
+├── containers
+│   └── [container]
+│       ├── [fs]
```
MikaelSmith: Hmm, this should be a Singleton.

website/content/docs/_index.md
```diff
@@ -218,4 +219,42 @@ To make `wash find`'s filtering less tedious and better performing, entries can
...
+│       │   ├── [file]
```
MikaelSmith: Interesting that the order of `file/dir` is swapped in `dir`. It looks kind of weird.

website/content/docs/core_plugins/_index.md
```diff
@@ -10,16 +10,17 @@ The [plugin] package defines a set of interfaces that a plugin can implement to
...
-
-Use [activity.Record](https://godoc.org/github.com/puppetlabs/wash/activity) for all plugin-related logging. Each plugin method that Wash calls is passed a `context.Context` object that is initialized with a Journal ID for use with `activity.Record`.
+Each entry that implements the `Parent` interface must provide a schema for its children. Use [activity.Record](https://godoc.org/github.com/puppetlabs/wash/activity) for all plugin-related logging. Each plugin method that Wash calls is passed a `context.Context` object that is initialized with a Journal ID for use with `activity.Record`.
```
MikaelSmith: The first and later sentences still seem like they should be separate paragraphs. They're unrelated.

website/content/docs/external_plugins/_index.md
```diff
@@ -188,6 +189,8 @@ Below is an example that includes pre-fetched method results for a static direct
 
 `list` adopts the standard error convention described in the [Errors](#errors) section.
 
+**NOTE:** If entry schemas are `on`, then every entry returned by list must also include a `type_id` key.
```
MikaelSmith: `on` isn't a technical term. Maybe italicize its use instead.

api/types/entrySchema.go
ekinanp: Ok, sounds good.

website/content/docs/_index.md
ekinanp: Ah yeah, that's because `apitypes.EntrySchema` stores the children in a `map[string]*EntrySchema`. I think I can fix that to be an array instead so insertion order's preserved.

ekinanp:
Updated to address the feedback.

PR #392, opened 2019-07-15
Update api docs

MikaelSmith:
Add API docs for the /fs/schema endpoint and improve /fs/exec docs.


api/schema.go
```diff
@@ -8,6 +8,22 @@ import (
 	"github.com/puppetlabs/wash/plugin"
 )
 
+// swagger:route GET /fs/schema schema entrySchema
+//
+// Schema for an entry at path
+//
+// Returns an EntrySchema object describing the plugin schema starting at the given path.
```
ekinanp: Looking at some other types (e.g. `execBody/execResponse`), I think you can document this as `map[string]*EntrySchema` noting that the first key in the map corresponds to `<path>`'s schema.

MikaelSmith: Ah, because that's how it'll be marshaled?

MikaelSmith: And are those apitypes.EntrySchema or plugin.EntrySchema?

MikaelSmith: Looks kind of challenging to document.

MikaelSmith:
What it looks like now:
Screen Shot 2019-07-15 at 2 40 51 PM

PR #399, opened 2019-07-16
Add more examples to "Wash by example"

### LINKED ISSUE #393: Add more useful docs examples
Add a "find all containers that started today" example in the "Wash by example" section.

Mention that tab-completion makes navigation a cinch.

Add small snippets of using Wash w/ other plugins like Kubernetes, AWS, and maybe some cool external plugin. Even if people can't replicate your setup on their machines, they can still see the breadth of Wash (vs. thinking "Oh, this makes Docker easy. Maybe it doesn't make AWS or Kube easy, and I really don't care enough about yet another software project to see for myself").

MikaelSmith:
Resolves #393.

### DIFF
--- a/website/content/_index.md
+++ b/website/content/_index.md
@@ -66,9 +66,9 @@ docker-compose -f swarm/docker-compose.yml up -d
 
 > If you don't have `svn` installed, you can `git clone https://github.com/puppetlabs/wash.git` and prefix `swarm/docker-compose.yml` with `wash/examples`.
 
-> When done, run `docker-compose -f swarm/docker-compose.yml down` to stop the example application.
-
 This starts a small [Flask](http://flask.pocoo.org) webapp that keeps a count of how often it's been accessed in a [Redis](http://redis.io) instance that maintains state in a Docker volume.
+
+> When done, run `docker-compose -f swarm/docker-compose.yml down` to stop the example application.
 
 Navigate the filesystem to view running containers
 ```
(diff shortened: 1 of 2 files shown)


website/content/_index.md
```diff
@@ -120,12 +123,89 @@ Config:
...
+EOF
+```
+
+### Listing AWS resources
+
+Wash also includes support for AWS. If you have your own and you've configured the AWS CLI on your workstation, you'll be able to use Wash to explore EC2 instances and S3 buckets.
+
+As an example, you might want to periodically check how many EC2 instances are running (and display that via [BitBar](https://getbitbar.com/)):
+```
+running=`find aws -action exec -meta .State.Name running 2>/dev/null | wc -l | xargs`
```
ekinanp: I think this and the S3 bucket example would be more expressive via an `spath` primary. `spath` matches on the schema path, which are the labels joined by a `/`. So the EC2 example would be
```
find aws -spath '*ec2*instance' -meta .State.Name
```

and the S3 bucket example would be
```
find aws -spath '*s3*bucket'
```

Could file an issue if you think that'd be useful.

website/content/_index.md
```diff
@@ -120,12 +123,89 @@ Config:
...
 ```
 
 Try exploring `docker/volumes` to interact with the volume created for Redis.
 
+### Finding with metadata
+
+Wash includes a powerful `find` command that can match based on an entry's metadata. For example, if we wanted to see what containers were created today, we would look at the `.Created` entry for Docker containers and the `.metadata.creationTimestamp` for Kubernetes pods. We can find all instances created in the last 24 hours with
+
+```
+find -meta .Created -24h -o -meta .metadata.creationTimestamp -24h
```
ekinanp: Nit: This would need to set the `daystart` option to work properly. Maybe phrase the example as "instances created in the last 24 hours" instead of "today" ?

ekinanp (approved):

website/content/_index.md
MikaelSmith: I'm not against the spath primary, but I don't think it's a big improvement either.

website/content/_index.md
MikaelSmith: I do: `We can find all instances created in the last 24 hours with`

The "created today" phrasing is accurate, but maybe throwing you off because I then use a slightly different example.

MikaelSmith: Changed it to say "created recently".

PR #406, opened 2019-07-16
Expose important errors more visibly

### LINKED ISSUE #171: Wash seems to freeze when opened with 'fish' in certain folders (like the wash folder)
**Describe the bug**
When running `wash server mnt` in the directory I cloned wash into (~/Dev/wash) it seems to freeze the terminal when I try to take actions in that directory, such as ls.  This also impacted a df -h.  It doesn't happen when I run wash server mnt in a different directory (~/Dev/wash_workspace)

**To Reproduce**
1. Clone wash
2. Run `wash server mnt` in the wash directory
3. open a new terminal, cd into the wash directory and ls.

**Expected behavior**
I expect the ls to complete successfully.

**Logs**
❰mikker❙~❱✘≻ wash ls --debug ~/Dev/wash
Error: unknown flag: --debug
❰mikker❙~❱✘≻ wash ls --loglevel debug ~/Dev/wash
Error: unknown flag: --loglevel

**Environment (please complete the following information):**
Mac OS 10.13.1

**Additional context**
None

MikaelSmith:
Add an `activity.Warnf` function that bumps the severity up on messages in journals and logs them at warning level so they should appear in the shell. Also document some known issues about using Wash with particular shell environments.

Fixes #171.

### DIFF
--- a/activity/core.go
+++ b/activity/core.go
@@ -5,7 +5,9 @@
 //
 // Wash plugins should use
 //	activity.Record(ctx context.Context, msg string, a ...interface{})
-// to record entries. The context contains the Journal ID.
+// to record entries, and
+//  activity.Warnf(ctx context.Context, msg string, a ...interface{})
+// to warn about errors. The context contains the Journal ID.
 package activity
 
 import (
@@ -80,6 +82,25 @@ func Record(ctx context.Context, msg string, a ...interface{}) {
 	}
 
 	journal.Record(msg, a...)
+}
+
+// Warnf writes a new entry to the journal identified by the ID at `activity.JournalKey` in the
+// provided context at WARN level. It also writes to the server logs at the same level. Use Warnf
+// in plugin `Init` methods for issues during setup that will result in degraded behavior.
+func Warnf(ctx context.Context, msg string, a ...interface{}) {
+	journal, ok := ctx.Value(JournalKey).(Journal)
+	if !ok {
+		log.Warnf(msg, a...)
+		return
+	}
+
+	if journal.ID == "" {
+		journal = deadLetterOfficeJournal
+	} else {
+		journal.addToHistory()
+	}
+
+	journal.Warnf(msg, a...)
 }
 
 func closeJournal(id string, obj interface{}) {
--- a/cmd/rootMain.go
+++ b/cmd/rootMain.go
@@ -4,8 +4,10 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	"github.com/puppetlabs/wash/cmd/internal/server"
 	cmdutil "github.com/puppetlabs/wash/cmd/util"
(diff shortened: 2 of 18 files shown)


MikaelSmith:
I'm exploring some options for how to deal with the SIGTTIN. I think Wash gets suspended because it tries to prompt for input after it's ceded control to the child shell.

MikaelSmith:
I'm not sure how Wash is being suspended. It doesn't appear to be getting any signals, so it may be the shell it invokes that's suspended.

MikaelSmith:
Success! Added commits so we don't suspend and instead grab control of STDIN when prompting for input.

MikaelSmith:
http://curiousthing.org/sigttin-sigttou-deep-dive-linux may be useful background.

activity/core.go
```diff
@@ -82,6 +84,27 @@ func Record(ctx context.Context, msg string, a ...interface{}) {
 	journal.Record(msg, a...)
 }
 
+// Warnf writes a new entry to the journal identified by the ID at `activity.JournalKey` in the
+// provided context at WARN level. It also writes to the server logs at the same level. If no ID
+// is registered, the entry is written to the server logs at the info level. If the ID is an empty
+// string - which can happen when the JournalID header is missing from an API call - it uses the
+// ID 'dead-letter-office'.
+func Warnf(ctx context.Context, msg string, a ...interface{}) {
```
ekinanp: I think a tip on when to use `Warnf` vs. `Record` would be useful (e.g. what type of errors would a Warnf make sense vs. a Record), because that isn't quite clear to me.

cmd/rootMain.go
```diff
@@ -125,6 +127,9 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
 	// Configure logrus to emit simple text
 	log.SetFormatter(&log.TextFormatter{DisableTimestamp: true})
 
+	// Ignore tty errors so we aren't suspended when prompting for input from the background.
+	signal.Ignore(syscall.SIGTTIN, syscall.SIGTTOU, syscall.SIGTSTP)
```
ekinanp: Not sure I understand this. When would those TTY errors happen? What do those errors do? A longer comment or a reference would help.

plugin/interactive.go
```diff
@@ -22,13 +24,45 @@ func IsInteractive() bool {
...
+
 // Prompt prints the supplied message, then waits for input on stdin.
 func Prompt(msg string) (string, error) {
 	if !IsInteractive() {
-		return "", fmt.Errorf("Not an interactive session")
+		return "", fmt.Errorf("not an interactive session")
 	}
-	fmt.Fprintf(os.Stderr, "%s: ", msg)
+
+	inFd := int(os.Stdin.Fd())
```
ekinanp: I think a comment giving a high-level overview of this logic would help (e.g. precise meaning of "We control Stdin" and "If we don't control STDIN, here's how we get back that control temporarily").

ekinanp (commented):
Looks good code-wise, main concerns are more comments since some of the stuff here requires specialized knowledge.

MikaelSmith:
Updated.

cmd/rootMain.go
ekinanp: Thanks, this helps. So what is `SIGTTOU` for?

plugin/interactive.go
ekinanp: Much clearer. I still think it'd be useful to explicitly state the pgrp connection (e.g. "control of stdin" means "stdin's pgrp is our pgrp" and "temporarily controlling stdin" means "temporarily setting its pgrp to our pgrp") but that's up to you.

cmd/rootMain.go
MikaelSmith: Updated it. The actual behavior's weird, because we're getting SIGTTOU but everything I read suggests I should be getting SIGTTIN. I think it makes sense to ignore both (but I don't think we need to ignore SIGTSTP).

plugin/interactive.go
MikaelSmith: I think I'm going to leave it as-is. I don't feel like stating `stdin's pgrp is our pgrp` provides anything that's not explained by reading the code.

ekinanp (approved):

PR #413, opened 2019-07-18
Mark external plugin roots as singleton + generate stub type ID for schema-less roots

ekinanp:
This was a regression from a previous PR.


MikaelSmith:
They still all say the same name. I'll dig more into why.

MikaelSmith (approved):

PR #416, opened 2019-07-20
Namespace external plugin type IDs

### LINKED ISSUE #396: External plugin entry type IDs should be namespaced
It is probably enough to namespace as `<plugin_name>::<raw_type_id>`. This way, we minimize type ID conflicts.

ekinanp:
This minimizes the chance of conflicting type IDs. Type IDs are
namespaced as <plugin_name>::<raw_type_id>.

Resolves https://github.com/puppetlabs/wash/issues/396

### DIFF
--- a/plugin/entrySchema.go
+++ b/plugin/entrySchema.go
@@ -47,9 +47,7 @@ func Schema(e Entry) (*EntrySchema, error) {
 				// Create a schema for root so that `stree <mountpoint>` can still display
 				// it.
 				childSchema = NewEntrySchema(root, CName(root))
-				// TODO: Namespace these to something like <root_name>::Root once
-				// https://github.com/puppetlabs/wash/issues/396 is resolved.
-				childSchema.TypeID = "__" + root.name() + "__"
+				childSchema.TypeID = namespace(root.name(), "Root")
 			}
 			childSchema.IsSingleton()
 			schema.Children = append(schema.Children, childSchema.TypeID)
--- a/plugin/externalPluginEntry.go
+++ b/plugin/externalPluginEntry.go
@@ -530,6 +530,16 @@ func (e *externalPluginEntry) rawTypeID() string {
 func (e *externalPluginEntry) splitTypeID() (string, string) {
 	substrings := strings.SplitN(e.TypeID(), "::", 2)
 	if len(substrings) < 2 {
+		// We can hit this case if only some of the entries specify type IDs.
+		// This is impossible in the (common) entry schema scenario because
+		// entry schema support always requires type IDs. So, it is not an
+		// error.
+		//
+		// If we hit this case, then the concept of a type ID doesn't make
+		// sense so we return empty here. Note that returning substrings[0]
+		// is not a good idea b/c the plugin author could namespace their
+		// type IDs with "::", so substrings[0] would contain that initial
+		// chunk.
 		return "__unknown__", ""
 	}
 	return substrings[0], substrings[1]
@@ -559,7 +569,7 @@ func (e *externalPluginEntry) unmarshalSchemaGraph(stdout []byte) (*linkedhashma
 		return nil, fmt.Errorf("expected a non-empty JSON object")
 	}
 	if rawGraph[rawTypeID] == nil {
-		return nil, fmt.Errorf("%v's schema is missing", e.rawTypeID())
+		return nil, fmt.Errorf("%v's schema is missing", rawTypeID)
 	}
 
 	// Convert each node in the rawGraph to an EntrySchema object. This step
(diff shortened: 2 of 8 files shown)


plugin/entrySchema.go
```diff
@@ -47,9 +47,7 @@ func Schema(e Entry) (*EntrySchema, error) {
 				// Create a schema for root so that `stree <mountpoint>` can still display
 				// it.
 				childSchema = NewEntrySchema(root, CName(root))
-				// TODO: Namespace these to something like <root_name>::Root once
-				// https://github.com/puppetlabs/wash/issues/396 is resolved.
-				childSchema.TypeID = "__" + root.name() + "__"
+				childSchema.TypeID = namespace(root.name(), "Root")
```
MikaelSmith: None of the tests seem to test this.

MikaelSmith: Oh, it doesn't test for the namespace form, it uses namespace to generate the expected result.

plugin/externalPluginEntry.go
```diff
@@ -516,12 +516,40 @@ func newStdoutDecodeErr(ctx context.Context, decodedThing string, reason error,
...
+func (e *externalPluginEntry) rawTypeID() string {
+	_, rawTypeID := e.splitTypeID()
+	return rawTypeID
+}
+
+// splits the type ID into a tuple (<plugin_name>, <raw_type_id>)
+func (e *externalPluginEntry) splitTypeID() (string, string) {
+	substrings := strings.SplitN(e.TypeID(), "::", 2)
+	if len(substrings) < 2 {
+		return "__unknown__", ""
```
MikaelSmith: Should the 2nd return `substrings[0]`? I assume SplitN always returns at least one, for an empty string.

plugin/externalPluginEntry.go
```diff
@@ -530,8 +558,8 @@ func (e *externalPluginEntry) unmarshalSchemaGraph(stdout []byte) (*linkedhashma
 	if len(rawGraph) <= 0 {
 		return nil, fmt.Errorf("expected a non-empty JSON object")
 	}
-	if rawGraph[e.typeID] == nil {
-		return nil, fmt.Errorf("%v's schema is missing", e.typeID)
+	if rawGraph[rawTypeID] == nil {
+		return nil, fmt.Errorf("%v's schema is missing", e.rawTypeID())
```
MikaelSmith: We have rawTypeID, use it directly instead of calling the function?

plugin/testdata/externalPluginSchema.json
```diff
@@ -1134,7 +1134,7 @@
         "children": [
             "aws.ec2InstanceConsoleOutput",
             "aws.ec2InstanceMetadataJSON",
-            "volume.FS"
+            "aws.volume.FS"
```
MikaelSmith: This change seems strange. Also not sure what caused it. It's not the plugin prefix, that'd be `aws::`.

MikaelSmith (changes requested):

plugin/externalPluginEntry.go
ekinanp: Oh yeah, that was from an earlier iteration.

plugin/externalPluginEntry.go
ekinanp: Hmm, that case would correspond to some entries specifying type IDs and some entries not specifying type IDs, which is impossible in the (common) entry schema scenario. I figured if we hit that case, then the type ID doesn't make sense. Returning `substrings[0]` might not make sense either in case the author namespaces their type IDs using `::`, but maybe we could document something along the lines of "If you're using type IDs w/o entry schema support, then don't namespace them with '::' else things will go wrong".

I don't have strong feelings about either approach.

MikaelSmith: I think it's worth commenting how we might get here and it's not an error.

plugin/testdata/externalPluginSchema.json
ekinanp: Oh, that was to make testing easier. I do a global find/replace in https://github.com/ekinanp/wash/blob/namespace-type-ids/plugin/externalPluginEntry_test.go#L919 and that was originally failing b/c the `volume` entries were not prefixed with `aws.`. I figured it was easiest to just add that prefix, but I guess I could also add another `Replace` line there (or maybe use a regex instead).

I like the global find/replace there b/c it makes the test symmetric with the operation it's testing (e.g. the schema captured in the schema graph is the user's schema but w/ methods replaced by actions and type IDs namespaced by plugin-name).

MikaelSmith: Ah. Might be trickier to regenerate.

ekinanp:
Updated

plugin/testdata/externalPluginSchema.json
ekinanp: I guess regenerating the fixtures wouldn't hurt, given the previous type ID changes of making the paths unique.

MikaelSmith (approved):

plugin/testdata/externalPluginSchema.json
ekinanp: Actually I think I'll do that later, when it is needed.

PR #419, opened 2019-07-22
Use plugin.Schema() in `wash validate`

ekinanp:

### DIFF
--- a/cmd/validate.go
+++ b/cmd/validate.go
@@ -179,11 +179,7 @@ type criteria struct {
 
 func newCriteria(entry plugin.Entry) criteria {
 	var crit criteria
-	if schema := entry.Schema(); schema != nil {
-		crit.label = schema.Label
-		crit.typeID = schema.TypeID
-		crit.singleton = schema.Singleton
-	}
+	crit.typeID = plugin.TypeID(entry)
 	crit.list = plugin.ListAction().IsSupportedOn(entry)
 	crit.read = plugin.ReadAction().IsSupportedOn(entry)
 	crit.stream = plugin.StreamAction().IsSupportedOn(entry)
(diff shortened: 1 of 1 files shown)


cmd/validate.go
```diff
@@ -289,7 +297,11 @@ func processEntry(ctx context.Context, pw progress.Writer, wp cmdutil.Pool, e pl
 					continue
 				}
 
-				ccrit := newCriteria(entry)
+				ccrit, err := newCriteria(entry)
```
MikaelSmith: We don't need to shell out here, so I'd prefer to separate criteria as a hash object from how we stringify it (which needs more info from the schema).

ekinanp: plugin.Schema is going to shell out if the external plugin doesn't prefetch its schemas, so not sure what you mean here.

MikaelSmith: Didn't we separate TypeID so you don't need to call `schema` for it? That's the only one required by the hash. The label and singleton are only used when deciding how to display it, which we do once at the beginning.

ekinanp: Oh I see what you're saying. Will update.

cmd/validate.go
```diff
@@ -292,7 +305,7 @@ func processEntry(ctx context.Context, pw progress.Writer, wp cmdutil.Pool, e pl
 				ccrit := newCriteria(entry)
 				// If we have a TypeID, only explore children if they are different from the parent.
 				// This prevents simple recursion like volume directories containing more dirs.
-				if ccrit.typeID == "" || ccrit != crit {
+				if ccrit.typeID == "" || ccrit.typeID != crit.typeID {
```
ekinanp: Wonder if this should be `ccrit.typeID == "" || (crit.typeID != "" && ccrit.typeID != crit.typeID)`

MikaelSmith (approved):

PR #420, opened 2019-07-22
Make Validate work better with external plugins

MikaelSmith:
Ensure stdin reads as EOF if we don't have actual input to send. Also fix external plugin stream closing so it correctly ends plugin execution.

### DIFF
--- a/plugin/externalPluginEntry.go
+++ b/plugin/externalPluginEntry.go
@@ -419,6 +419,7 @@ func (e *externalPluginEntry) Stream(ctx context.Context) (io.ReadCloser, error)
 	select {
 	case err := <-headerRdrCh:
 		if err != nil {
+			cmd.Terminate()
 			defer wait()
 			// Try to get more context from stderr
 			n, readErr := inv.stderr.ReadFrom(stderrR)
@@ -434,6 +435,7 @@ func (e *externalPluginEntry) Stream(ctx context.Context) (io.ReadCloser, error)
 		}()
 		return &stdoutStreamer{cmd, stdoutR}, nil
 	case <-timer:
+		cmd.Terminate()
 		defer wait()
 		// We timed out while waiting for the streaming header to appear.
 		// Return an appropriate error message using whatever was printed
@@ -506,6 +508,7 @@ func (s *stdoutStreamer) Read(p []byte) (int, error) {
 }
 
 func (s *stdoutStreamer) Close() error {
+	s.cmd.Terminate()
 	return s.cmd.Wait()
 }
 
--- a/plugin/externalPluginEntry.go
+++ b/plugin/externalPluginEntry.go
@@ -470,6 +470,10 @@ func (e *externalPluginEntry) Exec(ctx context.Context, cmd string, args []strin
 	cmdObj.SetStderr(execCmd.Stderr())
 	if opts.Stdin != nil {
 		cmdObj.SetStdin(opts.Stdin)
+	} else {
+		// Go's exec.Cmd reads from the null device if no stdin is provided. We instead provide
+		// an empty string for input so plugins can test whether there is content to read.
+		cmdObj.SetStdin(strings.NewReader(""))
 	}
 	activity.Record(ctx, "Starting %v", cmdObj)
 	if err := cmdObj.Start(); err != nil {
--- a/plugin/internal/command.go
+++ b/plugin/internal/command.go
@@ -13,8 +13,8 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/puppetlabs/wash/activity"
 	"github.com/kballard/go-shellquote"
+	"github.com/puppetlabs/wash/activity"
 )
 
 // Command is a wrapper to exec.Cmd. It handles context-cancellation cleanup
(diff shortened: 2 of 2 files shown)


plugin/externalPluginEntry.go
```diff
@@ -470,6 +470,8 @@ func (e *externalPluginEntry) Exec(ctx context.Context, cmd string, args []strin
 	cmdObj.SetStderr(execCmd.Stderr())
 	if opts.Stdin != nil {
 		cmdObj.SetStdin(opts.Stdin)
+	} else {
+		cmdObj.SetStdin(strings.NewReader(""))
```
ekinanp: I don't like this because it doesn't account for `echo -n | cat` and `cat`'s distinct behavior.

plugin/externalPluginEntry.go
```diff
@@ -419,7 +422,7 @@ func (e *externalPluginEntry) Stream(ctx context.Context) (io.ReadCloser, error)
 	select {
 	case err := <-headerRdrCh:
 		if err != nil {
-			defer wait()
```
ekinanp: Is this safe? I recall previous google searches indicating that you always want to read from stdout/stderr before calling wait, which is why I wrapped it in a defer.

plugin/externalPluginEntry.go
```diff
@@ -372,25 +372,28 @@ func (e *externalPluginEntry) Metadata(ctx context.Context) (JSONObject, error)
 }
 
 func (e *externalPluginEntry) Stream(ctx context.Context) (io.ReadCloser, error) {
-	inv := e.script.NewInvocation(ctx, "stream", e)
+	cancelCtx, cancelFunc := context.WithCancel(ctx)
```
ekinanp: Hmm, why not call cancelFunc before closing the stream in https://github.com/puppetlabs/wash/blob/master/cmd/validate.go#L329 ?

plugin/externalPluginEntry.go
MikaelSmith: I don't believe the docs you pointed to were saying that you should wait before reading from the stdout and stderr readers. I think it was saying that wait would exit when they'd been buffered in. But there's some room for ambiguity.

plugin/externalPluginEntry.go
MikaelSmith: I suppose that would work. But the semantics of closing the stream would still be wrong.

plugin/externalPluginEntry.go
MikaelSmith: When you execute just `cat`, it's waiting for you to supply stdin. This seems like it does capture that difference.

plugin/externalPluginEntry.go
ekinanp: Yeah, the ambiguity is why I wrapped that in a defer. https://blog.kowalczyk.info/article/wOYk/advanced-command-execution-in-go-with-osexec.html also seems to indicate that wait should be called once stdout/stderr were read.

plugin/externalPluginEntry.go
ekinanp: What are the semantics of closing the stream? I was thinking something like "Close is only called when the context is cancelled."

plugin/externalPluginEntry.go
MikaelSmith: Oh, we are using pipes. So looks like we might need to cancel the context but still defer wait to make sure we capture everything. Previously I don't think the 5s timeout worked if the plugin just hung, because it would wait for the pipes to close but they never would.

plugin/externalPluginEntry.go
MikaelSmith: io.Closer has its own definition. It feels weird to say you should only call this function after the context is cancelled. But it would be a smaller change for sure.

ekinanp: Ah ok, so this is to make sure that stdoutStreamer properly implements io.Closer semantics. That makes sense. If that's the case, maybe an `internal.Command#Terminate/Kill` helper could be useful? It's more expressive of what you want to do when closing the stream, but also a bit more invasive.

ekinanp: I think if you choose the `cancelCtx` approach, a comment explaining why it's used would help (something like "We use cancelCtx to ensure that stdoutStreamer kills the process if it is closed for some other reason besides context cancellation").

plugin/externalPluginEntry.go
ekinanp: Think this should be removed since we've resolved the Stdin issue earlier

MikaelSmith: I think this was still necessary. If there's no Stdin provided, go exec reads `/dev/null` but I think we really want it to say that it's empty and closed.

MikaelSmith: I can back it out if you want to explore this more separately.

ekinanp: Ah, that's ok then. A comment explaining what you just said would help.

ekinanp (approved):
LGTM.  https://github.com/puppetlabs/wash/pull/420#discussion_r306053685 is the only suggestion I have but it is not a blocker.

PR #422, opened 2019-07-23
Added puppetwash to list of plugins

timidri:
Signed-off-by: Dimitri Tischenko <dimitri@puppet.com>


ekinanp (approved):

ekinanp:
CI failed because the website wasn't updated with the new docs. Wonder if there's a way to make that less annoying. @MikaelSmith ?

MikaelSmith:
Install hugo and run it according to https://github.com/puppetlabs/wash/tree/master/website. I think the answer will be GitHub Actions, but we don't have that enabled yet.

ekinanp:
👍 Should probably file an issue to get that enabled, because something like this would annoy me if I were a community member.

PR #427, opened 2019-07-24
Fix panic running Exec on EC2 instance with null fields

### LINKED ISSUE #426: Panic running Exec on AWS EC2 instance with limited permissions
I tried to list the `fs` directory for an EC2 instance that I have limited permissions to access, and it paniced
```
2019/07/24 15:34:11 fuse: panic in handler for Read [ID=0x2 Node=0x68 Uid=501 Gid=20 Pid=99511] 0x2 3276 @0x0 dir=true fl=0 lock=0 ffl=OpenReadOnly: interface conversion: interface {} is nil, not string
goroutine 2809 [running]:
bazil.org/fuse/fs.(*Server).serve.func2(0x5fc75e0, 0xc0003be1c0, 0xc000643ef8, 0xc000643e67)
	/Users/michaelsmith/.go/pkg/mod/bazil.org/fuse@v0.0.0-20180421153158-65cc252bf669/fs/serve.go:857 +0x1a2
panic(0x59236e0, 0xc003687d10)
	/usr/local/Cellar/go/1.12.7/libexec/src/runtime/panic.go:522 +0x1b5
github.com/puppetlabs/wash/plugin/aws.(*ec2Instance).Exec(0xc003dd25a0, 0x5fdc5e0, 0xc003687cb0, 0x5c2f435, 0x4, 0xc0037f4550, 0xd, 0xd, 0x0, 0x0, ...)
	/Users/michaelsmith/.go/src/github.com/puppetlabs/wash/plugin/aws/ec2Instance.go:274 +0xa08
```

It would appear that some of the metadata fields we try to read can be null when we assume they'll be strings.

MikaelSmith:
Test that string fields aren't null before trying to convert them to strings.

Fixes #426.

### DIFF
--- a/plugin/aws/ec2Instance.go
+++ b/plugin/aws/ec2Instance.go
@@ -258,16 +258,16 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
 		return nil, err
 	}
 	var hostname string
-	if name, ok := meta["PublicDnsName"]; ok {
+	if name, ok := meta["PublicDnsName"]; ok && name != nil {
 		hostname = name.(string)
-	} else if ipaddr, ok := meta["PublicIpAddress"]; ok {
+	} else if ipaddr, ok := meta["PublicIpAddress"]; ok && ipaddr != nil {
 		hostname = ipaddr.(string)
 	} else {
 		return nil, fmt.Errorf("No public interface found for %v", inst)
 	}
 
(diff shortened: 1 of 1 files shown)


ekinanp (approved):

MikaelSmith:
Another option would be to test the conversion to string. I don't think that'll happen in practice, so 🤷‍♂

PR #429, opened 2019-07-24
Add GCP plugin

### LINKED ISSUE #390: GCP with compute and storage
We previously had this. Let's re-add them with the new plugin framework.

MikaelSmith:
Adds a GCP plugin with Google Compute Engine support.

Resolves #390.


MikaelSmith:
This is still a work in progress. I have some basic metadata and console output working, but not really complete. And Exec - based on https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys - is going to take some time. The goal is to have it as smooth as `gcloud compute ssh`.

MikaelSmith:
I have a few outstanding items, but I think the code that's present is pretty complete
- [ ] add tests
- [ ] provide an expiration date for SSH public keys that are uploaded to a project/instance
- [ ] check that this works for instances that explicitly prohibit project-wide SSH keys

I'm also open to suggestions on things that should be configurable, with supporting use-cases.

activity/journal.go
```diff
@@ -103,7 +103,7 @@ func (j Journal) Warnf(msg string, a ...interface{}) {
 // appends the message to that journal. Journals are stored in the user's cache directory under
 // `wash/activity/ID.log`.
 func (j Journal) Record(msg string, a ...interface{}) {
-	log.Debugf(msg, a...)
+	log.Printf(msg, a...)
```
ekinanp: Won't this cause a lot of noise to the shell experience (e.g. seeing info messages when I'm doing a wash exec)?

ekinanp:
Could you paste the stree output here? Curious to see what this looks like.

plugin/gcp/computeDir.go
```diff
@@ -0,0 +1,68 @@
...
+		EntryBase:             plugin.NewEntry("compute"),
+		computeProjectService: computeProjectService{Service: svc, projectID: projID},
+	}, nil
+}
+
+// List all services as dirs.
+func (c *computeDir) List(ctx context.Context) ([]plugin.Entry, error) {
+	var entries []plugin.Entry
+	instReq := c.Instances.AggregatedList(c.projectID)
+	if err := instReq.Pages(ctx, func(instancePage *compute.InstanceAggregatedList) error {
```
ekinanp: Nit: Could `err :=` move out of the if statement? Bit hard to read as it is right now.

plugin/gcp/root.go
```diff
@@ -0,0 +1,101 @@
+// Package gcp presents a filesystem hierarchy for Google Cloud Platform resources.
+//
+// It follows https://cloud.google.com/docs/authentication/production to find your credentials:
```
ekinanp: Isn't this already documented? Why state it more than once?

plugin/gcp/storageBucket.go
```diff
@@ -0,0 +1,70 @@
...
+	it := bucket.Objects(ctx, &storage.Query{Delimiter: delimiter, Prefix: prefix})
+	for {
+		objAttrs, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+		if objAttrs.Prefix != "" {
```
ekinanp: A high-level comment describing these checks (and how the storage API works) would be useful, because this isn't obvious to me w/o having to dig into the package myself.

plugin/gcp/computeDir.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+type computeProjectService struct {
+	*compute.Service
+	projectID string
+}
+
+type computeDir struct {
+	plugin.EntryBase
+	computeProjectService
```
ekinanp: Nit: Think it's better if embedding's only restricted to `EntryBase`. That way, it's easier to see that this is strictly an Entry, and that we're just passing the project service object around. The pattern's also consistent with the other core-plugins.

plugin/gcp/storageObjectPrefix.go
```diff
@@ -0,0 +1,42 @@
...
+
+type storageObjectPrefix struct {
+	plugin.EntryBase
+	bucket *storage.BucketHandle
+	prefix string
+}
+
+// Takes the name of the directory, as well as the full prefix path.
+// Attrs may be nil if they could not be retrieved. Some prefixes don't appear to have attributes.
+func newStorageObjectPrefix(bucket *storage.BucketHandle,
```
ekinanp: Strange that prefixes here have attributes but S3 prefixes don't.

plugin/gcp/storageObjectPrefix.go
```diff
@@ -0,0 +1,42 @@
...
+	name, prefix string, attrs *storage.ObjectAttrs) *storageObjectPrefix {
+	pre := &storageObjectPrefix{
+		EntryBase: plugin.NewEntry(name),
+		bucket:    bucket,
+		prefix:    prefix,
+	}
+	if attrs != nil {
+		pre.Attributes().SetMtime(attrs.Updated).SetSize(uint64(attrs.Size)).SetMeta(attrs)
```
ekinanp: What does Size mean here? Kind of weird seeing it set w/o implementing Readable.

ekinanp (commented):
Left some initial comments. I'll review the exec stuff tomorrow morning.

transport/ssh.go
```diff
@@ -39,51 +40,139 @@ func newAgent() (ssh.AuthMethod, error) {
...
 		sshConfig := &ssh.ClientConfig{
-			User:            user,
+			User:            conf.user,
 			Auth:            authmethod,
-			HostKeyCallback: hostKeyCallback,
+			HostKeyCallback: conf.hostKeyCallback,
+		}
+
+		// Retry until period expires or connection is established.
+		if retries > 0 {
```
ekinanp: Nit: You could also do `if retries <= 0 { retries = 1 }` to centralize this code.

activity/journal.go
MikaelSmith: Wash the shell defaults to `warn` log level. The commit message talks about the rationale a bit more.

MikaelSmith:
stree output (after I fixed it)
```
gcp
└── [project]
    ├── compute
    │   └── [instance]
    │       └── console.out
    └── compute
        └── [bucket]
            ├── [prefix]
            │   ├── [prefix]
            │   └── [object]
            └── [object]
```

activity/journal.go
MikaelSmith: The one area it could be messy is if we do #302, but in that case I think activity journal tracking becomes more complicated, and we need to send the logs back to the client that originated them anyway.

plugin/gcp/root.go
MikaelSmith: Different audiences. But you're probably right that it doesn't need to be in the godocs.

ekinanp (commented):
Overall Exec looks good. Just some comments.

activity/journal.go
ekinanp: Hmm, ok. When I do `./wash`, I still see INFO messages popping up when I try to do stuff (including server start up messages). This is from the latest Wash build.

plugin/gcp/root.go
ekinanp: Yeah, that's one less piece of documentation to maintain.

activity/journal.go
MikaelSmith: Do you have `loglevel: info` set in your config?

ekinanp: Ah that's why. Cool, I don't have any other concerns around this.

plugin/gcp/storageObjectPrefix.go
MikaelSmith: Yeah. Surprised me.

plugin/gcp/storageObjectPrefix.go
MikaelSmith: Usually 0. Directories do not universally have size 0 however. https://unix.stackexchange.com/a/77. I would interpret it as the size of the readdir response.

plugin/gcp/computeInst.go
```diff
@@ -36,3 +49,244 @@ func (c *computeInstance) ChildSchemas() []*plugin.EntrySchema {
...
+		return err
+	}
+
+	return nil
+}
+
+var errKeyAdded = fmt.Errorf("Key added to metadata")
+
+// SSH public keys are added to Common Instance Metadata as described in
+// https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys
```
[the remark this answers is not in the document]
MikaelSmith: > Adding the key to metadata in the prescribed format (mapping user to public key content) causes it to be added to `~/.ssh/authorized_keys` for the supplied user on associated instances.

I'll add it as a comment.

plugin/gcp/computeInst.go
```diff
@@ -36,3 +49,244 @@ func (c *computeInstance) ChildSchemas() []*plugin.EntrySchema {
...
+			return keys, fmt.Errorf("errors reading %v - %v - and %v - %v",
+				keys.publicKey, err1, keys.privateKey, err2)
+		}
+	}
+
+	return keys, nil
+}
+
+func generateKeys(publicKeyPath, privateKeyPath string) error {
+	privateKey, err := rsa.GenerateKey(rand.Reader, 4096)
```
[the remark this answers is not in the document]
MikaelSmith: I'm not sure what you mean. It's pretty much the only clear way to do it: generate an RSA key, serialize the public and private portions, write to file. There's no higher level abstraction for it.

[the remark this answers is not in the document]
ekinanp: Cool, then the comment's unnecessary. The reason I asked the Q was b/c there's a bunch of crypto stuff out there (most of which I don't know) that I was thinking maybe there was another way to generate these keys.

transport/ssh.go
MikaelSmith: I don't think that would have the same meaning.

MikaelSmith:
Updated, but I still have some TODOs to get to.

MikaelSmith:
Pushed an update to revamp how we check and update SSH keys so it works with more configurations.

MikaelSmith:
The key expiration doesn't do what I hoped it would (remove the key from the VM when it expires), so I left it out.

ekinanp (approved):

ekinanp (approved):
Comment's not a PR blocker.

plugin/metadataJSON.go
```diff
@@ -0,0 +1,54 @@
...
+
+// NewMetadataJSONFile creates a new MetadataJSONFile. If caching Metadata on the `other` entry is
+// disabled, it will use that to compute the file size upfront.
+func NewMetadataJSONFile(ctx context.Context, other Entry) (*MetadataJSONFile, error) {
+	meta := &MetadataJSONFile{
+		EntryBase: NewEntry("metadata.json"),
+		other:     other,
+	}
+
+	if other.getTTLOf(MetadataOp) == -1 {
```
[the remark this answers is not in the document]
MikaelSmith: It's not possible to set negative TTL via CachedOp. So it'd only be 0, which means "use the cache's default". DisableCachingFor sets to a negative (-1). So I'll change it to `< 0`, and add a comment about what 0 means for CachedOp.

PR #437, opened 2019-07-31
Namespace core plugin type IDs

### LINKED ISSUE #436: Namespace core plugin type IDs
We should probably do something like `<plugin_name>::<type_name>` so e.g. `aws::ec2Instance`; `docker::container`, etc.

Reason for this is (1) maintains symmetry w/ external plugin type IDs and (2) makes it easier to keep track of resource types for Google Analytics (enabling the latter is why this issue was filed).

ekinanp:
This commit also centralizes pluginName + rawTypeID.

Resolves #436

### DIFF
--- a/api/fs/dir.go
+++ b/api/fs/dir.go
@@ -36,12 +36,11 @@ func (d *dir) List(ctx context.Context) ([]plugin.Entry, error) {
 }
 
 func (d *dir) ChildSchemas() []*plugin.EntrySchema {
-	return []*plugin.EntrySchema{
-		d.Schema(),
-		(&file{fsnode: &fsnode{}}).Schema(),
-	}
+	return nil
 }
 
 func (d *dir) Schema() *plugin.EntrySchema {
-	return plugin.NewEntrySchema(d, "dir")
+	// Schema only makes sense for core plugins. apifs isn't a core
+	// plugin.
+	return nil
 }
--- a/plugin/entrySchema.go
+++ b/plugin/entrySchema.go
@@ -3,6 +3,7 @@ package plugin
 import (
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/ekinanp/jsonschema"
 	"github.com/emirpasic/gods/maps/linkedhashmap"
@@ -13,14 +14,13 @@ const registrySchemaLabel = "mountpoint"
 // TypeID returns the entry's type ID. It is needed by the API,
 // so plugin authors should ignore this.
 func TypeID(e Entry) string {
-	// Note that this helper's necessary because Schema() can
-	// shell out for external plugins, which is expensive.
-	if ep, ok := e.(externalPlugin); ok {
-		return ep.TypeID()
+	pluginName := pluginName(e)
+	rawTypeID := rawTypeID(e)
+	if pluginName == "" {
+		// e is the plugin registry
+		return rawTypeID
 	}
-	// e is a core plugin entry
-	t := unravelPtr(reflect.TypeOf(e))
-	return t.PkgPath() + "/" + t.Name()
+	return namespace(pluginName, rawTypeID)
 }
 
 // Schema returns the entry's schema. It is needed by the API,
@@ -36,7 +36,7 @@ func Schema(e Entry) (*EntrySchema, error) {
 		// registry's schema graph.
 		schema := NewEntrySchema(t, registrySchemaLabel).IsSingleton()
 		schema.graph = linkedhashmap.New()
-		schema.graph.Put(schema.TypeID, &schema.entrySchema)
+		schema.graph.Put(TypeID(t), &schema.entrySchema)
 		for _, root := range t.pluginRoots {
 			childSchema, err := Schema(root)
 			if err != nil {
@@ -47,10 +47,9 @@ func Schema(e Entry) (*EntrySchema, error) {
 				// Create a schema for root so that `stree <mountpoint>` can still display
 				// it.
 				childSchema = NewEntrySchema(root, CName(root))
-				childSchema.TypeID = namespace(root.name(), "Root")
 			}
 			childSchema.IsSingleton()
-			schema.Children = append(schema.Children, childSchema.TypeID)
+			schema.Children = append(schema.Children, TypeID(childSchema.entry))
 			childGraph := childSchema.graph
 			if childGraph == nil {
 				// This is a core-plugin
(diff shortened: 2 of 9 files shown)


api/fs/dir.go
```diff
@@ -36,12 +36,11 @@ func (d *dir) List(ctx context.Context) ([]plugin.Entry, error) {
...
 
 func (d *dir) ChildSchemas() []*plugin.EntrySchema {
-	return []*plugin.EntrySchema{
-		d.Schema(),
-		(&file{fsnode: &fsnode{}}).Schema(),
-	}
+	return nil
 }
 
 func (d *dir) Schema() *plugin.EntrySchema {
```
ekinanp: I wonder if we should just remove this package -- the shell experience cds to the mountpoint anyways so in the common case, users will not touch local files/dirs.

If we still want things to work with local files/dirs, I can file an issue to move this code over to a client wrapper.

plugin/externalPluginEntry.go
```diff
@@ -532,44 +532,8 @@ func newStdoutDecodeErr(ctx context.Context, decodedThing string, reason error,
...
-func namespace(pluginName string, typeID string) string {
-	if len(typeID) == 0 {
-		typeID = "unknown"
-	}
-	return pluginName + "::" + typeID
-}
-
-func (e *externalPluginEntry) unmarshalSchemaGraph(stdout []byte) (*linkedhashmap.Map, error) {
-	pluginName, rawTypeID := e.splitTypeID()
+func unmarshalSchemaGraph(e externalPlugin, stdout []byte) (*linkedhashmap.Map, error) {
```
ekinanp: Refactored this so that the casing in pluginName would work properly if e was an externalPluginRoot.

ekinanp:
How should analytics handle shared entries like `volume.file`, `volume.dir`? With the changes here, they'd be `<plugin_name>::volume.file`, `<plugin_name>::volume.dir`. Is that OK? Should we fine-grain them more to `<plugin_name>::fs::volume.file, `<plugin_name>::fs::volume.dir`?

I think it makes sense to treat `volume.file`/`volume.dir` as a single resource within the plugin (so am leaning towards `<plugin_name>::volume.file`, `<plugin_name>::volume.dir`).

plugin/entrySchema.go
```diff
@@ -259,11 +258,55 @@ func (s *EntrySchema) schemaOf(obj interface{}) (*JSONSchema, error) {
...
+	case externalPlugin:
+		rawTypeID := t.RawTypeID()
+		if rawTypeID == "" {
+			rawTypeID = "unknown"
+		}
+		return rawTypeID
+	default:
+		// e is either a core plugin entry or the plugin registry itself
+		reflectType := unravelPtr(reflect.TypeOf(e))
+		return reflectType.PkgPath() + "/" + reflectType.Name()
```
ekinanp: Per https://github.com/puppetlabs/wash/pull/437#issuecomment-516696489, I'd like this to return `reflectType.Name()`. However with volume files/dirs, this would mean we'd get `dir`, `file` for the raw type IDs so the overall type ID would be `docker::dir`, `docker::file`, which isn't readable. Changing `volume.dir` => `volume.volumeDir`, `volume.file` => `volume.volumeFile` would be one way to solve it.

api/fs/dir.go
MikaelSmith: I think the goal is for the shell experience to feel contained in that. I have issues up to display the path starting from the Wash root, and for `cd` to take you back to the Wash root. So I think it should be fine to error if you try to use Wash tools outside that.

plugin/entrySchema.go
MikaelSmith: How would using `reflectType.Name()` show up for other types? I'm not sure what the trade-offs are.

plugin/externalPluginRoot.go
```diff
@@ -99,27 +99,27 @@ func (r *externalPluginRoot) WrappedTypes() SchemaMap {
 
 // partitionSchemaGraph partitions graph into a map of <type_id> => <schema_graph>
 func (r *externalPluginRoot) partitionSchemaGraph(graph *linkedhashmap.Map) map[string]*linkedhashmap.Map {
-	var populate func(*linkedhashmap.Map, entrySchema, map[string]bool)
-	populate = func(g *linkedhashmap.Map, node entrySchema, visited map[string]bool) {
-		if visited[node.TypeID] {
+	var populate func(*linkedhashmap.Map, string, entrySchema, map[string]bool)
```
MikaelSmith: Why pre-declare the variable when you create it on the next line?

plugin/entrySchema.go
MikaelSmith: I'm not sure where I'd see these naming differences in everyday use.

plugin/externalPluginRoot.go
ekinanp: That's the Go workaround of getting recursion to work w/ closures.

plugin/externalPluginEntry.go
ekinanp: Specific reason is because when I create the schema graph in `Init`, the root's ID is empty so we'll hit the cases in https://github.com/ekinanp/wash/blob/namespace-coreplugin-typeids/plugin/entrySchema.go#L272-L286. If this is on externalPluginEntry, then it will hit the default branch (externalPluginEntry is an embedded type in externalPluginRoot). Making it take the entry as a param. fixes that.

plugin/entrySchema.go
ekinanp: Hmm, I was originally thinking of extracting the resource type from the type ID, but I think I'm leaning towards adding a core-plugin specific builder to EntrySchema that lets plugin authors set the resource type. That way, if a plugin has multiple packages for its resources (e.g. in AWS, we have an EC2 package for EC2 stuff; S3 package for S3 stuff), the analytics side of things will remain unchanged (resource type makes things stable against a massive code-organization-related refactor).

That would make most of this PR unnecessary (except maybe pluginName), although I do still need it to properly implement the kind primary so it may as well go in.

api/fs/dir.go
ekinanp: Ok, think a separate "Remove the apifs package" issue would make sense?

MikaelSmith: Sure, mostly as tech-debt.

ekinanp:
Updated.

MikaelSmith (approved):

PR #441, opened 2019-08-06
Add analytics framework

### LINKED ISSUE #383: Analytics to track adoption
- How many people use it
- What plugins do they use
- What resource types do they use

Add Google Analytics for the website

Add product analytics to wash (use Bolt as an example):
- super easy to opt-out of
- checkin on startup with what subcommand is run

ekinanp:
Resolves https://github.com/puppetlabs/wash/issues/383

### DIFF
--- a/activity/core.go
+++ b/activity/core.go
@@ -107,6 +107,8 @@ func Warnf(ctx context.Context, msg string, a ...interface{}) {
 // SubmitMethodInvocation submits a method invocation event to Google Analytics.
 // It then records the invocation to the journal identified by the ID at `activity.JournalKey`
 // in the provided context.
+//
+// Note that entryType will typically be the type ID.
 func SubmitMethodInvocation(ctx context.Context, plugin string, entryType string, method string) {
 	// Note that we could move this over to the plugin package. Doing so would shorten
 	// the type-signature to (ctx, entry, method), but it would also require us to export
@@ -145,13 +147,13 @@ func SubmitMethodInvocation(ctx context.Context, plugin string, entryType string
 		recorder.mIMux.RUnlock()
 		return
 	}
+	log.Debugf("Submitting %v method invocation for entry instance %v", method, entryType)
 	err := analytics.GetClient(ctx).Event(
 		"Invocation",
 		"Method",
 		analytics.Params{
-			"Label":      method,
-			"Plugin":     plugin,
-			"Entry Type": entryType,
+			"Label":  method,
+			"Plugin": plugin,
 		},
 	)
 	if err != nil {
--- a/activity/core.go
+++ b/activity/core.go
@@ -18,6 +18,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/puppetlabs/wash/analytics"
 	"github.com/puppetlabs/wash/datastore"
 	log "github.com/sirupsen/logrus"
 )
@@ -30,7 +31,7 @@ const JournalKey KeyType = iota
 
 // Enforce a limit on cache size to avoid running out of file descriptors. It'll be rare that we
 // have dozens of processes running simultaneously.
-var journalFileCache = datastore.NewMemCache().WithEvicted(closeJournal).Limit(50)
+var recorderCache = datastore.NewMemCache().WithEvicted(closeRecorder).Limit(50)
 var journalDir = func() string {
 	cdir, err := os.UserCacheDir()
 	if err != nil {
@@ -43,7 +44,7 @@ var expires = 30 * time.Second
 // CloseAll ensures open journals are flushed to disk and closed.
 // Use when the application is shutting down.
 func CloseAll() {
-	journalFileCache.Flush()
+	recorderCache.Flush()
 }
 
 // Dir gets the directory where journals are stored.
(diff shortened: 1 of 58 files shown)


analytics/client.go
```diff
@@ -0,0 +1,318 @@
...
+func (c *client) baseParams() Params {
+	params := Params{
+		// v => Protocol Version
+		"v": "1",
+		// cid => Client ID
+		"cid": c.userID.String(),
+		// tid => Tracking ID
+		//
+		// TODO: Set this to the main Wash GA account's tracking ID.
+		"tid": "UA-144659575-2",
```
ekinanp: Will do this once everything's approved.

analytics/client.go
```diff
@@ -0,0 +1,318 @@
...
+		// an => Application Name
+		"an": "Wash",
+		// av => Application Version
+		//
+		// TODO: Fill this in with a proper value
+		"av": "1.0",
```
ekinanp: This is a bit hard because the version's set in `cmd.version`. Wondering if this might be worth a separate version package. Maybe properly setting the version could be a separate issue?

ekinanp:
It'll take about a day for the data to show up in custom reports, so should probably hold off on merging until there's something we can view. However, I am seeing the data in the Realtime monitor of GA.

Note that we'll also need to create a separate view to see the usage data reported here (and also turn off "filter analytics sent from spam bots").

fuse/dir.go
```diff
@@ -34,7 +34,7 @@ func (d *dir) children(ctx context.Context) (map[string]plugin.Entry, error) {
 
 	// Cache List requests. FUSE often lists the contents then immediately calls find on individual entries.
 	if plugin.ListAction().IsSupportedOn(updatedEntry) {
-		return plugin.CachedList(ctx, updatedEntry.(plugin.Parent))
+		return plugin.List(ctx, updatedEntry.(plugin.Parent))
```
ekinanp: It'd be nice if there's an easy way to detect when we're auto-completing vs. invoking something. That way, we don't report analytics for auto-complete (unless we want to).

analytics/client.go
MikaelSmith: Reporting application version is pretty useful. A separate package that we update makes sense.

fuse/dir.go
MikaelSmith: I'm not concerned about reporting for auto-complete.

plugin/entrySchema.go
```diff
@@ -172,6 +174,44 @@ func (s *EntrySchema) SetMetadataSchema(obj interface{}) *EntrySchema {
...
+// EntryType returns the entry's type
+func (s *EntrySchema) EntryType() string {
+	return s.entryType
+}
+
+// SetEntryType sets the entry's type. This should typically be the entry's class name.
+// Note that entry type is optional. It should only be set when you want analytics to
+// collect data on that entry.
+func (s *EntrySchema) SetEntryType(entryType string) *EntrySchema {
+	// We could try to get this information from the type ID; however, it is difficult
```
MikaelSmith: I don't feel like this is useful enough for the added complexity. I don't agree that we need to leave out simple directories; it's easy enough to filter them out explicitly, or we can omit entries that don't have useful attributes/metadata and don't implement Read/Exec. I don't think capturing whether a plugin's types are used is something the plugin author should need to think about.

I'm also not sure the schema should be considered fragile. It's user-facing, so we shouldn't refactor it on a whim, and we won't be capturing long-running data.

analytics/config.go
```diff
@@ -0,0 +1,96 @@
...
+		// This should never happen
+		err = fmt.Errorf("could not marshal the analytics config: %v", err)
+		return
+	}
+	err = ioutil.WriteFile(analyticsConfigFile, bytes, 0644)
+	return
+}
+
+func readAnalyticsConfigFile(path string) (Config, error) {
+	_, err := os.Stat(path)
```
MikaelSmith: I don't think calling Stat just before reading gives us any new information. ioutil.ReadFile should return the same error if it doesn't exist. So we should just call ReadFile.

api/server.go
```diff
@@ -108,6 +115,7 @@ func StartAPI(registry *plugin.Registry, mountpoint string, socketPath string) (
 
 	r := mux.NewRouter()
 
+	r.Handle("/analytics/screenview", screenviewHandler).Methods(http.MethodPost)
```
MikaelSmith: I find it kind of annoying that this shows up in the activity journal:
```
Aug  6 14:37:56.528 API: POST /analytics/screenview
Aug  6 14:37:56.532 API: POST /analytics/screenview complete
```

cmd/internal/server/core.go
```diff
@@ -129,6 +146,14 @@ func (s *Server) shutdown() {
 	// Close any open journals on shutdown to ensure remaining entries are flushed to disk.
 	activity.CloseAll()
 
+	// Flush any outstanding analytics hits
+	ticker := time.NewTicker(analytics.FlushDuration)
+	defer ticker.Stop()
+	go func() {
+		s.analyticsClient.Flush()
+	}()
+	<-ticker.C
```
MikaelSmith: This waits a full 250ms, right? We should stop the ticker if the goroutine completes before then.

cmd/root.go
```diff
@@ -33,6 +37,76 @@ func toRunE(main commandMain) runE {
...
+		}
+		// We use client.ForUNIXSocket instead of cmdutil.NewClient because the socketPath
+		// could change. The latter's possible if cmd.Name() == "wash", i.e. the root
+		// command.
+		_ = client.ForUNIXSocket(socketPath).Screenview(name, analytics.Params{})
+	}()
+}
+
+// GA => Google Analytics
+func waitForGARegistration() {
```
MikaelSmith: This seems like it should wait for the GA flush to complete, or FlushDuration, whichever comes first. Not the full FlushDuration.

cmd/rootMain.go
```diff
@@ -187,6 +187,7 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
 		cmdutil.ErrPrintf("Unable to start server: %v\n", err)
 		return exitCode{1}
 	}
+	registerInvocationToGA(cmd, socketpath)
```
MikaelSmith: Having these run over the socket seems weird, and results in an extra journal entry for starting the Wash shell.

MikaelSmith (changes requested):

cmd/internal/server/core.go
```diff
@@ -129,6 +146,14 @@ func (s *Server) shutdown() {
 	// Close any open journals on shutdown to ensure remaining entries are flushed to disk.
 	activity.CloseAll()
 
+	// Flush any outstanding analytics hits
+	ticker := time.NewTicker(analytics.FlushDuration)
+	defer ticker.Stop()
+	go func() {
+		s.analyticsClient.Flush()
```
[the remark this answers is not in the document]
ekinanp: Think it was to prevent Flush from blocking the server process from the rest of its cleanup (in case the network takes too long). I'll add a comment about that.

analytics/client.go
ekinanp: Should that be a separate issue or should I tack it on here? Think that may need some changes to homebrew-puppet.

analytics/config.go
```diff
@@ -0,0 +1,96 @@
...
+type Config struct {
+	Disabled bool      `yaml:"disabled"`
+	UserID   uuid.UUID `yaml:"user-id"`
+}
+
+// GetConfig returns Wash's analytics config, which is located at
+// $HOME/.puppetlabs/wash/analytics.yaml.
+func GetConfig() (config Config, err error) {
+	// Check if analytics is disabled via the environment before
+	// returning
```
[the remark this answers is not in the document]
ekinanp: The defer enforces precedence -- disabling analytics via environment has greater precedence than everything else. Otherwise, if we check environment first, then unmarshalling could overwrite our disabled value and the additional logic to check that could be annoying.

Will add a comment about this if you think it is OK.

cmd/rootMain.go
ekinanp: Ah yeah, wanted to maintain the symmetry with the other commands. I guess we could also say "Server start" => "Someone is using Wash", so that this stuff could move to the `internal/server` package.

cmd/root.go
ekinanp: Yeah, makes sense. I'll add that.

analytics/config.go
ekinanp: Oh yes, ReadFile should be good enough. Will update.

plugin/entrySchema.go
ekinanp: Open to alternative ways of getting the resource type, or omitting it altogether.

I do agree that knowing what resources are most popular doesn't really have any practical meaning aside from being something "cool to know." Plugin makes sense -- if AWS is popular, we should maintain AWS and maintain it well. That automatically includes its resources. So, I think I agree with you in that we should eliminate tracking the resource type altogether and just stick to method invocations (since the latter's simple enough to track).

api/server.go
ekinanp: I don't think it's that bad. But we could ignore "/analytics" endpoint pings if it is annoying.

analytics/client.go
MikaelSmith: I see it as part of this issue.

cmd/internal/server/core.go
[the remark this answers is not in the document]
MikaelSmith: Yeah, as a timer that makes sense.

analytics/config.go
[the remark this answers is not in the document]
MikaelSmith: Why not just return immediately if the environment variable is set? If it's set we don't need an ID and I don't see a reason to create the analytics file.

plugin/entrySchema.go
MikaelSmith: I'd be fine with that for now.

analytics/config.go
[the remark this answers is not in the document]
ekinanp: That also makes sense. Will update.

plugin/entrySchema.go
ekinanp: Ok. I think it's still useful for the journal to record the method invocations, so will record that stuff via typeID instead of entryType.

activity/core.go
```diff
@@ -107,6 +107,8 @@ func Warnf(ctx context.Context, msg string, a ...interface{}) {
 // SubmitMethodInvocation submits a method invocation event to Google Analytics.
 // It then records the invocation to the journal identified by the ID at `activity.JournalKey`
 // in the provided context.
+//
+// Note that entryType will typically be the type ID.
```
ekinanp: I want to keep entryType as-is here in case we do decide that reporting the entry type is useful.

ekinanp:
I've addressed all the feedback except for the version package. Working on that next.

MikaelSmith (commented):
Looks good, just waiting on adding version. .travis.yml and homebrew-puppet will need to be updated for the new path. Might make it worth adding a simple script to do release builds.

MikaelSmith (approved):

MikaelSmith:
Oh, just realized we need to add a mention of this to the installation docs.

ekinanp:
Yeah, was about to do that.

PR #447, opened 2019-08-09
Limit analytics from CI and local testing

MikaelSmith:
Disable analytics in Travis CI.

Skips sending analytics when Wash version is "unknown". This avoids sending analytics data when running tests or otherwise working from source.

### DIFF
--- a/analytics/client.go
+++ b/analytics/client.go
@@ -41,9 +41,16 @@ type Client interface {
 // NewClient returns a new Google Analytics client for submitting Wash analytics.
 func NewClient(config Config) Client {
 	if config.Disabled {
-		log.Debugf("Analytics opt-out is set, analytics will be disabled")
+		log.Debug("Analytics opt-out is set, analytics will be disabled")
 		return &noopClient{}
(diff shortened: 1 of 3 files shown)


analytics/client.go
```diff
@@ -41,9 +41,16 @@ type Client interface {
 // NewClient returns a new Google Analytics client for submitting Wash analytics.
 func NewClient(config Config) Client {
 	if config.Disabled {
-		log.Debugf("Analytics opt-out is set, analytics will be disabled")
+		log.Debug("Analytics opt-out is set, analytics will be disabled")
 		return &noopClient{}
 	}
+
+	// Don't submit analytics for untagged builds
+	if version.BuildVersion == "unknown" {
```
MikaelSmith: I thought about using a more restrictive check, like a regex on semver. But we can do the same thing in Google Analytics and it'll be relatively rare for people to apply non-semver tags. I just do it locally to remember which version I have installed.

ekinanp (approved):

PR #449, opened 2019-08-12
Add the kind primary

### LINKED ISSUE #404: Add a `kind` primary to wash find that filters on the kinds of entries
This primary only makes sense for plugins that support entry schemas.  The primary would give users a way to expressively filter on specific kinds of entries. This lets them answer higher-level questions like "How many Docker containers/EC2 instances/Kubernetes pods/S3 buckets do I have?". The current way of doing this is a bit magical. For example,
```
find -meta .Created -24h -o -meta .metadata.creationTimestamp -24h
```
returns all Docker containers + Kube pods that started recently. The reason it works is because Docker containers/Kube pods are the only entries whose metadata schemas satisfy the schema predicate, but figuring that out is a bit magical (and the query's fragile because there could be other Wash entries that could have similarly structured metadata schemas). The query also breaks when a user has external plugins that don't provide entry schemas b/c the meta primary, for a better UX, still recurses into those entries (so it'd generate a lot of API requests in the worst-case).

With the kind primary, the query would become
```
find \( -kind 'docker*container' -o -kind 'kube*pod' \) -crtime -24h
```
which is far more expressive. It is also optimized.

ekinanp:
Resolves https://github.com/puppetlabs/wash/issues/404

### DIFF
--- a/api/types/entrySchema.go
+++ b/api/types/entrySchema.go
@@ -10,12 +10,13 @@ import (
 	"github.com/puppetlabs/wash/plugin"
 )
 
-// EntrySchema describes an entry's schema, which
-// is what's returned by the /fs/schema endpoint.
+// EntrySchema describes an entry's schema, which is what's returned by
+// the /fs/schema endpoint.
 //
 // swagger:response
 type EntrySchema struct {
 	plugin.EntrySchema
+	path     string
 	typeID   string
 	children []*EntrySchema
 	// graph is an ordered map of `<TypeID>` => `<EntrySchema>`. We store it to make
@@ -38,7 +39,59 @@ func (s EntrySchema) MarshalJSON() ([]byte, error) {
 	return json.Marshal(s.EntrySchema)
 }
 
-// UnmarshalJSON unmarshals the entry's schema from JSON.
+// UnmarshalJSON unmarshals the entry's schema from JSON. EntrySchemas
+// are unmarshalled into their stree representation. Specifically, given
+// stree output like
+//
+// docker
+// ├── containers
+// │   └── [container]
+// │       ├── log
+// │       ├── metadata.json
+// │       └── fs
+// │           ├── [dir]
+// │           │   ├── [dir]
+// │           │   └── [file]
+// │           └── [file]
+// └── volumes
+//     └── [volume]
+//         ├── [dir]
+//         │   ├── [dir]
+//         │   └── [file]
+//         └── [file]
+//
+// The unmarshalled graph will have nodes "containers", "container", "log",
+// "metadata.json", "fs", "dir", "file", "file" to mirror the first half
+// of the tree (the second "file" node represents "fs"' "file" child while
+// the first "file" node represents "dir"'s "file" child). The second half
+// of the tree is captured by nodes "volumes", "volume", "dir", "file", "file"
+// (here, the second "file" node represents "volume"'s "file" child while
+// the first "file" node represents "dir"'s "file" child). The stree's root
+// is the "docker" node.
+//
+// Unmarshalling the graph this way makes things symmetric with what a user
+// sees via stree. It also simplifies the implementation of some `wash find`
+// primaries.
+//
+// NOTE: One way to think about this representation is as follows. Consider a
+// node A. Let R-B-A and R-C-D-A be two possible paths from the stree root R to
+// A. Then the unmarshalled graph will have two different "A" nodes -- one for
+// the path "R-B-A", and another for the path "R-C-D-A". Thus, the unmarshalled
+// graph represents all possible paths from the stree root R (the starting entry)
+// to its descendants (the other nodes). This is the precise definition of an
+// entry's stree.
+//
+// NOTE: The type ID uniquely identifies a specific class of entries. The path
+// identifies a specific kind of entries. For example in the above stree output,
+// docker/containers/container/fs/dir and docker/volumes/volume/dir both share a
+// common "volumeDir" class. However, the former represents entries that are
+// directories in a Docker container's enumerated filesystem while the latter
+// represents entries that are directories inside a Docker volume.
+//
+// NOTE: The above "path" => "kind" analogy is not always correct. For example,
+// "docker/containers/container/fs/file" and "docker/containers/container/fs/dir/file"
+// represent the same kind of entry (a file inside a Docker container). However,
+// the analogy is good enough for most cases.
 func (s *EntrySchema) UnmarshalJSON(data []byte) error {
 	rawGraph := linkedhashmap.New()
 	if err := rawGraph.FromJSON(data); err != nil {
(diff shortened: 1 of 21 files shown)


api/types/entrySchema.go
```diff
@@ -47,49 +100,80 @@ func (s *EntrySchema) UnmarshalJSON(data []byte) error {
...
+	(*s) = (*root)
+	s.graph = graph
 
 	return nil
 }
 
-// TypeID returns the entry's type ID. This should be unique.
+// Path returns the unique path to this specific entry's schema
+// (relative to the stree root). The path consists of
+//    <root_label>/<parent1_label>/.../<label>
```
ekinanp: This format makes the kind primary a bit annoying to use from the mountpoint. For example, if I want to find only Docker containers, I'll need to do something like `find -k "mountpoint/docker*container"` (I can't do `find -k "*docker*container"` because that will return containers in the `mock_docker` plugin).

I've been thinking about some alternatives:
1. Change `root_label` => `.` (which would indicate that the path is relative to the root). Then the previous example becomes something like `find -k "./docker*container"`. Problem with this is if one has multiple paths, e.g. `find <path1> <path2>`, it makes it impossible to filter on the kind of `<path1>`, `<path2>`.

2. Add some munging to the glob where if it starts with `.`, we will consider that the kind of `<path1>`, `<path2>`, depending on which paths' being walked. So, `./docker*container` would be munged to `mountpoint/docker*container`.

3. Change the mountpoint label to `.`, but keep everything else as-is. Think this makes the most sense, since the mountpoint label isn't being used anywhere.

MikaelSmith:
Hmm, I wonder why Travis hasn't started yet.

cmd/internal/find/primary/kind.go
```diff
@@ -0,0 +1,70 @@
...
+	negated bool
+}
+
+func (p kindPredicate) Negate() predicate.Predicate {
+	return kindP(p.g, !p.negated)
+}
+
+// TODO: Fill this description out
+const kindDetailedDescription = `
+(-k|-kind) glob
```
ekinanp: I'll flesh this out once https://github.com/puppetlabs/wash/pull/449#discussion_r313008477 is addressed.

api/types/entrySchema.go
```diff
@@ -38,7 +39,59 @@ func (s EntrySchema) MarshalJSON() ([]byte, error) {
 	return json.Marshal(s.EntrySchema)
 }
 
-// UnmarshalJSON unmarshals the entry's schema from JSON.
+// UnmarshalJSON unmarshals the entry's schema from JSON. EntrySchemas
```
ekinanp: Hmm, I think this could move to the EntrySchema type declaration b/c [fs/schema's response is a different type](https://github.com/puppetlabs/wash/blob/master/api/schema.go#L16).

ekinanp: Regardless, I consider the refactor here a bug-fix (even though the previous unmarshalling didn't cause any issues in Wash's normal operation).

api/types/entrySchema.go
MikaelSmith: What about treating an absolute path as the start of the schema path? So `/docker*container`? Not sure the `.` is necessary to clarify that description.

cmd/internal/find/types/testdata/triangle.json
```diff
@@ -1,15 +1,18 @@
 {
     "A": {
+        "label": "A",
```
MikaelSmith: Is there an example where `label` isn't redundant?

ekinanp: Yes, see https://github.com/ekinanp/wash/blob/kind-primary-round2/cmd/internal/find/types/testdata/real_world_aws.json

api/types/entrySchema.go
ekinanp: Yeah, could also do that too. Would that resolve the issue in (1)?

MikaelSmith: It's not clear to me how any of them enable that. Maybe I'm getting confused by what `mountpoint` is. I assumed that would normally be whatever we mount Wash in, e.g. `mnt28139`.

Maybe the `/` is unnecessary because you're expecting a full match. For the initial example, why doesn't `docker/*container` match all (and only) containers? For example (1), could you give some concrete examples?

ekinanp: `mountpoint` is plugin.Registry's label.

In the initial example, the schema paths of entries in the `docker` and the `mock_docker` plugin will start with `mountpoint/docker` and `mountpoint/mock_docker`, respectively (assuming we're invoking `find` in the mountpoint -- the mountpoint's entry is plugin.Registry). Since users should not care about the mountpoint label, they'd need to do something like `*docker*container` to match Docker containers -- `*docker` matches both `docker` and `mock_docker`.

The `<path1>`, `<path2>` stuff is a bit theoretical. Here's one example,
```
find <ec2_instance> <s3_bucket> -k 'instance' -o -k 'bucket'
```

Here you're filtering on entries of kind EC2 Instance or kind Bucket, so you'd print out `<ec2_instance>` and `<s3_bucket>`. The schema paths of `<ec2_instance>` and `<s3_bucket>` is (currently) `instance` and `bucket`, which is why this works. Munging the schema paths to `./` or `/` would lose their kinds, so the above expression would be incorrect.

Even though the above example is theoretical, I'd still like to get it right because there's no reason why you shouldn't be able to do that (all other predicates make it possible for you to print `<ec2_instance>` and `<s3_bucket>`, for example). 

I'm starting to like Option 3 more.

ekinanp: Note that you can also do
```
find docker -k '*container'
```

to filter out Docker containers. We could also document that in the kind primary's documentation.

MikaelSmith: Ok, so the problem is you're using glob somewhat unconventionally. In Linux, the `*` doesn't normally match separators and you have to provide a full path. Somewhere the `**` pattern was added to match directory trees. You're not passing any separators to the glob library, so it's matching them with a simple `*`.

My inclination would be to include separators because I think it avoids some common ambiguities. So your examples would be (omitting out the mountpoint label because we shouldn't use it)
- `docker/**/container`
- `**/instance`

The way you're configuring the glob calls, those examples have to be
- `/docker/*/container` to avoid ambiguities such as a plugin named `mock_docker` or `docker_new`, and new things like `dead_container` (if you're trying to be that specific). In practice maybe `/docker/*container` is fine.
- I have no idea how you're making `instance` work, because the glob library still needs a full match. I don't think the code you have here supports it, so you'd need to do something like `*instance` or `*/instance`.

The details about what the mountpoint should be called seem unimportant because glob works on full paths, and we should just exclude the mountpoint completely.

MikaelSmith: Ok, I'm caught up. `find -path` is a pattern that treats separators as normal characters, so it doesn't behave the same as shell globbing. That's the behavior we're basing this new primary on.

ekinanp:
Ok, this is ready for review.

MikaelSmith (approved):

PR #453, opened 2019-08-15
Update examples to use the kind primary

ekinanp:

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -365,7 +365,7 @@ EOF</pre></td></tr></table>
 
 <p>Wash also includes support for AWS. If you have your own and you&rsquo;ve configured the AWS CLI on your workstation, you&rsquo;ll be able to use Wash to explore EC2 instances and S3 buckets.</p>
 
-<p>As an example, you might want to periodically check how many EC2 instances are running (and display that via <a href="https://getbitbar.com/">BitBar</a>):</p>
+<p>As an example, you might want to periodically check how many execable instances are running in the AWS plugin (and display that via <a href="https://getbitbar.com/">BitBar</a>):</p>
 <div class="highlight"><div style="background-color:#fff;-moz-tab-size:4;-o-tab-size:4;tab-size:4">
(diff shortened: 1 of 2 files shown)


ekinanp:
I wonder if we should update `ctime`'s description to mean creation time (and link the corresponding issue to let people know that we intend to fix it). Saying that b/c right now, ctime => creationTime and it may be useful to some people.

docs/index.html
```diff
@@ -373,8 +373,8 @@ <h3 id="listing-aws-resources">Listing AWS resources</h3>
 </span><span style="margin-right:0.4em;padding:0 0.4em 0 0.4em;color:#7f7f7f">3
 </span></pre></td>
 <td style="vertical-align:top;padding:0;margin:0;border:0;">
-<pre style="background-color:#fff;-moz-tab-size:4;-o-tab-size:4;tab-size:4">running=`find aws -action exec -meta .State.Name running 2&gt;/dev/null | wc -l | xargs`
-total=`find aws -action exec -meta .State.Name -exists 2&gt;/dev/null | wc -l | xargs`
+<pre style="background-color:#fff;-moz-tab-size:4;-o-tab-size:4;tab-size:4">running=`find aws -k &#39;*ec2*instance&#39; -meta .State.Name running 2&gt;/dev/null | wc -l | xargs`
```
MikaelSmith: I'm not sure I'd change this one. If we had other things like ECS instances it might still work, where-as your change restricts it to ec2. The s3 bucket example below seems like enough.

ekinanp: We should rephrase the example to be "all execable instances" (or compute if we're sticking to "compute" => "execable"). The current phrasing is better expressed via the kind primary.

MikaelSmith: Ok.

ekinanp: Hmm, that might be tough to do with the metadata schema (e.g. ECS instances could have different schemas than EC2 instances). Could we stick to EC2 instances instead? Also OK w/ rephrasing the example as "find all execable instances in the AWS plugin".

MikaelSmith: Let's rephrase the example. I haven't checked ECS instances, but I suspect AWS is fairly consistent with `.State.Name`.

ekinanp: Updated.

MikaelSmith (approved):

PR #460, opened 2019-08-16
Capture OS and architecture of wash execution

MikaelSmith:
Combines these as a single custom dimension because we're limited on how many we can add, and in most cases we view this is a specific build target.

### DIFF
--- a/analytics/client.go
+++ b/analytics/client.go
@@ -41,16 +41,9 @@ type Client interface {
 // NewClient returns a new Google Analytics client for submitting Wash analytics.
 func NewClient(config Config) Client {
 	if config.Disabled {
-		log.Debug("Analytics opt-out is set, analytics will be disabled")
+		log.Debugf("Analytics opt-out is set, analytics will be disabled")
 		return &noopClient{}
 	}
-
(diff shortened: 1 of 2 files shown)


ekinanp (approved):

ekinanp:
Think https://github.com/puppetlabs/wash/blob/master/analytics/client_test.go#L258 also needs to be updated.

analytics/client.go
```diff
@@ -41,16 +41,9 @@ type Client interface {
 // NewClient returns a new Google Analytics client for submitting Wash analytics.
 func NewClient(config Config) Client {
 	if config.Disabled {
-		log.Debug("Analytics opt-out is set, analytics will be disabled")
+		log.Debugf("Analytics opt-out is set, analytics will be disabled")
 		return &noopClient{}
 	}
-
-	// Don't submit analytics for untagged builds
-	if version.BuildVersion == "unknown" {
```
MikaelSmith: @grimradical suggested reverting this for folks running from source. I haven't figured out how to automatically set a version though.

ekinanp (approved):

PR #464, opened 2019-08-19
Include more kind primary examples

ekinanp:


cmd/internal/find/primary/meta.go
```diff
@@ -83,6 +83,14 @@ with a specified label. In general, the meta primary can be used to filter
 on any property that's specified in the entry's meta attribute. See the
 EXAMPLES section for some interesting real-world examples.
 
+NOTE: The meta primary should be used in conjunction with the kind primary.
+The latter's useful for explicitly specifying the kind of entries you're
+filtering on, which enables find to take full advantage of entry schema
+optimizations. This is especially useful when the entry's schema doesn't
+include metadata schemas (possible for external plugins). The kind primary's
```
ekinanp: This is nice. Means Python/Ruby plugin authors can still give their users a way to optimize metadata queries.

MikaelSmith (approved):

PR #468, opened 2019-08-20
Add crtime attribute

### LINKED ISSUE #405: Add a proper creation time (crtime) attribute
We misunderstood the `ctime` attribute to mean `creation time` when it means `changed time`. We should still keep `ctime` around because it's a filesystem attribute. The work in this issue is to add a separate `crtime` attribute for the creation time, which mirrors some FSes like Ext 3/4 and NTFS, and to clean-up the core plugins to ensure things that once set `ctime` to creation time now set `crtime` to creation time.

The work here would also include adding a `crtime` primary to Wash find so users can filter on it.

ekinanp:
Resolves https://github.com/puppetlabs/wash/issues/405


plugin/aws/s3Object.go
```diff
@@ -39,8 +39,9 @@ func newS3Object(o *s3Client.Object, name string, bucket string, key string, cli
 	mtime := awsSDK.TimeValue(o.LastModified)
 	s3Obj.
 		Attributes().
-		SetCtime(mtime).
+		SetCrtime(mtime).
```
MikaelSmith: This seems like it should be something else. Does S3 return anything besides LastModified (which definitely corresponds to mtime)?

plugin/aws/s3Bucket.go
```diff
@@ -146,23 +146,24 @@ func listObjects(ctx context.Context, client *s3Client.S3, bucket string, prefix
...
+	bucket.crtime = crtime
 	bucket.client = s3Client.New(session)
 	bucket.session = session
 	bucket.
 		Attributes().
-		SetCtime(bucket.ctime).
-		SetMtime(bucket.ctime).
-		SetAtime(bucket.ctime)
+		SetCrtime(bucket.crtime).
+		SetMtime(bucket.crtime).
```
MikaelSmith: Should we bother setting the rest? We don't have any specific information.

MikaelSmith (commented):
This looks good. Trying to double-check some things around ctime/mtime/atime.

plugin/aws/s3Object.go
ekinanp: Unfortunately no -- see https://github.com/puppetlabs/wash/blob/master/plugin/aws/s3Object.go#L32-L35

MikaelSmith (approved):

plugin/aws/s3Bucket.go
ekinanp: Don't know. It's a pattern we've been using across multiple things which is why I replicated it here. Personally I prefer we use the information we have, so the answer would be "No." What about you?

PR #471, opened 2019-08-21
Rework the website introduction to be more problem focused

### LINKED ISSUE #466: Rework website introduction to be more problem-focused
We feel like we've learned some about this from working on blog post material. Update the website introduction section based on that.

ekinanp:
Resolves https://github.com/puppetlabs/wash/issues/466

### DIFF
--- a/website/content/_index.md
+++ b/website/content/_index.md
@@ -5,7 +5,7 @@ description = ""
 draft= false
 +++
 
-Wash helps you deal with all your remote or cloud-native infrastructure using the UNIX-y patterns and tools you already know and love.
+Wash helps you manage your remote infrastructure using well-established UNIX-y patterns and tools to free you from having to remember multiple ways of doing the same thing.
 
 <div id="horizontalmenu">
     • <a href="#introduction">introduction</a>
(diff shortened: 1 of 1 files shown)


website/content/_index.md
```diff
@@ -5,7 +5,7 @@ description = ""
 draft= false
 +++
 
-Wash helps you deal with all your remote or cloud-native infrastructure using the UNIX-y patterns and tools you already know and love.
+Wash helps you manage a bunch of different things using well-established UNIX-y patterns and tools to free you from having to learn multiple ways of doing the same thing.
```
ekinanp: I think this phrasing's OK. Open to better ways of stating this. My main goal here is to make it familiar for newer people (which is why I removed the "you already know and love" bit) while also highlighting the main problem Wash is trying to solve (to free you from having to learn multiple ways of doing the same thing).

website/content/_index.md
```diff
@@ -16,22 +16,86 @@ Wash helps you deal with all your remote or cloud-native infrastructure using th
...
 </div>
 
-<script id="asciicast-mX8Mwa75rr1bJePLi3OnIOkJK" src="https://asciinema.org/a/mX8Mwa75rr1bJePLi3OnIOkJK.js" async></script>
-
 ## Introduction
 
-Exploring, understanding, and inspecting modern infrastructure should be simple and straightforward. Whether it's containers, VMs, network devices, IoT stuff, or anything in between...they all have different ways of enumerating what you have, getting a stream of output, running commands, etc. Every vendor has its own tools and APIs that expose these features, each one different, each one bespoke. Thus, they are difficult to compose together to solve higher-level problems. And that's no fun at all!
-
-[UNIX's philosophy](https://en.wikipedia.org/wiki/Unix_philosophy#Origin) and abstractions have worked for decades. They're pretty good, and more importantly, they're _familiar_ to millions of people. Wash intends to apply those same philosophies and abstractions to modern, distributed infrastructure.
+Have you ever had to:
```
ekinanp: The intro seems long mostly because of the tables. Otherwise, there's not a lot of text that people have to read.

website/content/_index.md
```diff
@@ -16,22 +16,86 @@ Wash helps you deal with all your remote or cloud-native infrastructure using th
...
 
-Wash aims to:
-
-* help you keep track of what systems you're running
-* make scripting across your new-fangled infrastructure as easy as writing a local shell script
-* provide easy access to files, logs, metadata, and any other activity for viewing, editing, and UNIXy slicing-and-dicing
-* help you build better tools that work across a variety of cloud APIs
-* let you easily build your own integrations in whatever language you want
-* be extremely simple to get up-and-running; if it takes you more than a few minutes, let us know!
+<script id="asciicast-mX8Mwa75rr1bJePLi3OnIOkJK" src="https://asciinema.org/a/mX8Mwa75rr1bJePLi3OnIOkJK.js" async></script>
```
ekinanp: Going to spend tomorrow thinking about some ideas on how to improve the screencast. Main thing is to show some of the listed invocations of Table 2 in action.

I think with this intro, we can get away with a slightly longer screencast (to showcase some of the cooler features of Wash, and to also spend time explaining some of its concepts).

website/content/_index.md
grimradical: Re #1, I think we need to be more specific than "a bunch of different things". The thing I like about remote / cloud native is that it describes the actual plugins we've got currently.

I'd also change "from having to learn" (learning is a good thing, i figure!) to something else...maybe "from having to remember"? "having to deal with"?

I think we should also add a clause (separate sentence) around letting folks easily build higher-level tools, instead of wrestling with multiple, lower-level ones.

website/content/_index.md
```diff
@@ -16,22 +16,86 @@ Wash helps you deal with all your remote or cloud-native infrastructure using th
...
+
+* You no longer have to learn different commands to execute a task across different things. All you need is one command (`find` for List/Find; `cat` for Read; and `wexec` for Exec).
+
+* You no longer have to install a bunch of different tools. All you need to install is the Wash binary.
+
+* You no longer have to learn different DSLs for filtering stuff. All you need to learn is find's expression syntax and its individual primaries. Once you do that, you can filter on almost any conceivable property of your specific thing.
+
+And this is only scratching the surface of Wash's capabilities. Checkout the screencast below to see some more (and to see Wash in action):
 
-Wash aims to:
```
MikaelSmith: I feel like some of this is still useful and missing from the introduction above.

MikaelSmith: Primarily about building better tools across a variety of cloud APIs and being easy to extend.

ekinanp: When I tried to precisely define what we mean by "building better tools across a variety of cloud APIs," I ended up reducing the problem down to "centralizing a bunch of basic and fundamental stuff" because when you have that down, the "better tools" come naturally. The tables in the intro already hint at the latter, so I don't know if stating the "better tools" stuff is necessary. I also think that's something we should show people via a collage of interesting problems that are extremely easy to do with Wash (e.g. like navigating stuff, tailing a bunch of logs, getting SSHFS-like functionality for free on a bunch of different things, finding/filtering things, etc.). I'm thinking of adding some more of those to the screencast, but I also think a separate section of "Here are some cool, non-trivial sysadmin problems that are really easy to do with Wash" would be nice (we could tack those on as blog posts).

I think making a note of Wash being easy to extend is a good idea. Where do you think that should go?

ekinanp: Theme is instead of telling people "You can build better stuff with Wash!", we should show them what building stuff with Wash would look like. The product's good enough that the "Oh my god, I really can build better tools with this" insight will likely come naturally to the user.

MikaelSmith: Extending can probably be a note at the end.

ekinanp: Per discussion on Slack, think it's better to leave "Extending Wash" to the features section. The intro contains enough information already that adding an additional "NOTE: You can extend Wash via ..." is not worth imposing a context-switch of "Ok, from centralizing commands => external plugins" on the reader.

MikaelSmith (approved):

PR #474, opened 2019-08-22
Move "wash by example" to a tutorial page

### LINKED ISSUE #409: Move "Wash by example" to tutorial page, replace with short examples
Move the "Wash by example" section of the site to a new tutorial page that we link from the main page. Replace it with a set of short examples of doing different things with Wash that omit the setup and messy details, so it's easy to see some of the things it can do at a glance.

MikaelSmith:
It's a lot of description for the front page. Most of what it showed off is also highlighted in the new introduction, so I don't see a need to add an examples section to the front page.

Resolves #409.

### DIFF
--- a/website/config.toml
+++ b/website/config.toml
@@ -33,6 +33,10 @@ pygmentsStyle = "vs"
     name = "news"
     weight = 1
   [[menu.main]]
+    url = "tutorial"
+    name = "tutorial"
+    weight = 2
+  [[menu.main]]
     url = "docs"
     name = "docs"
     weight = 4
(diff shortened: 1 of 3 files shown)


MikaelSmith:
Specific `tail -f` stuff is the only thing not currently covered. Any suggestions on where that should go?

ekinanp (approved):

ekinanp:
Tutorials might be a good place for the screencasts (or their transcripts). Think `tail -f` could be something we worry about later.

Using the docker-compose example would be good for the intro screencast when navigating through the Docker plugin.

PR #490, opened 2019-08-28
Ensure Wash aliases are set if zshrc sets similar aliases

### LINKED ISSUE #489: zsh prezto overrode 'find' alias
My zshrc includes https://github.com/sorin-ionescu/prezto. It appears to specify an alias for `find`, and because it's loaded after zshenv it overrides Wash's aliases.

Wash sets aliases in zshenv because they may be used in a script. It should ensure they're also set after loading zshrc to preserve the loading intent:
- load default user config if there's no custom washenv/washrc
- set wash config
- allow override with washenv/washrc

MikaelSmith:
Fixes #489.


ekinanp (approved):
Have some questions, but they're not blockers.

ekinanp (approved):

Issue #504, opened 2019-09-03
Allow SSH to instances using private IP address


angrychimp:
## Use Case
I would like to be able to use wash to interact with compute instances (specifically on AWS) using the private IP address. I have an established VPN connection to my VPC and none of my instances have public IPs or hostnames.

## Describe the Solution You Would Like
I would like to see a configuration option that allows me to override the default behavior of attempting to SSH to an instance using the public IP or hostname. When this configuration option is set, it will _only_ attempt to connect to the instance via the configured metadata value (whether PrivateIpAddress, PublicIpAddress, etc.)

## Describe Alternatives You've Considered
I'm unable to assign public IPs or hostnames to my instances, so I have no other alternative at present.

MikaelSmith:
This seems pretty straight-forward. https://github.com/puppetlabs/wash/blob/0.9.0/plugin/aws/ec2Instance.go#L262-L269 handles selecting what connection information to use. If there's no public IP address/DNS hostname we could fall back to private IP address.

Making it configurable to only use a specific metadata seems useful. That's a little more effort to add, as it needs to be passed down from plugin initialization at https://github.com/puppetlabs/wash/blob/0.9.0/plugin/aws/root.go#L64 to `profile` -> `resourcesDir` -> `ec2Dir` -> `ec2InstancesDir` -> `ec2Instance`. This suggests to me it'd be helpful to have a way to transparently pass config to other places in the plugin (something also requested in #349. In this case attaching config to every context we pass to plugin operations might be simple and cover most cases.

I'm going to be out on parental leave the next couple weeks, so may not get to it immediately. Do you have any interest in trying to add either the private IP fallback, or a config option?

PR #506, opened 2019-09-10
PrivateIpAddress as failover target for SSH connections to AWS EC2 instance

angrychimp:
This adds `PrivateIpAddress` as a last option for `hostname` when attempting to connect to an EC2 instance. If PrivateIpAddress is not available (which admittedly, it almost always is), then the `Exec` method will still throw an Error. 

This addresses one potential solution to the problem identified in issue #504

### DIFF
--- a/plugin/aws/ec2Instance.go
+++ b/plugin/aws/ec2Instance.go
@@ -264,8 +264,10 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
 		hostname = name.(string)
 	} else if ipaddr, ok := meta["PublicIpAddress"]; ok && ipaddr != nil {
 		hostname = ipaddr.(string)
+	} else if ipaddr, ok := meta["PrivateIpAddress"]; ok && ipaddr != nil {
+		hostname = ipaddr.(string)
 	} else {
-		return nil, fmt.Errorf("No public interface found for %v", inst)
+		return nil, fmt.Errorf("No available interface found for %v", inst)
 	}
 
 	var identityfile string


ekinanp (approved):
Thanks for the PR! This makes sense based on the discussion in https://github.com/puppetlabs/wash/issues/504. Still would like to wait for @MikaelSmith to review it once he's back on Sept. 23rd. Best to leave the CI failures until then too since that's a fix on our end.

MikaelSmith:
New version of golangci-lint?

MikaelSmith (approved):
I think this is reasonable for now. I'd like to be able to quickly error if that ip address won't be reachable, but not sure how to do that yet. Maybe only try if we're on the same subnet?

Anyway, I think this is useful enough to add. We should continue to look into improving it.

angrychimp:
> Maybe only try if we're on the same subnet?

This might fail depending on the subnet mask. For example, my instances are on a 172.0.200.0/23 subnet while my VPN NAT and clients are on a 172.0.10.0/24 subnet. Connections work due to the routing and security group rules.

But I agree with the concept of failing on a connection issue - not sure if there's a simple way to quickly check for a service listener on the target port with a short timeout. You don't want to have to wait for 30 secs while a SSH connect attempt times out, but then again maybe that's just what happens? If you figure out the "make config options available to nested resources" thing then perhaps you can specify a SSH connection timeout override in `wash.yaml`.

But now we're drifting back towards the original issue and away from targeted discussion of this PR :)

plugin/aws/ec2Instance.go
```diff
@@ -264,8 +264,10 @@ func (inst *ec2Instance) Exec(ctx context.Context, cmd string, args []string, op
 		hostname = name.(string)
 	} else if ipaddr, ok := meta["PublicIpAddress"]; ok && ipaddr != nil {
 		hostname = ipaddr.(string)
+	} else if ipaddr, ok := meta["PrivateIpAddress"]; ok && ipaddr != nil {
+		hostname = ipaddr.(string)
```
MikaelSmith: I think the one thing I want to add is an `activity.Record` message to help identify why the connection might be failing. I can add that when I get this merged.

MikaelSmith:
Thanks for the PR!

plugin/aws/ec2Instance.go
MikaelSmith: https://github.com/puppetlabs/wash/pull/512

PR #508, opened 2019-09-24
Omit extra newline when printing winfo and meta output

ekinanp:


MikaelSmith:
Is that because the EntryMap includes a newline when marshalled? Also we still need to fix CI.

ekinanp:
Oh yeah, probably. I'll look into fixing it there.

MikaelSmith (approved):

MikaelSmith:
I think the marshal behavior is fine.

PR #515, opened 2019-10-01
Feature/kubernetes contexts in list

### LINKED ISSUE #129: Dynamically load the Kubernetes contexts so that if they get updated, then Wash will reload them.
Currently contexts are loaded from config in the Init function. Move that to the plugin root's List function (where they're used) so it can be cached and reloaded as needed.

pop:
fixes #129 

Hey @MikaelSmith here's some code changes! If you have any feedback please ping me, or feel free to make changes on my branch.

I neglected to add automated tests, but I outlined the two manual tests I ran below. You mentioned something about adding tests in #129, so if you still think we can add some automated tests to this I'd love to work on those.

## Manual (minikube) Testing

I'm sure you know how to test this, but here's how I went about it.

TLDR create a minikube clusters, run wash, rename the Kubernetes context, verify that wash changes contexts accordingly.

1. Create default minikube cluster.
2. In `~/.kube/config` you have a `cluster`, `context`, and `user` named `minikube`.
3. Run wash
```
$ wash
wash . > ls kubernetes
minikube
```
4. Rename the `minikube` context to `notminikube`. This used to break in wash.
```
$ kubectl config rename-context minikube notminikube
wash . > ls kubernetes
notminikube
```
5. Cleanup: delete the `minikube` cluster and delete the `notminikube` context.

You can also verify this change by creating two Kubernetes clusters (`cluster1` and `cluster2`) and switching contexts with `kubectl use-context cluster2` while using wash.


pop:
Should I do this signoff thing?

.gitignore
```diff
@@ -0,0 +1 @@
+wash
```
pop: I almost `git commit`ed the wash binary. Woops!

plugin/kubernetes/root.go
```diff
@@ -21,7 +21,6 @@ import (
 // Root of the Kubernetes plugin
 type Root struct {
 	plugin.EntryBase
-	contexts []plugin.Entry
```
pop: This is the one change I'm most nervous about. I tried to keep it around, but it wasn't really necessary so any attempts seemed weird and made the code needlessly confusing.

pop (commented):
Pretty much just lifted the logic in `Init` and moved it to `List`. Fairly straight forward. Definitely open to feedback and changes.

plugin/kubernetes/root.go
MikaelSmith: I agree we don't need it anymore. Thanks for the PR!

MikaelSmith:
This looks great, thanks!

MikaelSmith (approved):

PR #516, opened 2019-10-02
Change munge.parseMode => munge.ToUintMode

ekinanp:
This makes mode serialization more flexible for external plugin authors
while also preserving the correctness of the individual bits.

### DIFF
--- a/plugin/entryAttributes.go
+++ b/plugin/entryAttributes.go
@@ -285,13 +285,11 @@ func (a *EntryAttributes) UnmarshalJSON(data []byte) error {
 		a.SetCrtime(t)
 	}
 	if mode, ok := mp["mode"]; ok {
-		// Even though os.FileModes are uint32 types, json.Unmarshal unmarshals them as float64.
-		// That's ok, because float64 has sufficient precision to represent all uint32 types.
-		if raw, ok := mode.(float64); ok {
-			a.SetMode(os.FileMode(raw))
-		} else {
-			return attrMungeError("mode", fmt.Errorf("mode was unexpected type %T: %v", mode, mode))
+		md, err := munge.ToUintMode(mode)
+		if err != nil {
+			return attrMungeError("mode", err)
 		}
+		a.SetMode(os.FileMode(md))
 	}
 	if size, ok := mp["size"]; ok {
 		sz, err := munge.ToSize(size)
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,7 @@ require (
 	github.com/gobwas/glob v0.2.3
 	github.com/gogo/protobuf v1.2.1 // indirect
 	github.com/golang-collections/collections v0.0.0-20130729185459-604e922904d3
+	github.com/google/btree v1.0.0 // indirect
 	github.com/google/uuid v1.1.1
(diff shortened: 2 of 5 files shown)


plugin/entryAttributes.go
```diff
@@ -285,13 +285,11 @@ func (a *EntryAttributes) UnmarshalJSON(data []byte) error {
 		a.SetCrtime(t)
 	}
 	if mode, ok := mp["mode"]; ok {
-		// Even though os.FileModes are uint32 types, json.Unmarshal unmarshals them as float64.
-		// That's ok, because float64 has sufficient precision to represent all uint32 types.
-		if raw, ok := mode.(float64); ok {
-			a.SetMode(os.FileMode(raw))
-		} else {
-			return attrMungeError("mode", fmt.Errorf("mode was unexpected type %T: %v", mode, mode))
+		md, err := munge.ToUintMode(mode)
```
ekinanp: @MikaelSmith I think this should still preserve what you intended in https://github.com/puppetlabs/wash/commit/c7a7d43262fb2d98aecf4144f4e62a81234e48f8

ekinanp: Or just change `munge.ToFileMode` to return `os.FileMode(<whatever ToUintMode returns>)`

ekinanp:
Looks like there might be another issue with the mode that I found while debugging `local_fs`. Specifically, I expect that `stat -f '%Dp' local_fs/Desktop` should match `stat -f '%Dp' $HOME/Desktop`. But here's what I actually get:

```
wash . ❯ stat -f '%Dp' local_fs/Desktop
33216
```
```
wash . ❯ stat -f '%Dp' $HOME/Desktop
16832
```

We should probably file a separate issue for this discrepancy.

plugin/entryAttributes.go
MikaelSmith: The bit representation of os.FileMode is not the same as that returned by stat. So this is incorrect. See the commit message for https://github.com/puppetlabs/wash/commit/c7a7d43262fb2d98aecf4144f4e62a81234e48f8#diff-db372dbae47f5fb2b6db0b72a78758c4.

MikaelSmith (changes requested):
Please revert the changes to entryAttributes. They munge two different things.

plugin/entryAttributes.go
ekinanp: I don't understand the issue. The previous code took a `float64` mode and passed in to `os.FileMode` as-is (and the reason makes sense). The current code uses `ToUintMode` to make sure that we also allow hexadecimal/octal modes.

MikaelSmith: Ah, I must've been thinking about ToFileMode.

MikaelSmith (approved):

MikaelSmith:
That discrepancy looks weird on your local_fs looks strange. I don't get problems looking at a Linux VM in GCP. So this might be surprising behavior with the stat call we use on macOS specifically.

ekinanp:
Hmm, OK. mode isn't extremely important for the external plugin tutorial, so I'll just make a note that's along the lines of what you just said in the footnote

PR #522, opened 2019-10-06
Feature: Kubernetes pods now have containers.

### LINKED ISSUE #228: Update Kubernetes pods to support multiple containers
Add support for multiple containers in a pod. Our current implementation uses a shortcut available only if the pod has a single container; for multiple containers it errors.

Make the pod a directory that lists its containers, with the containers doing all the work the pod currently does (for exec, metadata, etc).

Interacts with #203.

pop:
* Kubernetes pods are now represented as directories holding containers.
* Individual containers within a pod can be logged and executed on.

@MikaelSmith You pretty much called it. I created a `container` struct and put most of the logic from 'pod.go` into `container.go`.

To test this I ran `wexec` on the containers within a pod to exercise `Exec`, and ran `cat` to exercise `Open` and `tail -f` to exercise `Stream` on the container files. I used the following Kubernetes manifest file to spawn the Pods:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: moon
spec:
  containers:
  - image: alpine:latest
    name: face
    command: ["/bin/sh", "-c", "apk -U add curl && curl https://raw.githubusercontent.com/puppetlabs/wash/master/plugin/kubernetes/pod.go && sleep 10000"]
  - image: alpine:latest
    name: dark-side
    command: ["/bin/sh", "-c", "apk -U add curl && curl https://raw.githubusercontent.com/puppetlabs/wash/master/plugin/kubernetes/podsDir.go"]

---

apiVersion: v1
kind: Pod
metadata:
  name: sun
spec:
  containers:
  - image: alpine:latest
    name: shine 
    command: ["/bin/sh", "-c", "apk -U add curl && curl https://raw.githubusercontent.com/puppetlabs/wash/master/README.md && sleep 12345"]
```

Let me know if anything should be tweaked or implemented differently. I'm glad to make changes.

Since there weren't any automated tests in that part of the project I didn't add any. I'm feel like Kubernetes has to have some mocking framework we can use for stuff like this...

Fixes #228

### DIFF
--- a/plugin/kubernetes/container.go
+++ b/plugin/kubernetes/container.go
@@ -1,0 +1,150 @@
+package kubernetes
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/pkg/errors"
+	"github.com/puppetlabs/wash/activity"
+	"github.com/puppetlabs/wash/plugin"
+	corev1 "k8s.io/api/core/v1"
+	k8s "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
+	k8exec "k8s.io/client-go/util/exec"
+)
+
+type container struct {
+	plugin.EntryBase
+	client *k8s.Clientset
+	config *rest.Config
+	ns	 string
+	pod *corev1.Pod
+}
+
+func newContainer(ctx context.Context, client *k8s.Clientset, config *rest.Config, ns string, c *corev1.Container, p *corev1.Pod) (*container, error) {
+	cntnr := &container{
+		EntryBase: plugin.NewEntry(c.Name),
+	}
+	cntnr.client = client
+	cntnr.config = config
+	cntnr.ns = ns
+	cntnr.pod = p
+
+	// TODO: Get container startup time from ContainerStatus from the parent Pod
+	cntnr.
+		Attributes().
+		SetMeta(plugin.ToJSONObject(c))
+
+	return cntnr, nil
+}
+
+func (c *container) Schema() *plugin.EntrySchema {
+	return plugin.
+		NewEntrySchema(c, "container").
+		SetMetaAttributeSchema(corev1.Container{})
+}
+
+func (c *container) Open(ctx context.Context) (plugin.SizedReader, error) {
+	logOptions := corev1.PodLogOptions{
+		Container: c.Name(),
+	}
+	req := c.client.CoreV1().Pods(c.ns).GetLogs(c.pod.Name, &logOptions)
+	rdr, err := req.Stream()
+	if err != nil {
+		return nil, err
+	}
+	var buf bytes.Buffer
+	var n int64
+	if n, err = buf.ReadFrom(rdr); err != nil {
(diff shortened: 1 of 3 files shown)


pop:
One thing I didn't implement, but I totally could, is re-adding `Exec` to `pods`.

MikaelSmith:
Thanks for taking this on! I think it makes sense to keep things uniform right now over convenient. So I'd skip adding `Exec` to `pods` because it would only work for pods with a single container, and it's not much harder to address the individual container.

I'll get to reviewing the rest of this later today.

plugin/kubernetes/container.go
```diff
@@ -0,0 +1,150 @@
...
+func newContainer(ctx context.Context, client *k8s.Clientset, config *rest.Config, ns string, c *corev1.Container, p *corev1.Pod) (*container, error) {
+	cntnr := &container{
+		EntryBase: plugin.NewEntry(c.Name),
+	}
+	cntnr.client = client
+	cntnr.config = config
+	cntnr.ns = ns
+	cntnr.pod = p
+
+	// TODO: Get container startup time from ContainerStatus from the parent Pod
```
MikaelSmith: How easy is it to add this?

MikaelSmith (changes requested):
I'd like to resolve the TODO. Otherwise looks like it was pretty straight-forward.

MikaelSmith:
For testing, we'll most likely need to create an interface for the parts of the kubernetes API we want and pass in a mock version of the API client.

plugin/kubernetes/container.go
pop: The metadata is located somewhere in the depths of the pod. I think `Pod` has a `PodStatus` which has a set of `ContainerStatus` which include these times. If the metadata isn't necessary we can defer it for later. I just noticed that some other objects had creation time.

pop: Fixed.

plugin/kubernetes/container.go
```diff
@@ -33,9 +34,20 @@ func newContainer(ctx context.Context, client *k8s.Clientset, config *rest.Confi
 	cntnr.ns = ns
 	cntnr.pod = p
 
-	// TODO: Get container startup time from ContainerStatus from the parent Pod
+	// Find when the container was started; set this as the creation time
+	creationTimestamp := time.Now()
```
MikaelSmith: It's ok to leave creationTimestamp unset if it's not in ContainerStatuses. So I'd only update them if you find a match.

You could put
```
cntnr.Attributes().SetAtime(ecs.State.Running.StartedAt.Time)...
```
inside the if block, then break.

ekinanp (approved):
I'll need to update Exercise 6 in https://puppetlabs.github.io/wash/tutorials/01_fundamentals/plugins-actions-entries once this is released. Code LGTM though.

plugin/kubernetes/container.go
pop: Good catch. I'll get to that tonight.

pop: Fixed.

MikaelSmith (approved):

PR #528, opened 2019-10-08
Fix unit test that assumed Wash binary existed

MikaelSmith:
Unit tests shouldn't assume that a `wash` binary exists. Switch to
verifying that a shell script exists where we expect it.


MikaelSmith:
I didn't come up with a better way to test the functionality. I'm open to ideas.

ekinanp (approved):

PR #537, opened 2019-10-09
Load plugins in parallel

MikaelSmith:
Starting Wash slows down as you add more external plugins. Some of that
is unavoidable because we're doing more work. Make it faster by loading
all plugins concurrently.

### DIFF
--- a/cmd/internal/server/core.go
+++ b/cmd/internal/server/core.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"runtime/pprof"
+	"sync"
 	"time"
 
 	"github.com/Benchkram/errz"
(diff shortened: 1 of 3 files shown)


cmd/internal/server/core.go
```diff
@@ -203,16 +204,24 @@ func (s *Server) Stop() {
...
 	for name, root := range s.plugins {
 		log.Infof("Loading %v", name)
-		if err := registry.RegisterPlugin(root, s.opts.PluginConfig[name]); err != nil {
-			// %+v is a convention used by some errors to print additional context such as a stack trace
-			log.Warnf("%v failed to load: %+v", name, err)
-			anyFailed = true
-		}
+		wg.Add(1)
+		go func(name string, root plugin.Root) {
+			if err := registry.RegisterPlugin(root, s.opts.PluginConfig[name]); err != nil {
```
ekinanp: Do we need a lock in https://github.com/puppetlabs/wash/blob/master/plugin/registry.go#L38 ?

MikaelSmith: Yes, we probably do. Good catch.

plugin/registry.go
```diff
@@ -36,36 +38,40 @@ var pluginNameRegex = regexp.MustCompile("^[0-9a-zA-Z_-]+$")
 // RegisterPlugin initializes the given plugin and adds it to the registry if
 // initialization was successful.
 func (r *Registry) RegisterPlugin(root Root, config map[string]interface{}) error {
-	registerPlugin := func() {
+	registerPlugin := func(checkName bool) {
+		r.mux.Lock()
+		if checkName {
```
ekinanp: Think `checkName` => `initSucceeded` -- latter provides more context about what's going on here. It took me a bit to understand why you did this.

ekinanp (approved):

Issue #284, opened 2019-05-23
Improve 'wash ls'


MikaelSmith:
We should make it something you want to use in-place of 'ls' because it provides useful context.

grimradical:
Output ideas:

We could still default a long listing, but use a modeline of sorts to display which protocols are supported by each line item:

```
# wash ls 
d--- 23 May 18:02 foo
-rs- 23 May 18:02 logfile
d--x 23 May 18:02 ssh.mydomain.com
```

Where the modeline refers to:

```
<list?><read?><stream?><exec?> <ctime> <name>
```

This would probably look a lot better with color, a la how [exa](https://the.exa.website/) does it.

ekinanp:
Think we should also omit `.` unless the user wants to list hidden directories. That makes it a bit more consistent with the built-in `ls`.

ekinanp:
I'm starting to think we should keep its default output symmetric with `ls`, and make the current info we show optional (maybe match it with whatever option `ls` has for that behavior?). This would let us directly alias `list` => `ls`, and it also means I can use our `list` in the screencast instead of `ls` (I think the current state of our `list` is a bit too verbose for a tutorial).

PR #541, opened 2019-10-14
Update external plugin tutorial w/ tech writer edits

ekinanp:

### DIFF
--- a/docs/_tutorials/03_extending_wash/writing-an-external-plugin.md
+++ b/docs/_tutorials/03_extending_wash/writing-an-external-plugin.md
@@ -1,11 +1,11 @@
 ---
 title: Writing an external plugin
 ---
-In this tutorial, we will write the `local_fs` plugin. `local_fs` lets you navigate your local filesystem. Although it isn’t practically useful, it illustrates the key concepts behind external plugin development.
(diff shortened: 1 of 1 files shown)


docs/_tutorials/03_extending_wash/writing-an-external-plugin.md
```diff
@@ -94,7 +94,7 @@ wash . ❯ ls
 aws        docker     gcp        kubernetes local_fs
 ```
 
-If you try to `ls local_fs`, you should get an error
+We're not done yet, so if you try to `ls local_fs`, you'll get this error:
 
 ```
 wash . ❯ ls local_fs
```
MikaelSmith: The repetition in the output is pretty annoying. `list` doesn't do that. I'm not sure why `ls` seems to retry a bunch before failing.

MikaelSmith (approved):

docs/_tutorials/03_extending_wash/writing-an-external-plugin.md
ekinanp: Yeah, I'm not sure either. Should fix that once https://github.com/puppetlabs/wash/issues/284 is done.

PR #542, opened 2019-10-14
Omit restricted resource types in GCP and AWS

### LINKED ISSUE #534: GCP plugin creates invalid `compute` directory when compute API is unavailable
## Description

The GCP plugin will create a `compute` directory even if the GCP Compute Engine API is disabled. If one attempts to list the `compute` directory an error will be emitted:

```
wash gcp/s2-backups ❯ ls compute
WARN FUSE: List /gcp/<project>/compute errored: googleapi: Error 403: Access Not Configured. Compute Engine API has not been used in project <project> before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/compute.googleapis.com/overview?project=<project> then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry., accessNotConfigured
cannot access 'compute': Input/output error (os error 5) 
```

## Expected Behavior
The `gcp/compute` directory should not be created when the Compute Engine API is not enabled.

## Steps to Reproduce
1. Create a new GCP project with `gcloud projects create ...`
2. Attempt to enter the compute directory with wash

## Environment
 - Wash bb77569
 - Ubuntu 18.04

MikaelSmith:
GCP and AWS access controls can limit what types of APIs a user has
access to. A common one is to limit users to certain subsets of the API,
such as only Compute or Storage. When that happens, any operations on
the representative directory will fail.

Create a new Inaccessible attribute that entries set when they are
inaccessible due to lack of permissions. Inaccessible entries will be
hidden from list and find commands, with the failure reason logged to
the relevant activity journal.

In practice, you would see
```
$ cd gcp/Storage-Only-Project
$ ls
storage
$ stree
.
├── compute
│   └── ...
└── storage
└── ...
$ whistory
...
7   2019-10-14 16:13  ls
$ whistory 7
Oct 14 16:13:12.752 FUSE: List /gcp/Storage-Only-Project
Oct 14 16:13:13.654 Omitting Storage-Only-Project/compute: googleapi:
Error 403: user@example.com does not have compute.list access to p...
```

Fixes #534.

### DIFF
--- a/plugin/cache.go
+++ b/plugin/cache.go
@@ -27,9 +27,9 @@ func InitCache() {
 	}
 }
 
-// SetTestCache sets the cache to the provided mock. It can only be called
-// by the tests
-func SetTestCache(c datastore.Cache) {
+// SetTestCache sets the cache to the provided mock. It can only be called by the tests.
+// Returns a context that includes a parent ID so later cache operations will succeed.
+func SetTestCache(c datastore.Cache) context.Context {
 	if notRunningTests() {
 		panic("SetTestCache can only be called when running the tests")
 	}
@@ -39,6 +39,7 @@ func SetTestCache(c datastore.Cache) {
 	}
 
 	cache = c
+	return context.WithValue(context.Background(), parentID, "/")
 }
 
 // UnsetTestCache unsets the test cache. It can only be called
@@ -172,6 +173,11 @@ func cachedList(ctx context.Context, p Parent) (map[string]Entry, error) {
 					SecondChildSlashReplacer: entry.slashReplacer(),
 					CName:                    cname,
 				}
+			}
+
+			if entry.isInaccessible() {
+				// Skip entries that are expected to be inaccessible.
+				continue
 			}
 			searchedEntries[cname] = entry
 
--- a/plugin/entryBase.go
+++ b/plugin/entryBase.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"flag"
 	"time"
+
+	"github.com/puppetlabs/wash/activity"
 )
 
 type defaultOpCode int8
@@ -43,6 +45,7 @@ type EntryBase struct {
 	ttl             [3]time.Duration
 	wrappedTypesMap SchemaMap
 	prefetched      bool
+	inaccessible    bool
 }
 
 // NewEntry creates a new entry
@@ -131,6 +134,17 @@ func (e *EntryBase) Attributes() *EntryAttributes {
 func (e *EntryBase) SetAttributes(attr EntryAttributes) *EntryBase {
 	e.attr = attr
 	return e
+}
+
+func (e *EntryBase) isInaccessible() bool {
+	return e.inaccessible
+}
+
+// MarkInaccessible sets the inaccessible attribute and logs a message about why the entry is
+// inaccessible.
+func (e *EntryBase) MarkInaccessible(ctx context.Context, err error) {
+	activity.Record(ctx, "Omitting %v: %v", e.id(), err)
+	e.inaccessible = true
 }
 
 // Prefetched marks the entry as a prefetched entry. A prefetched entry
--- a/plugin/types.go
+++ b/plugin/types.go
@@ -52,6 +52,7 @@ type Entry interface {
 	wrappedTypes() map[interface{}]*JSONSchema
 	setWrappedTypes(map[interface{}]*JSONSchema)
 	isPrefetched() bool
+	isInaccessible() bool
 }
 
 /*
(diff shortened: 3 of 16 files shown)


MikaelSmith:
@adrienthebo mind kicking the tires on this?

MikaelSmith:
I think there are other types it should probably cover, but this seems like one of the major issues.

ekinanp:
I think the more general issue here is "Does the user have access to this entry?". That could be captured by a `Restricted`/`Inaccessible` property on the entry, which generalizes well to external plugins because we can make that an (optional) key in the entry JSON (`plugin.CachedList` would have additional code to omit restricted entries). I think we should go with that for this PR, because I don't see `SelectListable` generalizing to external plugins.

Long-term, the more general issue is "What are 'Wash' permissions?" Entry access is one such permission while something like "what are the allowable supported actions" could be another (with each action having its own set of permissions; action-specific permissions could be useful for the upcoming Signal action b/c then we can enforce API-specific "signal" permissions [e.g. you can start/stop a VM but can't hibernate it]).

adrienthebo:
LGTM for the specific fix:

```
wash gcp ❯ cd s2-backups                                                                                                                                                                                                                                                    ✔  17:03:23 
wash gcp/s2-backups ❯ ls
dr-xr-x--- adrien adrien 4 KB Tue Oct 15 17:03:15 2019   storage                                                                                                                                                                                                           ✔  17:03:24 
wash gcp/s2-backups ❯
```

@ekinanp raises a good point, though one that might be worth discussing in detail in the larger picture - could one use file permissions to indicate access?

But for this specific issue, this solved the issue that I discovered and I'm happy with the fix.

MikaelSmith:
Working on what changing to an Inaccessible attribute might look like.

MikaelSmith:
I've reworked the implementation. Entries now define a new function to call for a permission check. This moves most of the actual checking internally, so we can enforce some standard interactions.

plugin/types.go
```diff
@@ -193,3 +193,11 @@ type Readable interface {
 	Entry
 	Open(context.Context) (SizedReader, error)
 }
+
+// Permissible is used to define how to check that adequate permissions to access the resource are
```
MikaelSmith: To make this work with external plugins we'd need a Permitted method to check. Obviously that could be statically declared when listing the items.

cmd/docs.go
```diff
@@ -37,5 +37,13 @@ func docsMain(cmd *cobra.Command, args []string) exitCode {
 	} else {
 		cmdutil.Println("No description provided.")
 	}
+
+	if schema.PermissibleChildren {
+		// Intentionally adds an empty newline after the provided description.
+		cmdutil.Println(`
+If a subdirectory is listed in 'stree' but not visible in your directory then you are likely
```
ekinanp: I think we should move this over to `stree` and just add a note somewhere in its output. Or move it to the website.

MikaelSmith: So rather than track this, we'll look at including a note about it in documentation on the `stree` command and the website.

plugin/types.go
MikaelSmith: An attribute will be simpler if we're getting rid of the need to test whether an entry is Permissible (or making it an explicit thing a plugin needs to set).

plugin/cache.go
```diff
@@ -189,7 +190,21 @@ func cachedList(ctx context.Context, p Parent) (map[string]Entry, error) {
 		return nil, err
 	}
 
-	return cachedEntries.(map[string]Entry), nil
+	// Return a copy of the map so we don't mutate the cached version. Future calls to list should
```
MikaelSmith: If we're going to use an attribute on the entry to determine whether we have permission to operate on it, deciding whether or not to include that entry in a list of entries can be moved closer to where it's displayed to the user.

MikaelSmith:
Updated with the latest take on this using an attribute.

plugin/entryAttributes.go
```diff
@@ -179,6 +180,15 @@ func (a *EntryAttributes) SetSize(size uint64) *EntryAttributes {
 	return a
 }
 
+// Inaccessible returns true if accessing an entry is expected to fail due to missing permissions.
+func (a *EntryAttributes) Inaccessible() bool {
```
ekinanp: I think this should move over to the `EntryBase` class. Main reason is b/c the `inaccessible` attribute isn't something that you're going to filter on.

MikaelSmith: Making it an attribute made it straight-forward to set from an external plugin.

MikaelSmith: I'll look into how the external plugin should work.

MikaelSmith: Added external plugin support.

plugin/entryBase.go
```diff
@@ -133,6 +135,13 @@ func (e *EntryBase) SetAttributes(attr EntryAttributes) *EntryBase {
 	return e
 }
 
+// MarkInaccessible sets the inaccessible attribute and logs a message about why the entry is
+// inaccessible.
+func (e *EntryBase) MarkInaccessible(ctx context.Context, err error) {
```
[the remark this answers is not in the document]
MikaelSmith: Done.

ekinanp (approved):

PR #545, opened 2019-10-15
Make Wash ls more symmetric with the built-in ls

ekinanp:
The commits contain all the details.

### DIFF
--- a/cmd/ls.go
+++ b/cmd/ls.go
@@ -7,23 +7,18 @@ import (
 	"github.com/spf13/cobra"
 
 	apitypes "github.com/puppetlabs/wash/api/types"
-	"github.com/puppetlabs/wash/cmd/internal/config"
 	cmdutil "github.com/puppetlabs/wash/cmd/util"
 	"github.com/puppetlabs/wash/plugin"
 )
 
 func lsCommand() *cobra.Command {
-	aliases := []string{}
-	if !config.Embedded {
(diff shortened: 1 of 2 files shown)


ekinanp:
Here's what the new `ls` looks like:

```
wash . ❯ ls docker/containers/wash_tutorial_redis_1
fs/
log
metadata.json
wash . ❯ ls -l docker/containers/wash_tutorial_redis_1
NAME            MODIFIED    ACTIONS
fs/             <unknown>   list
log             <unknown>   read, stream
metadata.json   <unknown>   read
wash . ❯ ls docker/containers/wash_tutorial_redis_1/metadata.json
metadata.json
```

ekinanp:
I'm not yet sure how to handle "hidden" entries. I considered defining them as entries whose cname begins with a `.`, but a definition like that's a bit too early so I left it out for now.

MikaelSmith:
Does that mean "hidden" files will always be shown?

MikaelSmith (changes requested):
This looks good. Can you also update the website to reflect the changes to `ls`? In particular I'm thinking about how it prints errors.

ekinanp:
For now, yeah. Hidden files will also be included.

ekinanp:
And ah yeah, I forgot about the website. Will do.

MikaelSmith (approved):

PR #551, opened 2019-10-18
Create method wrappers for each entry method

### LINKED ISSUE #550: Create suitable wrappers to each entry method
There's some inconsistency in how entry methods are being invoked. For example, instead of invoking `p.List`, one must invoke `plugin.CachedList` because List's results are cached, and `CachedList` also does some extra stuff like cname validation, and setting entry IDs. Similarly, to invoke `r.Open`, one must invoke `plugin.CachedOpen` because we cache Open's result. However to invoke `Exec`, one can invoke `e.Exec` because we don't do anything special with an Exec invocation. Similar situations hold with `Stream`.

This asymmetry is annoying, especially now that we are adding `Delete`. That's because `Delete` will require a `plugin.Delete` wrapper that does additional stuff like clear the entry's cache, and deleting the entry from the parent's CachedList map (there's no good reason to keep this in the API since "deleting an entry" is a general plugin thing).

Thus, it is best if we just tell people to invoke `plugin.<Method>` whenever they want to invoke an entry's method. This matches what we already do with `plugin.Name`, `plugin.CName`, `plugin.Schema`, etc. It's OK if `plugin.<Method>` calls `e.<Method>` directly -- the resulting consistency makes that worthwhile. At a higher-level, `plugin.<Method>` enforces the following separation of concerns:

* What a plugin author needs to care about (implementing `e.<Method>`)
* What the Wash server needs to care about (ensuring that `e.<Method>` will do the right thing, where doing the right thing could require additional work like cname validation for plugin.CachedList or clearing an entry's cache [work that plugin authors don't need to care about]).

The work in this issue is to create `plugin.<Method>` wrappers for _all_ invoked Wash methods. We should rename the wrappers in `analyticsWrappers.go` to `plugin.<Method>WithAnalytics`. So, `plugin.List` => `plugin.ListWithAnalytics`.

ekinanp:
Resolves https://github.com/puppetlabs/wash/issues/550


plugin/methodWrappers.go
```diff
@@ -111,3 +111,13 @@ func Schema(e Entry) (*EntrySchema, error) {
 func List(ctx context.Context, p Parent) (map[string]Entry, error) {
 	return cachedList(ctx, p)
 }
+
+// Open reads the entry's content. Note that Open's results could be cached. Thus, when
+// using the reader returned by this method, use idempotent read operations such as ReadAt
+// or wrap it in a SectionReader. Using Read operations on the cached reader will change it
+// and make subsequent uses of the cached reader invalid.
+//
+// TODO: Could we change this to Read? E.g. plugin.Read.
```
ekinanp: I think this would make more sense. Keeps it symmetric with the actions. But this is more of a nit.

plugin/methodWrappers.go
```diff
@@ -0,0 +1,139 @@
...
+
+// Exec execs the command on the given entry.
+func Exec(ctx context.Context, e Execable, cmd string, args []string, opts ExecOptions) (ExecCommand, error) {
+	return e.Exec(ctx, cmd, args, opts)
+}
+
+// Stream streams the entry's content for updates.
+func Stream(ctx context.Context, s Streamable) (io.ReadCloser, error) {
+	return s.Stream(ctx)
+}
```
ekinanp: I like this better because now I can add `Delete` and centralize the cache-related cleanup involved when deleting an entry.

plugin/methodWrappers.go
```diff
@@ -102,3 +103,11 @@ func IsPrefetched(e Entry) bool {
 func Schema(e Entry) (*EntrySchema, error) {
 	return schema(e)
 }
+
+// List lists the parent's children. It returns a map of <entry_cname> => <entry_object>
+// to optimize querying a specific entry.
+//
+// Note that List's results could be cached.
```
ekinanp: "could be" is meant to account for the possibility that someone disabled caching. Open to better ways of wording this because it reads a bit weird.

api/metadata.go
```diff
@@ -37,7 +37,7 @@ var metadataHandler handler = func(w http.ResponseWriter, r *http.Request) *erro
 		return errResp
 	}
 
-	metadata, err := plugin.CachedMetadata(ctx, entry)
+	metadata, err := plugin.Metadata(ctx, entry)
```
MikaelSmith: We don't have analytics collection on Metadata?

plugin/methodWrappers.go
MikaelSmith: The only thing that felt weird to me is that we return a reader that you also call Read on. But The symmetry is compelling, so I wouldn't mind changing it.

MikaelSmith (approved):

api/metadata.go
ekinanp: Nope. Should we? Seems fundamental enough that I don't know if it is necessary.

MikaelSmith: Probably not.

PR #553, opened 2019-10-18
Remove include-meta option from winfo

### LINKED ISSUE #552: Remove include-meta option from winfo
People can always get the meta attribute via `meta --attribtue`.

ekinanp:
People can get this via meta --attribute, so no need to have two
different ways of getting the same thing.

Resolves https://github.com/puppetlabs/wash/issues/552


MikaelSmith:
This seems consistent with the rationale put forth in https://github.com/puppetlabs/wash/pull/485.

Issue #479, opened 2019-08-26
Delete resources


MikaelSmith:
It's common to delete cloud resources when they're no longer used. We believe Wash provides some help around finding resources you would want to delete, and it should be easy to then delete them.

Examples of resources you can delete:
- VMs
- anything in storage (prefix/directory, object/file, bucket)
- project, namespace

Add a new Wash primitive that plugin entries can implement to allow deleting a resource:
```
type Deletable interface {
  Delete(context.Context) error
}
```

Plugins that implement Delete on an entry should ensure it and all its children are removed. It may error, such as if other dependencies need to be deleted first.

Things that should implement Delete
- EC2 and GCE VMs, K8s pods, Docker containers
- S3 and Google Storage objects and prefixes (as folders)
- The Volume package, so you can delete files and directories on target filesystems.
- The external plugin interface should support delete via a command like: "<script> delete <path> ..."

The API should respond to `DEL /fs?path=<path>` by deleting the resource at that path.

Also create our own `delete` that
- Deletes entries.
- Should prompt before deleting each thing it will explicitly delete, with '-f' as an option to skip confirmation.
- Accepts multiple paths to delete.

See https://docs.google.com/document/d/1NnJ1Qxeu7Sh4_ymOpo0enibKb3V5tzsSsX3H-N0PPDY/edit?usp=sharing for further discussion on the concept.

MikaelSmith:
Updated to reflect discussion that `-r` seems like an option that arbitrarily makes it more complex with no obvious benefit. Things you would think of as an item you delete - such as a VM - required `rm -r`. Also switched the name to `delete` as we're not copying many of the semantics of `rm`.

PR #554, opened 2019-10-20
Add the delete action

ekinanp:
Subsequent PRs will implement Delete for the core plugins.

### DIFF
--- a/plugin/entryMap_test.go
+++ b/plugin/entryMap_test.go
@@ -1,7 +1,10 @@
 package plugin
 
 import (
+	"strconv"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/suite"
 )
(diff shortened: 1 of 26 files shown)


ekinanp:
I tested this with the `local_fs` plugin; its delete implementation delegated to `rm -rf`. Things seem to be working OK.

cmd/rm.go
```diff
@@ -0,0 +1,69 @@
...
+	rmCmd := &cobra.Command{
+		Use:     use + " <path> [<path>]",
+		Aliases: aliases,
+		Short:   "Deletes the entries at the specified paths",
+		Long: `Deletes the entries at the specified paths, prompting the user for confirmation
+before deleting each entry.`,
+		Args: cobra.MinimumNArgs(1),
+		RunE: toRunE(rmMain),
+	}
+	rmCmd.Flags().BoolP("recurse", "r", false, "Delete directories (parent entries)")
```
ekinanp: Hmm, given the asymmetry between the `fs/delete` endpoint and the `Delete` method's signature (something that isn't an issue in our other actions), I'm starting to think that maybe the `recurse` option isn't such a good idea. Is there a compelling reason for us to enforce it? It seems simple enough to just say "`wrm` removes the given entries" (or, equally, that "rm"'s semantics don't cleanly apply to Wash, which is a good reason to change `wrm` => `delete` so that people don't get confused).

(If we talked about not having `-r` on Wednesday's meeting and I spoke against it, then apologies for that).

plugin/types.go
```diff
@@ -193,3 +193,12 @@ type Readable interface {
...
 	Open(context.Context) (SizedReader, error)
 }
+
+// Deletable is an entry that can be deleted. Entries that implement Delete
+// should ensure that it and all its children are removed. If the entry has
+// any dependencies that need to be deleted, then Delete should return an
+// error.
+type Deletable interface {
+	Entry
+	Delete(context.Context) error
```
ekinanp: Deleting an EC2 instance is equivalent to terminating it. According to https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/terminating-instances.html, termination is irreversible. However, terminated EC2 instances do show up in the AWS console for an hour.

Reason I bring this up is b/c if we delete an EC2 instance in Wash, that EC2 instance will disappear from the `instances` directory (breaking consistency with the AWS console). Is that OK? If so, we should probably filter out terminated EC2 instances when listing the `instances` directory.

ekinanp: I lean towards not implementing `Delete` for EC2 instances and maybe outlining the reason in `docs <ec2_instance>`.

MikaelSmith: I feel like this might be a common behavior. Deleting things may not be instantaneous, and we need to come up with some way to reflect that they're being removed.

MikaelSmith: I think EC2 should respond to Delete, same as GCP. In each case, it looks like they may go through a TERMINATED state before being removed. I think for now, we should consider TERMINATED as absent and not list those instances. The Delete implementation for them should only return once they've arrived at the TERMINATED state.

The API should trigger Delete on batched instances in parallel so that waiting on multiple to go through that lifecycle doesn't take too long.

cmd/rm.go
MikaelSmith: Updated the issue description, I think it's clear to remove `-r` and rename the command.

plugin/methodWrappers_test.go
```diff
@@ -67,19 +86,80 @@ func newmethodWrappersTestsMockEntry(name string) *methodWrappersTestsMockEntry
...
 }
 
+func (suite *MethodWrappersTestSuite) TestDelete_ReturnsDeleteError() {
+	ctx := context.Background()
+	e := newMethodWrappersTestsMockEntry("foo")
+
+	expectedErr := fmt.Errorf("an error")
+	e.On("Delete", ctx).Return(expectedErr)
+
+	var entry Entry = e
```
MikaelSmith: Why the cast to `Entry`?

MikaelSmith (changes requested):

plugin/entryMap_test.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+		count++
+		return true
+	})
+	// <= 1 is to account for the fact that Go map's iteration order
+	// is random.
+	suite.True(count <= 1)
+}
+
+func (suite *EntryMapTestSuite) TestConcurrentReadWrite() {
```
ekinanp: This test does panic if any of Load/Delete/Range fail to acquire a lock or release an acquired lock. It also panics if Delete acquires a read lock instead of a write lock.

ekinanp:
Updated to address the PR feedback.

MikaelSmith (approved):

MikaelSmith:
Initial work for https://github.com/puppetlabs/wash/issues/479.

PR #557, opened 2019-10-23
Update Delete to handle long-running deletions

ekinanp:


docs/docs/external-plugins.md
```diff
@@ -208,6 +209,19 @@ bash-3.2$ /path/to/myplugin.rb schema /myplugin/foo ''
 }
 ```
 
+## delete
+`<plugin_script> delete <path> <state>`
+
+When `delete` is invoked, the script must output a boolean JSON. `true` means that the entry was deleted. `false` means that the entry is marked for deletion and will eventually be deleted.
+
+**Note:** If you anticipate `delete` taking a long time (> 30 seconds), then output `false`.
```
ekinanp: I wonder if we can use a similar pattern for `signal`.

plugin/methodWrappers.go
```diff
@@ -139,18 +139,26 @@ func Stream(ctx context.Context, s Streamable) (io.ReadCloser, error) {
...
-func Delete(ctx context.Context, d Deletable) error {
-	if err := d.Delete(ctx); err != nil {
-		return err
+func Delete(ctx context.Context, d Deletable) (deleted bool, err error) {
+	deleted, err = d.Delete(ctx)
+	if err != nil {
+		return
 	}
+	if !deleted {
+		// The entry will eventually be deleted so leave the cache alone.
```
ekinanp: Should we refresh the entry here? Thinking a `parent.List` call, though the latter could add up when deleting multiple entries from the same parent.

I lean towards leaving it alone until it becomes an issue.

MikaelSmith: Instead of explicitly refreshing, I'd clear the cache entry. That delays the List call until it's actually needed, and raises the likelihood that the entry has been removed.

plugin/types.go
```diff
@@ -195,11 +195,8 @@ type Readable interface {
 	Open(context.Context) (SizedReader, error)
 }
 
-// Deletable is an entry that can be deleted. Entries that implement Delete
-// should ensure that it and all its children are removed. If the entry has
-// any dependencies that need to be deleted, then Delete should return an
-// error.
+// Deletable is an entry that can be deleted.
```
MikaelSmith: Why'd you remove the description? We also need to describe what the boolean return value represents.

plugin/methodWrappers.go
ekinanp: Ah yeah, that is better. Thanks, will update.

MikaelSmith (changes requested):

plugin/types.go
ekinanp: I was trying to place the plugin-author specific action info in the external plugin docs and leave the comments here to core-plugin specific subtleties. I guess it wouldn't hurt to include a brief description.

plugin/types.go
```diff
@@ -195,7 +195,8 @@ type Readable interface {
 	Open(context.Context) (SizedReader, error)
 }
 
-// Deletable is an entry that can be deleted.
+// Deletable is an entry that can be deleted. Its semantics are described
+// in https://puppetlabs.github.io/wash/docs/external-plugins#delete
```
ekinanp: I think people should be able to figure out what the return values mean from the external plugin docs.

ekinanp:
Updated to handle cache refreshing and also made some docs updates.

plugin/types.go
MikaelSmith: Someone - including our future selves - writing core plugins might not think to go look there.

MikaelSmith: The core plugin documentation should largely stand on its own.

plugin/cache.go
```diff
@@ -74,6 +74,22 @@ func opKeysRegex(path string) *regexp.Regexp {
 	return regexp.MustCompile(expr)
 }
 
+// opKeyRegex returns a regex that matches the key <op>::<path>.
+// It is useful for clearing a specific cached op for the entry
+// corresponding to <path>.
+func opKeyRegex(op string, path string) *regexp.Regexp {
```
MikaelSmith: What changed here?

plugin/types.go
MikaelSmith: Ah, I guess you link to it. But it feels weird to me for the core plugin documentation to link out to an external page just to explain what the function does and returns.

plugin/cache.go
ekinanp: I want to clear out `List::<parent>` only in `plugin.Delete`. `opKeysRegex` clears out `Op::<parent>` and `Op::<children>`.

ekinanp:
Ok, updated again to address the feedback.

MikaelSmith (approved):

PR #559, opened 2019-10-24
Implement delete for all core plugin entries except the volume package

ekinanp:
The volume package will be a bit more involved, so that work will be done separately.

### DIFF
--- a/plugin/aws/s3Bucket.go
+++ b/plugin/aws/s3Bucket.go
@@ -200,7 +200,7 @@ func (b *s3Bucket) Delete(ctx context.Context) (bool, error) {
 	// According to https://docs.aws.amazon.com/AmazonS3/latest/dev/delete-or-empty-bucket.html,
 	// we must delete the bucket's objects and object versions (for versioned buckets) before
 	// deleting the bucket itself. We defer deleting object versions as part of supporting
-	// versioned buckets, which is a separate piece of work.
+	// versioned buckets, which is a separate piece of work that's tracked in https://github.com/puppetlabs/wash/issues/561
 	err := deleteObjects(ctx, b.client, b.Name(), "")
 	if err != nil {
 		return false, fmt.Errorf("error deleting all the objects: %v", err)
--- a/plugin/aws/s3Bucket.go
+++ b/plugin/aws/s3Bucket.go
@@ -14,6 +14,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	s3Client "github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 )
 
 // listObjects is a helper that lists the objects which start with a specific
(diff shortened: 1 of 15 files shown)


plugin/aws/s3Bucket.go
```diff
@@ -186,6 +196,21 @@ func (b *s3Bucket) List(ctx context.Context) ([]plugin.Entry, error) {
 	return listObjects(ctx, b.client, b.Name(), "")
 }
 
+func (b *s3Bucket) Delete(ctx context.Context) (bool, error) {
+	// According to https://docs.aws.amazon.com/AmazonS3/latest/dev/delete-or-empty-bucket.html,
+	// we must delete the bucket's objects and object versions (for versioned buckets) before
+	// deleting the bucket itself. We defer deleting object versions as part of supporting
+	// versioned buckets, which is a separate piece of work.
```
ekinanp: Should this be an issue, or should we wait until someone requests versioned bucket support?

ekinanp:
Any activity that'd be useful to log here? The implementations seem simple enough that I figured there was no need.

plugin/aws/s3Bucket.go
MikaelSmith: We should create an issue, but probably won't start on it yet.

plugin/docker/container.go
```diff
@@ -88,6 +88,14 @@ func (c *container) List(ctx context.Context) ([]plugin.Entry, error) {
 	return []plugin.Entry{clf, cm, vol.NewFS("fs", c, 3)}, nil
 }
 
+func (c *container) Delete(ctx context.Context) (bool, error) {
+	err := c.client.ContainerRemove(ctx, c.id, types.ContainerRemoveOptions{
+		// TODO: Should we set RemoveVolumes and RemoveLinks too?
```
MikaelSmith: I'm not sure I understand what these mean. I'd guess we don't want to. I think those are things that can persist after and be reused, so they should be removed separately.

plugin/kubernetes/pod.go
```diff
@@ -65,3 +66,7 @@ func (p *pod) List(ctx context.Context) ([]plugin.Entry, error) {
 	return entries, nil
 }
 
+func (p *pod) Delete(ctx context.Context) (bool, error) {
+	err := p.client.CoreV1().Pods(p.ns).Delete(p.Name(), &v1.DeleteOptions{})
```
MikaelSmith: `v1` is already imported as `metav1`.

plugin/kubernetes/pvc.go
```diff
@@ -61,6 +62,11 @@ func (v *pvc) List(ctx context.Context) ([]plugin.Entry, error) {
 	return volume.List(ctx, v)
 }
 
+func (v *pvc) Delete(ctx context.Context) (bool, error) {
+	err := v.pvci.Delete(v.Name(), &v1.DeleteOptions{})
```
MikaelSmith: `v1` is already imported as `metav1`.

MikaelSmith (changes requested):

plugin/docker/container.go
ekinanp: I noticed those options in `docker rm` so thought I'd ask. I'll remove the TODO.

ekinanp:
Updated to address the feedback. Also filed https://github.com/puppetlabs/wash/issues/561 to track versioned buckets.

MikaelSmith (approved):

MikaelSmith:
Supports #479.

PR #567, opened 2019-11-05
Implement Delete for volumes

ekinanp:
Supports #479.

### DIFF
--- a/plugin/docker/volume.go
+++ b/plugin/docker/volume.go
@@ -91,12 +91,11 @@ func (v *volume) createContainer(ctx context.Context, cmd []string) (string, err
 	return created.ID, nil
 }
 
-func (v *volume) VolumeList(ctx context.Context, path string) (volpkg.DirMap, error) {
-	// Use a larger maxdepth because volumes have relatively few files and VolumeList is slow.
-	maxdepth := 10
-
-	// Create a container that mounts a volume and inspects it. Run it and capture the output.
-	cid, err := v.createContainer(ctx, volpkg.StatCmd(mountpoint+path, maxdepth))
+// Runs cmd in a temporary container. If the exit code is 0, then it returns the cmd's output.
+// Otherwise, it wraps the cmd's output in an error object.
+func (v *volume) runInTemporaryContainer(ctx context.Context, cmd []string) ([]byte, error) {
+	// Create a container that mounts a volume and deletes its file. Run rm -rf on it.
+	cid, err := v.createContainer(ctx, cmd)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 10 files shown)


volume/core.go
```diff
@@ -29,16 +31,25 @@ type Interface interface {
 	VolumeOpen(ctx context.Context, path string) (plugin.SizedReader, error)
 	// Accepts a path and streams updates to the content associated with that path.
 	VolumeStream(ctx context.Context, path string) (io.ReadCloser, error)
+	// Deletes the volume node at the specified path. Mirrors plugin.Deletable#Delete
+	VolumeDelete(ctx context.Context, path string) (bool, error)
```
ekinanp: It's unlikely that VolumeDelete will return false, but the general case is easy enough that we might as well support that capability if people ever want to use it. Also maintains symmetry w/ plugin.Deletable#Delete.

plugin/docker/volume.go
```diff
@@ -226,6 +227,58 @@ func (v *volume) VolumeStream(ctx context.Context, path string) (io.ReadCloser,
 	return plugin.CleanupReader{ReadCloser: output, Cleanup: killAndDelete}, nil
 }
 
+func (v *volume) VolumeDelete(ctx context.Context, path string) (bool, error) {
+	// Create a container that mounts a volume and deletes its file. Run rm -rf on it.
```
MikaelSmith: This is almost entirely a duplicate of VolumeList. Can we extract it into a common helper?

A bit more work would be to extract common elements of all the Volume* commands here into a helper struct that we can safely cleanup.

plugin/kubernetes/pvc.go
```diff
@@ -265,6 +266,42 @@ func (v *pvc) VolumeStream(ctx context.Context, path string) (io.ReadCloser, err
 	return plugin.CleanupReader{ReadCloser: output, Cleanup: delete}, nil
 }
 
+func (v *pvc) VolumeDelete(ctx context.Context, path string) (bool, error) {
+	// Create a container that mounts a pvc and inspects it. Run rm -rf on it.
```
MikaelSmith: This is also almost an exact copy of VolumeList. Only difference is the command run and that VolumeList uses the output in its return value.

MikaelSmith (changes requested):

ekinanp:
One sec, I see that VolumeOpen for pods runs cat on the pod, so am going to use the helper for that.

MikaelSmith (approved):

ekinanp:
Ok, updated.

MikaelSmith (approved):

PR #573, opened 2019-11-07
Download busybox if not present for listing Docker volumes

### LINKED ISSUE #312: Docker volume listing fails if "busybox" container unavailable
**Describe the bug**
I don't have the "busybox" container on my laptop. This results in errors listing docker volumes (maybe k8s pvc as well?).

**To Reproduce**
1. Remove the busybox container from your docker image list
2. Run wash
3. Try to do list a docker volume
4. You'll get something like: `could not get children of ./docker/volumes/fd_redis: puppetlabs.wash/errored-action: The list action errored on /Users/deepak/Library/Caches/wash/mnt612519020/docker/volumes/fd_redis: Error: No such image: busybox`

**Expected behavior**
Ideally, when we load the docker plugin we should do a check for the presence of busybox and, if it's not there, we should just download it on the user's behalf.

MikaelSmith:
The Docker plugin uses a busybox image to explore persistent volumes by spinning up an instance that mounts the volume and then interacting with it. If the busybox image is not present, creating the instance will fail rather than pull the image. Explicitly pull the busybox:latest image if it's missing.

Note that this isn't a problem for kubernetes as kubernetes already pulls images on-demand.

Fixes #312.

### DIFF
--- a/activity/writer.go
+++ b/activity/writer.go
@@ -1,0 +1,17 @@
+package activity
+
+import (
+	"context"
+	"strings"
+)
+
+// Writer logs the output as a call to Record per Write.
+type Writer struct {
+	context.Context
+	Prefix string
+}
+
+func (a Writer) Write(p []byte) (int, error) {
+	Record(a.Context, "%v: %v", a.Prefix, strings.TrimSpace(string(p)))
+	return len(p), nil
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ jobs:
   - name: Test Docker plugin
     before_script:
     - go install
-    - docker pull busybox
     - docker-compose -f examples/swarm/docker-compose.yml up -d
     script:
     - wash validate docker
--- a/activity/writer_test.go
+++ b/activity/writer_test.go
@@ -1,0 +1,61 @@
+package activity
+
+import (
+	"context"
+	"io"
+	"io/ioutil"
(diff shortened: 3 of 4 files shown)


activity/writer.go
```diff
@@ -0,0 +1,17 @@
+package activity
+
+import (
+	"context"
+	"strings"
+)
+
+// Writer logs the output as a call to Record per Write.
+type Writer struct {
```
ekinanp: This could wrap around an existing writer (decorate it w/ the ability to record to the activity journal).

ekinanp (approved):
Looks good, just a comment on generalizing `activity.Writer`.

activity/writer.go
MikaelSmith: That seems like an additional feature we don't need yet, not a change to the current solution.

MikaelSmith: I may not be clear on what you're suggesting. What does "decorate it w/ the ability to record to the activity journal" mean? That it would record to the activity journal while also recording to its original destination?

That can also be accomplished with this writer plus https://golang.org/pkg/io/#MultiWriter.

MikaelSmith: This is specifically a Writer interface to the activity journal.

ekinanp: Ah, didn't know about MultiWriter. Nevermind then.

Issue #548, opened 2019-10-16
Signal resources


MikaelSmith:
Many resources have state transitions that would be useful to be able to trigger easily, like start/stop/restart a VM or service.

Add a new Signal primitive that plugin entries can implement to allow sending arbitrary signals to a resource. Signal names are case-insensitive.
```
type Signalable interface {
  Signal(context.Context, string) error
}
```

We would like to establish some common conventions for this that make it easier to script without overly constraining the possible applications. As such, add to the plugin documentation around Signalable that `start`, `stop`, and `restart` are common signal names that you should re-use if possible.

Additionally `EntrySchema` should be updated to include a map of supported signals and descriptions of how the entry will respond to them. Those descriptions should be accessible through the `docs` command.

The Wash API should respond to `POST /fs/signal?path=<path>` by invoking Signal on the specified entry and returning whether sending the signal succeeded.

Add a new `signal` command that accepts `signal <signal name> path/to/resource`. It should error and display supported signals if the supplied signal name doesn't match one accepted by the entry at the specified path. It should return after sending the signal; waiting until the signaled operation has completed is an interesting concept, but not one we can do generally for any signal.

See https://docs.google.com/document/d/1NnJ1Qxeu7Sh4_ymOpo0enibKb3V5tzsSsX3H-N0PPDY/edit?usp=sharing for further discussion on the concept.

ekinanp:
Update:

`signal`'s usage was changed to `signal <path> <signal>`. This makes it easier to extend the `signal` command if we ever decide to extend the `signal` action. For example, if we decide to go with `Signal(ctx, signal, opts)`, then the `signal` command can have usage like:

```
signal --help
```
Print the default usage

```
signal <path> --help
```
Print the entry's supported signals/signal groups (to avoid cluttering docs if we add more CRUD primitives later on).

```
signal <path> <signal> --help
```
Print the supported options associated with `<signal>`. For example, if `<path>` is a Docker container, and `<signal>` is `stop`, then this would list all the options for Docker's `containers/stop` endpoint (or whatever other options the Docker plugin author makes available with the `stop` signal).

PR #574, opened 2019-11-07
Implement the signal action

ekinanp:
Signal for core plugins will be implemented separately. The website will also be updated as part of that work.

Supports #548.

### DIFF
--- a/api/types/signal.go
+++ b/api/types/signal.go
@@ -1,0 +1,7 @@
+package apitypes
+
+// SignalBody encapsulates the payload for a call to a plugin's Signal function
+type SignalBody struct {
+	// Name of the signal that's to be sent
+	Signal string `json:"signal"`
+}
--- a/plugin/entrySchema.go
+++ b/plugin/entrySchema.go
@@ -66,13 +66,14 @@ func schema(e Entry) (*EntrySchema, error) {
 }
 
 type entrySchema struct {
-	Label               string      `json:"label"`
-	Description         string      `json:"description,omitempty"`
-	Singleton           bool        `json:"singleton"`
-	Actions             []string    `json:"actions"`
-	MetaAttributeSchema *JSONSchema `json:"meta_attribute_schema"`
-	MetadataSchema      *JSONSchema `json:"metadata_schema"`
-	Children            []string    `json:"children"`
+	Label               string            `json:"label"`
+	Description         string            `json:"description,omitempty"`
+	Singleton           bool              `json:"singleton"`
+	Signals             map[string]string `json:"signals,omitempty"`
+	Actions             []string          `json:"actions"`
+	MetaAttributeSchema *JSONSchema       `json:"meta_attribute_schema"`
+	MetadataSchema      *JSONSchema       `json:"metadata_schema"`
+	Children            []string          `json:"children"`
 }
 
 // EntrySchema represents an entry's schema. Use plugin.NewEntrySchema
--- a/plugin/entrySchema.go
+++ b/plugin/entrySchema.go
@@ -66,14 +66,14 @@ func schema(e Entry) (*EntrySchema, error) {
 }
 
 type entrySchema struct {
-	Label               string            `json:"label"`
-	Description         string            `json:"description,omitempty"`
-	Singleton           bool              `json:"singleton"`
-	Signals             map[string]string `json:"signals,omitempty"`
-	Actions             []string          `json:"actions"`
-	MetaAttributeSchema *JSONSchema       `json:"meta_attribute_schema"`
-	MetadataSchema      *JSONSchema       `json:"metadata_schema"`
-	Children            []string          `json:"children"`
+	Label               string         `json:"label"`
+	Description         string         `json:"description,omitempty"`
+	Singleton           bool           `json:"singleton"`
+	Signals             []SignalSchema `json:"signals,omitempty"`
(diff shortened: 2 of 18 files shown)


ekinanp:
I tested this by adding signal support to the `mock_docker` plugin. Here's what the output looks like for docs (note I used AWS plugin root's docs to represent a typical description):
```
wash . ❯ docs mock_docker/mock_containers_dir/chalice
This is the AWS plugin root. The AWS plugin reads the AWS_SHARED_CREDENTIALS_FILE
environment variable or $HOME/.aws/credentials and AWS_CONFIG_FILE environment
variable or $HOME/.aws/config to find profiles and configure the SDK. The profiles
it lists can be limited by adding

aws:
  profiles: [profile_1, profile_2]

to Wash’s config file.

The AWS plugin currently supports EC2 and S3. IAM roles are supported when configured
as described here. Note that currently region will also need to be specified with the
profile.

If using MFA, Wash will prompt for it on standard input. Credentials are valid for 1 hour.
They are cached under wash/aws-credentials in your user cache directory so they can be
re-used across server restarts. Wash may have to re-prompt for a new MFA token in response
to navigating the Wash environment to authorize a new session.

SUPPORTED SIGNALS
* stop
    Stop the container
* start
    Start the container
```

ekinanp:
As for testing signal, signal for the mock_docker plugin's containers writes the passed-in signal to a file. For example,
```
wash . ❯ signal foo mock_docker/mock_containers_dir/chalice
puppetlabs.wash/bad-action-request: Bad request for signal action on /Users/enis.inan/Library/Caches/wash/mnt568356764/mock_docker/mock_containers_dir/chalice: invalid signal foo. Valid signals are start, stop
wash . ❯ signal start mock_docker/mock_containers_dir/chalice
wash . ❯ exit
Goodbye!
➜  wash git:(signal) ✗ cat container_signal
start
```

api/types/signal.go
```diff
@@ -0,0 +1,7 @@
+package apitypes
+
+// SignalBody encapsulates the payload for a call to a plugin's Signal function
+type SignalBody struct {
```
ekinanp: I chose this format because it's easier to extend it if `Signal` ever supports more arguments.

plugin/entrySchema.go
```diff
@@ -66,13 +66,14 @@ func schema(e Entry) (*EntrySchema, error) {
...
-	Description         string      `json:"description,omitempty"`
-	Singleton           bool        `json:"singleton"`
-	Actions             []string    `json:"actions"`
-	MetaAttributeSchema *JSONSchema `json:"meta_attribute_schema"`
-	MetadataSchema      *JSONSchema `json:"metadata_schema"`
-	Children            []string    `json:"children"`
+	Label               string            `json:"label"`
+	Description         string            `json:"description,omitempty"`
+	Singleton           bool              `json:"singleton"`
+	Signals             map[string]string `json:"signals,omitempty"`
```
ekinanp: I also considered supporting a tuple for methods like `(<method_name>, <method_schema>)` but that required a bit of a refactor and didn't generalize well to methods like `read`, `exec`, `schema` that didn't need `<method_schema>`. This way's also symmetric with wash ruby where there's a separate `signals` helper that lets you set the supported signals.

ekinanp: Hmm, it looks like for Docker containers you can also send any one of the supported Linux signals (like `SIGHUP`, `SIGSTOP`, etc.) via `docker kill`. It'd be tedious for a user to have to specify these signals in the schema. Wonder if we could provide a short-hand for that.

Since these can be vendor-specific, I think we should change `Signals` to `map[signals]SignalSchema`

ekinanp: Thinking about something like `map[string]SignalSchema` where `SignalSchema` is something like
```
type SignalSchema struct {
   Description
   Regex
}
```

If a Regex is provided, then we'll validate the provided signal using that. (At a higher-level, SignalSchemas w/ a Regex represent the schema of a group of signals, while w/o a Regex we are talking about a single Signal). Otherwise, we'll go w/ the signal name. This way, I can specify OS-specific signals as
```
"linux" => {
  "Description" => "Linux signals",
  "Regex" => "SIG*"
}
```

I'd prefer something like this instead of a hard-coded "linux" constant that people could use b/c that preserves Signal's general nature.

Could also change `Regex` to `Signals` to make validation easier.

MikaelSmith (changes requested):

plugin/entrySchema.go
ekinanp: Yeah, I think storing this as an array to preserve the order makes sense.

Linux signals introduces the more general problem of "My API has a bunch of signals, many of which fit a specific group/category. I'd like to have a way of telling Wash that information without having to list every signal because not only is that tedious, but it will also annoy users of my plugin, especially since those users will (likely) be aware of what that specific group/category of signals represent when they view my entry's set of supported signals."

Having a Linux constant isn't a good solution to that problem, esp. when the general solution is (probably) not that much harder.

Not sure what you mean by extra flag though. What would that look like?

MikaelSmith: Ah. I think I'd missed your last points. a Regex seems fine.

ekinanp:
Ok, updated to support signal groups.

MikaelSmith (approved):

PR #585, opened 2019-11-12
Implement signal for core plugins + update docs

### LINKED ISSUE #548: Signal resources

ekinanp:
Resolves https://github.com/puppetlabs/wash/issues/548

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -20,6 +20,8 @@ title: Docs
   * [wash tail](#wash-tail)
   * [wash validate](#wash-validate)
   * [wash docs](#wash-docs)
+  * [wash delete](#wash-delete)
+  * [wash signal](#wash-signal)
 * [Config](#config)
   * [wash.yaml](#wash-yaml)
   * [wash shell](#wash-shell)
@@ -29,6 +31,8 @@ title: Docs
   * [read](#read)
   * [stream](#stream)
   * [exec](#exec)
+  * [delete](#delete)
+  * [signal](#signal)
 * [Attributes](#attributes)
   * [crtime](#crtime)
   * [mtime](#mtime)
@@ -136,7 +140,15 @@ Each line represents validation of an entry type. The `lrsx` fields represent su
 
 ### wash docs
 
-Displays the entry's documentation. This is currently its description.
+Displays the entry's documentation. This is currently its description and any supported signals/signal groups.
+
+### wash delete
+
+Deletes the entries at the specified paths, prompting the user for confirmation before deleting each entry.
+
+### wash signal
+
+Sends the specified signal to the entry at the specified path.
 
 ## Config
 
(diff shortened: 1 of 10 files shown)


plugin/aws/ec2Instance.go
```diff
@@ -159,7 +159,12 @@ func (inst *ec2Instance) Schema() *plugin.EntrySchema {
 	return plugin.
 		NewEntrySchema(inst, "instance").
 		SetDescription(ec2InstanceDescription).
-		SetMetaAttributeSchema(ec2InstanceMetadata{})
+		SetMetaAttributeSchema(ec2InstanceMetadata{}).
+		AddSignal("start", "Starts the EC2 instance").
```
ekinanp: Think I'm going to refactor wash-ruby to adopt similar syntax.

ekinanp: https://github.com/puppetlabs/wash-ruby/pull/20

docs/docs/index.md
```diff
@@ -136,7 +140,15 @@ Each line represents validation of an entry type. The `lrsx` fields represent su
...
-Displays the entry's documentation. This is currently its description.
+Displays the entry's documentation. This is currently its description and any supported signals/signal groups.
+
+### wash delete
+
+Deletes the entries at the specified paths, prompting the user for confirmation before deleting each entry.
+
+### wash signal
+
+Sends the specified signal to the entry at the specified path
```
MikaelSmith: Should end with a period.

plugin/aws/ec2Instance.go
```diff
@@ -255,10 +260,7 @@ func (inst *ec2Instance) checkLatestConsoleOutput(ctx context.Context) (*ec2Inst
 }
 
 func (inst *ec2Instance) Delete(ctx context.Context) (bool, error) {
-	_, err := inst.client.TerminateInstancesWithContext(ctx, &ec2Client.TerminateInstancesInput{
-		InstanceIds: awsSDK.StringSlice([]string{inst.id}),
-	})
-	return false, err
+	return false, inst.Signal(ctx, "terminate")
```
MikaelSmith: It's weird that there's two ways to do this. But not necessarily a bad idea.

ekinanp: Yeah, opted for keeping things symmetrical w/ AWS docs

docs/docs/index.md
ekinanp: Thanks, updated.

ekinanp:
Might be interesting to have something like `signal(ctx, signal, opts)` where `opts` is JSON specifying the signal options. SignalSchema can be updated to include these options that we can check during validation. That would let us have access to all options for a given API request.

We can include the options for a given signal/signal group as part of docs' output.

Am OK with deferring that till it's needed though.

MikaelSmith:
Yes we should defer additional signal options. I'd want a little time to hash out the CLI options for it and some signal that it's needed.

MikaelSmith (approved):

PR #589, opened 2019-11-14
Add wash gem blog post

ekinanp:

### DIFF
--- a/docs/_posts/2019-11-14-wash-gem.md
+++ b/docs/_posts/2019-11-14-wash-gem.md
@@ -1,0 +1,160 @@
+---
+title: "Using the wash gem to write your external plugins"
+description: "Shows how the wash gem can be used to write external plugins"
+author: enisinan
+---
+
+Wash ships with plugins that let you interact with your AWS, GCP, Docker, and Kubernetes resources. You can also write your own plugins via the [external plugin interface]({{ '/docs/external-plugins' | relative_url }}). An external plugin consists of a *plugin* script. Wash shells out to this script whenever it needs to invoke an entry’s supported action (like *list*), or if it needs to query something about the entry (like its metadata). In a more general sense, Wash shells out to the plugin script whenever it needs to invoke an entry’s supported *method*. All plugin script invocations adopt the following calling convention
+
+```
+<plugin_script> <method> <path> <state> <args...>
+```
+
+where
+
+* `<method>` is the Wash method that’s being invoked. This includes Wash actions like `list` and `exec`, and also non-Wash actions like `schema` and `metadata`.
+
+* `<path>` is the entry’s filesystem path rooted at Wash’s mountpoint. For example, `/my_plugin` would be the `my_plugin`'s plugin root. `/my_plugin/foo` would be the `foo` child of the `my_plugin` entry.
+
+* `<state>` consists of the minimum amount of information required to reconstruct the entry inside the plugin. It can be _any_ string. For example, `'{"klass": "Root"}'` could be a JSON object representing the plugin root in a higher-level programming language like Python or Ruby.
+
+* `<args...>` are `<method>`’s arguments. For example if `<method>` is `exec`, then the exec’ed command would be included in `<args...>`.
+
+Now let `<entry> = <path> <state>`. Then the plugin script’s usage becomes
+
+```
+<plugin_script> <method> <entry> <args...>
+```
+
+From this usage, we see that `<path>` and `<state>` are two different representations of an entry. `<path>` is useful for simple plugins where reconstructing the entry is easy. `<state>` is useful for more complicated plugins where entries could be represented as classes.
+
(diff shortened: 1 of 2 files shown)


docs/_posts/2019-11-14-wash-gem.md
```diff
@@ -0,0 +1,132 @@
...
+end
+{% endhighlight %}
+
+Notice that
+
+* Each class extends the `Wash::Entry` class.
+* Entry schemas are declared via class-level helpers like `label`, `is_singleton`, and `parent_of`.
+* Each Wash method corresponds to an instance method in the class. For example, a report's `read` action is implemented by `Report#read`. Similarly, a PE instance's `list` action is implemented by `PEInstance#list`.
+* Entry attributes are declared via the `attributes` helper and can be set/accessed as instance-level fields. For example, a report's `mtime` and `meta` attributes are set in the constructor (`initialize` method).
+* The `state` helper lets you specify instance-level fields that should be serialized as part of the `state` key in the entry's JSON. The `wash gem` will set these fields after it reconstructs your entry. Thus, you can safely access those fields' values when implementing the entry's methods. For example, the `@pe_name` field in `Report` is initialized in the constructor and then referenced in `Report#read`.
```
MikaelSmith: This is the first place you talk about invoking the plugin as a script and serializing state. You also don't tie state to entry later. That feels like assuming that they read a significant part of the docs before this blog post.

I'd suggest moving the introduction of how plugin's are invoked (and introducing the concept of state) before walking through an excerpt of the implementation.

docs/_posts/2019-11-14-wash-gem.md
```diff
@@ -0,0 +1,132 @@
...
+
+The `Wash::Entry` class makes all of this possible. It also comes with even more helpers that let you do other things like specify cache TTLs (see `Wash::Entry#cache_ttls`) or any prefetched methods (see `Wash::Entry#prefetch`).
+
+Recall that plugin script invocations adopt the following calling convention
+
+```
+<plugin_script> <method> <entry> <args...>
+```
+
+All plugin script invocations are handled by the `Wash.run` function. Here's how `Wash.run` is used in `puppetwash`
```
MikaelSmith: I'd call out more clearly that `puppetwash` here refers to the script, and that the call to `Wash.run` is added at the end of it to actually run things. I don't feel like that's quite clear from what you've said here.

MikaelSmith (changes requested):

docs/_posts/2019-11-14-wash-gem.md
```diff
@@ -0,0 +1,132 @@
+---
+title: "Using the wash gem to write your external plugins"
+summary: "Shows how the wash gem can be used to write external plugins"
+author: Enis Inan
+---
+
+Wash ships with plugins that let you interact with your AWS, GCP, Docker, and Kubernetes resources. You can also write your own plugins via the [external plugin interface]({{ '/docs/external-plugins' | relative_url }}). If you're a Ruby developer, then the [wash gem](http://127.0.0.1:4000/wash/) makes this even easier by letting you model your entries as classes. We'll show you what this looks like by walking you through some of the [puppetwash](https://github.com/puppetlabs/puppetwash/tree/dadfbc3bb7b4f8fe379086182c3006237028ffc8) plugin's implementation, specifically the [puppetwash.rb](https://github.com/puppetlabs/puppetwash/blob/dadfbc3bb7b4f8fe379086182c3006237028ffc8/puppetwash.rb) file.
```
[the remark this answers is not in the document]
ekinanp: Yeah, let's tag puppetwash, that would be easier

docs/_posts/2019-11-14-wash-gem.md
```diff
@@ -0,0 +1,132 @@
+---
+title: "Using the wash gem to write your external plugins"
+summary: "Shows how the wash gem can be used to write external plugins"
+author: Enis Inan
```
MikaelSmith: Now that there's an authors section, you can reference it with `enisinan`.

ekinanp:
Updated. I think I've addressed all the feedback.

docs/_posts/2019-11-14-wash-gem.md
```diff
@@ -0,0 +1,160 @@
...
+
+Now let `<entry> = <path> <state>`. Then the plugin script’s usage becomes
+
+```
+<plugin_script> <method> <entry> <args...>
+```
+
+From this usage, we see that `<path>` and `<state>` are two different representations of an entry. `<path>` is useful for simple plugins where reconstructing the entry is easy. `<state>` is useful for more complicated plugins where entries could be represented as classes.
+
+If you're a Ruby developer, then the [wash gem](https://github.com/puppetlabs/wash-ruby/tree/0.1.2) makes external plugin development easy by letting you model your entries as classes. We'll show you what this looks like by walking you through some of the [puppetwash](https://github.com/puppetlabs/puppetwash/tree/dadfbc3bb7b4f8fe379086182c3006237028ffc8) plugin's implementation, specifically the [puppetwash.rb](https://github.com/puppetlabs/puppetwash/blob/dadfbc3bb7b4f8fe379086182c3006237028ffc8/puppetwash.rb) file.
```
MikaelSmith: This isn't a tagged version. Were you going to tag it?

MikaelSmith (approved):

docs/_posts/2019-11-14-wash-gem.md
ekinanp: Is tagging all we need to release it?

MikaelSmith: Yeah, I think so.

PR #597, opened 2019-11-17
Add blog post for the signal action

ekinanp:

### DIFF
--- a/docs/_posts/2019-11-16-signal.md
+++ b/docs/_posts/2019-11-16-signal.md
@@ -1,0 +1,77 @@
+---
+title: "Introducing signal"
+description: "Introduces the signal command and action"
+author: enisinan
+---
+
+Wash 0.13.0 includes a new `signal` command that lets you send an arbitrary signal to a given entry. For example, you can use `signal` to start a stopped container.
+
+```
+wash . ❯ meta docker/containers/wash_tutorial_redis_1
+...
+State:
+  Dead: false
+  Error: ""
+  ExitCode: 0
+  FinishedAt: "2019-11-16T18:58:03.7284542Z"
+  OOMKilled: false
+  Paused: false
+  Pid: 0
+  Restarting: false
+  Running: false
+  StartedAt: "2019-11-12T21:47:13.9406019Z"
+  Status: exited
+wash . ❯ signal docker/containers/wash_tutorial_redis_1 start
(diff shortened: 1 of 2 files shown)


docs/_posts/2019-11-16-signal.md
```diff
@@ -0,0 +1,77 @@
...
+
+Note that we're open to extending the `signal` action's power depending on user demand. Some things we're considering are:
+
+* Extending `signal` to take-in signal-specific API options. This would enable something like `signal docker/containers/wash_tutorial_redis_1 stop` to accept all of `docker stop`'s options. In a more general sense, this lets `signal` take full advantage of the plugin's API by allowing the user to pass-in API-specific options.
+
+* Extending `signal` to return an (optional) result of the signaled operation. Combined with (1), this would enable `signal` to be the "kitchen-sink" Wash action, where plugin authors can specify all vendor-specific operations that are not captured by any of the other Wash actions as signals (similar to an arbitrary method invocation).
```
ekinanp: This could be an interesting, formal definition of `signal` and would probably be a useful thing to support in the future.

docs/_posts/2019-11-16-signal.md
```diff
@@ -0,0 +1,77 @@
+---
+title: "Introducing signal"
+description: "Introduces the signal command and action"
+author: enisinan
+---
+
+Wash 0.13.0 includes a new `signal` command that lets you send an arbitrary signal to a given entry. For example, you can use `signal` to start a stopped container.
+
+```
+wash . ❯ meta docker/containers/wash_tutorial_redis_1
```
MikaelSmith: Could use `jq` to simplify this, though it's not that long anyway.

MikaelSmith (approved):

docs/_posts/2019-11-16-signal.md
MikaelSmith: I'm not sure how useful that is as a standard pattern. Much of the value comes from having broadly useful, comprehensible commands. If signal becomes a grab-bag for anything I think it loses those characteristics.

ekinanp: It can serve both purposes. This mostly came from trying to find a formal definition of signal. Started with "anything that has operational state", then the issue with that is maybe you may want to signal entries that don't necessarily have an operational state (or send a signal that may not change the operational state), so that means the definition has to be more general.

PR #600, opened 2019-11-20
Add Google Cloud Pub/Sub topics

### LINKED ISSUE #579: List and publish to GCP Pub/Sub
List available Pub/Sub message topics, and support publishing messages to them by piping a string (a la `echo message > path/to/pubsub/topic`). Also be able to tail a topic to see what messages are being published to it (without consuming them).

## Example

In one terminal, start watching the topic to see messages come in
```
tail -f path/to/topic
```

For a live topic, you should see the messages published to the queue without consuming them (so existing subscriptions still receive them). Not entirely sure if this is possible, a similar idea is to watch whatever logs exist on the topic.

In another terminal, submit a new message to the queue.
```
echo "test message" > path/to/topic
```
This message should then appear in the output from tailing the topic. This provides a simple way to get messages on the queue for testing downstream things that consume them.

## Implementation Notes

Prior art for listing Pub/Sub topics was https://github.com/puppetlabs/wash/blob/d9c5c276ad84a9181a0a9c4aa4a14211d6efa09d/gcp/pubsub.go. We'll need to figure out how to create a [named pipe](http://man7.org/linux/man-pages/man7/fifo.7.html) or something similar with libfuse and how to model it in Wash.

Publishing a message requires a new concept in Wash. We need to dig into this problem a little, then design what the interface should look like.

Note: publishing a message could be a custom command, but POSIX has a file whose concept mirrors a message queue - a FIFO - so we'd rather use that.

MikaelSmith:
Adds Google Cloud Pub/Sub topics as files that you can stream from to see messages (via a temporary subscription) and append to (`echo hello >> topic`) to publish a message.

Resolves #579.

### DIFF
--- a/fuse/file.go
+++ b/fuse/file.go
@@ -23,7 +23,7 @@ func newFile(p *dir, e plugin.Entry) *file {
 	return &file{newFuseNode("f", p, e)}
 }
 
-// Open a file for reading.
+// Open a file for reading or writing.
 func (f *file) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.OpenResponse) (fs.Handle, error) {
 	activity.Record(ctx, "FUSE: Open %v", f)
 
@@ -34,6 +34,9 @@ func (f *file) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.OpenR
 		return nil, err
 	}
 
+	var fh fileHandle
+	fh.id = f.String()
+
 	// Initiate content request and return a channel providing the results.
 	if plugin.ReadAction().IsSupportedOn(updatedEntry) {
 		content, err := plugin.OpenWithAnalytics(ctx, updatedEntry.(plugin.Readable))
@@ -43,31 +46,44 @@ func (f *file) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.OpenR
 		}
 
 		activity.Record(ctx, "FUSE: Opened %v", f)
-		return &fileHandle{r: content, id: f.String()}, nil
+		fh.r = content
 	}
+
+	if plugin.WriteAction().IsSupportedOn(updatedEntry) {
+		fh.w = updatedEntry.(plugin.Writable)
+	}
+
+	if fh.r != nil || fh.w != nil {
+		return &fh, nil
+	}
+
 	activity.Record(ctx, "FUSE: Open unsupported on %v", f)
 	return nil, fuse.ENOTSUP
 }
 
 type fileHandle struct {
 	r  io.ReaderAt
+	w  plugin.Writable
 	id string
 }
 
 var _ fs.Handle = (*fileHandle)(nil)
 var _ = fs.HandleReleaser(fileHandle{})
 var _ = fs.HandleReader(fileHandle{})
+var _ = fs.HandleWriter(fileHandle{})
 
-// Release closes the open file.
 func (fh fileHandle) Release(ctx context.Context, req *fuse.ReleaseRequest) error {
 	activity.Record(ctx, "FUSE: Release %v", fh.id)
 	if closer, ok := fh.r.(io.Closer); ok {
 		return closer.Close()
 	}
+
+	if closer, ok := fh.w.(io.Closer); ok {
+		return closer.Close()
+	}
 	return nil
 }
 
-// Read fills a buffer with the requested amount of data from the file.
 func (fh fileHandle) Read(ctx context.Context, req *fuse.ReadRequest, resp *fuse.ReadResponse) error {
 	buf := make([]byte, req.Size)
 	n, err := fh.r.ReadAt(buf, req.Offset)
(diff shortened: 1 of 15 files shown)


MikaelSmith:
This is functional. There's some follow-on work I'm doing to refactor the reading and writing interfaces to be more consistent.

plugin/types.go
```diff
@@ -195,6 +195,12 @@ type Readable interface {
 	Open(context.Context) (SizedReader, error)
 }
 
+// Writable is an entry that can have data appended to it.
```
ekinanp: Hmm, are we sticking with append because that is the (currently) supported behavior?

plugin/gcp/pubsubTopic.go
```diff
@@ -0,0 +1,160 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	watcher := &pubsubTopicWatcher{Subscription: sub, List: list.New(), Context: ctx, CancelFunc: cancel}
+
+	bufferMessages := func(_ context.Context, msg *pubsub.Message) {
+		msg.Ack()
+		watcher.Lock()
+		watcher.PushBack(msg)
+		watcher.Unlock()
+	}
+	go func() { watcher.err = sub.Receive(ctx, bufferMessages) }()
```
ekinanp: Should this be in a lock? Guessing no but wanted to ask anyways.

plugin/gcp/pubsubTopic.go
```diff
@@ -0,0 +1,160 @@
...
+type pubsubTopicWatcher struct {
+	*pubsub.Subscription
+	sync.Mutex
+	*list.List
+	context.Context
+	context.CancelFunc
+	err error
+}
+
+func newPubsubTopicWatcher(sub *pubsub.Subscription) *pubsubTopicWatcher {
```
ekinanp: Nit: This could be a little cleaner as `t.newWatcher() (*pubsubTopicWatcher, error)` because then you can put creation of the temp. subscription in there. It's not a huge deal though.

plugin/types.go
MikaelSmith: I need to update this help text.

plugin/gcp/pubsubTopic.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (t *pubsubTopic) Write(ctx context.Context, _ int64, b []byte) error {
+	result := t.topic.Publish(ctx, &pubsub.Message{Data: b})
+	_, err := result.Get(ctx)
+	return err
+}
+
+func (t *pubsubTopic) Schema() *plugin.EntrySchema {
+	return plugin.NewEntrySchema(t, "topic").
+		SetMetaAttributeSchema(&pubsubTopicMeta{}).
```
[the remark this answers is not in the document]
MikaelSmith: Oh right, those are different pieces.

fuse/file.go
```diff
@@ -78,3 +94,13 @@ func (fh fileHandle) Read(ctx context.Context, req *fuse.ReadRequest, resp *fuse
 	resp.Data = buf[:n]
 	return err
 }
+
+func (fh fileHandle) Write(ctx context.Context, req *fuse.WriteRequest, resp *fuse.WriteResponse) error {
+	// The Write interface doesn't allow returning partial writes because they never occur for remote writes.
+	err := fh.w.Write(ctx, req.Offset, req.Data)
```
ekinanp: Would it hurt to return the # of bytes written? If it's easy, we might as well keep the symmetry with `io.Writer`. People may want to use Wash for non-remote stuff too.

plugin/gcp/pubsubTopic.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (t *pubsubTopic) Schema() *plugin.EntrySchema {
+	return plugin.NewEntrySchema(t, "topic").
+		SetMetaAttributeSchema(&pubsubTopicAttrMeta{}).
+		SetMetadataSchema(&pubsubTopicMetadata{}).
+		SetDescription(pubsubTopicDescription)
+}
+
+const pubsubTopicDescription = `
+A Cloud Pub/Sub topic. You can pipe text to it to publish messages.
```
ekinanp: You could also refer to the `pubsub` dir description to avoid repeating yourself with the pipe example.

plugin/gcp/pubsubTopic.go
```diff
@@ -0,0 +1,173 @@
...
+	watcher := &pubsubTopicWatcher{Subscription: sub, List: list.New(), Context: subCtx, CancelFunc: cancel}
+
+	bufferMessages := func(_ context.Context, msg *pubsub.Message) {
+		msg.Ack()
+		watcher.Lock()
+		watcher.PushBack(msg)
+		watcher.Unlock()
+	}
+	go func() {
+		watcher.mux.Lock()
```
MikaelSmith: This breaks everything because receive is blocking.

plugin/gcp/pubsubTopic.go
```diff
@@ -0,0 +1,160 @@
...
+		Subscriptions:   subs,
+	}), nil
+}
+
+func (t *pubsubTopic) Delete(ctx context.Context) (bool, error) {
+	return true, t.topic.Delete(ctx)
+}
+
+// A ReadCloser that subscribes to a topic and buffers all messages that appear there.
+type pubsubTopicWatcher struct {
```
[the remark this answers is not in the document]
ekinanp: Still confusing, esp. now that there are two muxes.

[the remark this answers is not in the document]
MikaelSmith: Updated again.

ekinanp (approved):

PR #602, opened 2019-11-21
Add creation time and permission mask to ls

MikaelSmith:
Makes `ls` a bit more useful to check permissions. It's also more common
for an entry to have creation time than modification time, so add
creation time to long output.

### DIFF
--- a/cmd/ls.go
+++ b/cmd/ls.go
@@ -1,7 +1,7 @@
 package cmd
 
 import (
-	"strings"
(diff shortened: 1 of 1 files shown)


MikaelSmith:
If this looks like a good idea I'll create an issue.

ekinanp:
Creation time yes. Permissions no because it looks like we're just listing the supported actions, so why introduce an extra level of indirection?

PR #605, opened 2019-11-21
Add cloud functions to GCP plugin

### LINKED ISSUE #578: List and view logs for GCP Cloud Functions
Add GCP Cloud Functions as a Wash resource. Be able to list and delete them, as well as tail their logs.

API access is available via https://godoc.org/google.golang.org/api/cloudfunctions/v1, and https://cloud.google.com/functions/docs/monitoring/logging talks about accessing logs.

ekinanp:
Resolves #578

### DIFF
--- a/plugin/gcp/cloudFunction.go
+++ b/plugin/gcp/cloudFunction.go
@@ -3,6 +3,7 @@ package gcp
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/puppetlabs/wash/plugin"
@@ -12,12 +13,18 @@ import (
 type cloudFunction struct {
 	plugin.EntryBase
 	service cloudFunctionsProjectService
+	region  string
 }
 
 func newCloudFunction(function *cloudfunctions.CloudFunction, service cloudFunctionsProjectService) *cloudFunction {
+	// functionPath is formatted as projects/<project_id>/locations/<region>/functions/<function_name>
+	functionPath := function.Name
+	segments := strings.Split(functionPath, "/")
+	region, functionName := segments[3], segments[5]
 	cf := &cloudFunction{
-		EntryBase: plugin.NewEntry(function.Name),
+		EntryBase: plugin.NewEntry(functionName),
 		service:   service,
+		region:    region,
 	}
 	mtime, err := time.Parse(time.RFC3339, function.UpdateTime)
 	if err != nil {
@@ -31,7 +38,7 @@ func newCloudFunction(function *cloudfunctions.CloudFunction, service cloudFunct
 }
 
 func (cf *cloudFunction) List(ctx context.Context) ([]plugin.Entry, error) {
-	cfl, err := newCloudFunctionLog(ctx, cf.service, cf.Name())
+	cfl, err := newCloudFunctionLog(ctx, cf.service, cf.region, cf.Name())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


plugin/gcp/cloudFunction.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+type cloudFunction struct {
+	plugin.EntryBase
+	service cloudFunctionsProjectService
+}
+
+func newCloudFunction(function *cloudfunctions.CloudFunction, service cloudFunctionsProjectService) *cloudFunction {
+	cf := &cloudFunction{
+		EntryBase: plugin.NewEntry(function.Name),
```
ekinanp: Looks like cloud function names are of the form `project/<project_id>/location/<region>/functions/<function_name>`. Given this, how should we organize them? I'm thinking `<region>::<function_name>`.

plugin/gcp/cloudFunction.go
```diff
@@ -0,0 +1,55 @@
...
+	cfl, err := newCloudFunctionLog(ctx, cf.service, cf.Name())
+	if err != nil {
+		return nil, err
+	}
+	return []plugin.Entry{cfl}, nil
+}
+
+func (cf *cloudFunction) Delete(ctx context.Context) (bool, error) {
+	_, err := cf.service.Projects.Locations.Functions.Delete(cf.Name()).Context(ctx).Do()
+	return false, err
```
ekinanp: Cloud functions have some shut down stuff that needs to happen before they're deleted, which is why this is false.

plugin/gcp/cloudFunction.go
MikaelSmith: VM instances are also in a specific region/zone. As long as name uniqueness is enforced across zones I didn't put it in.

plugin/gcp/cloudFunction.go
```diff
@@ -0,0 +1,55 @@
...
+	plugin.EntryBase
+	service cloudFunctionsProjectService
+}
+
+func newCloudFunction(function *cloudfunctions.CloudFunction, service cloudFunctionsProjectService) *cloudFunction {
+	cf := &cloudFunction{
+		EntryBase: plugin.NewEntry(function.Name),
+		service:   service,
+	}
+	mtime, err := time.Parse(time.RFC3339, function.UpdateTime)
```
MikaelSmith: Huh, only some services add a creation timestamp so it's in "Metadata".

ekinanp: Yeah, it's weird.

plugin/gcp/cloudFunctionLog.go
```diff
@@ -0,0 +1,182 @@
...
+	afterTimestamp string
+	cfl            *cloudFunctionLog
+}
+
+func (s *cloudFunctionLogStreamer) Read(p []byte) (n int, err error) {
+	for {
+		if len(s.currentEntries) > 0 {
+			break
+		}
+		time.Sleep(2 * time.Second)
```
MikaelSmith: Why sleep for 2 seconds?

MikaelSmith: Oh, because copy will busyread.

MikaelSmith: I wish there were a better way, but I don't see it.

MikaelSmith (approved):

ekinanp:
Updated to omit project + region from function name.

MikaelSmith (approved):

PR #613, opened 2019-11-25
Support cloud run services

### LINKED ISSUE #612: List and view logs for Cloud Run services
Add GCP cloud run services as a Wash resource. Be able to list and delete them, and tail their logs.

API access: https://godoc.org/google.golang.org/api/run/v1

Accessing logs is likely similar to what's done for cloud functions in https://github.com/puppetlabs/wash/blob/master/plugin/gcp/cloudFunctionLog.go. We may be able to re-use some of the streaming logic.

ekinanp:
Resolves https://github.com/puppetlabs/wash/issues/612

### DIFF
--- a/plugin/gcp/cloudRunDir.go
+++ b/plugin/gcp/cloudRunDir.go
@@ -1,0 +1,62 @@
+package gcp
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/puppetlabs/wash/plugin"
+	"google.golang.org/api/option"
+	"google.golang.org/api/run/v1"
+)
+
+type cloudRunProjectAPIService struct {
+	*run.APIService
+	projectID string
+	// We need to pass this around to access cloud function logs
+	client *http.Client
+}
+
+type cloudRunDir struct {
+	plugin.EntryBase
+	apiService cloudRunProjectAPIService
+}
+
+func newCloudRunDir(ctx context.Context, client *http.Client, projID string) (*cloudRunDir, error) {
+	svc, err := run.NewService(context.Background(), option.WithHTTPClient(client))
+	if err != nil {
+		return nil, err
+	}
+	cr := &cloudRunDir{
+		EntryBase:  plugin.NewEntry("cloud_run"),
+		apiService: cloudRunProjectAPIService{APIService: svc, projectID: projID, client: client},
+	}
+	if _, err := plugin.List(ctx, cr); err != nil {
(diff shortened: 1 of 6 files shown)


plugin/gcp/cloudRunServiceLog.go
```diff
@@ -0,0 +1,54 @@
...
+	"google.golang.org/api/logging/v2"
+)
+
+type cloudRunServiceLog struct {
+	plugin.EntryBase
+	*cloudLogFile
+}
+
+func newCloudRunServiceLog(ctx context.Context, apiService cloudRunProjectAPIService, region string, serviceName string) (*cloudRunServiceLog, error) {
+	fields := []cloudLogEntryField{
```
ekinanp: I couldn't find the cloud run service equivalent of `gcloud functions log read <function_name>`, so kind of guessed these field values based on symmetry w/ a cloud function's log.

plugin/gcp/cloudRunDir.go
```diff
@@ -0,0 +1,62 @@
...
+	apiService cloudRunProjectAPIService
+}
+
+func newCloudRunDir(ctx context.Context, client *http.Client, projID string) (*cloudRunDir, error) {
+	svc, err := run.NewService(context.Background(), option.WithHTTPClient(client))
+	if err != nil {
+		return nil, err
+	}
+	cr := &cloudRunDir{
+		EntryBase:  plugin.NewEntry("cloud_run"),
```
MikaelSmith: I guess `run` might be a little too generic. Would be symmetrical with `Cloud Storage` though.

MikaelSmith (approved):

plugin/gcp/cloudRunServiceLog.go
MikaelSmith: https://godoc.org/google.golang.org/api/logging/v2#LogEntry talks about the default when Severity is emptying being `DEFAULT`.

plugin/gcp/cloudRunDir.go
ekinanp: I went with what's shown in the console. There, `Cloud Storage` is `Storage` while `Cloud Run` is `Cloud Run`.

plugin/gcp/cloudRunServiceLog.go
ekinanp: Ah thanks, updated.

MikaelSmith (approved):

PR #616, opened 2019-11-25
(WIP) Move unknown size calculation to FUSE

ekinanp:
This is where it's needed.

Fixes <TODO: Link issue #>

### DIFF
--- a/fuse/core.go
+++ b/fuse/core.go
@@ -162,6 +162,18 @@ func (f *fuseNode) Attr(ctx context.Context, a *fuse.Attr) error {
 		return err
 	}
 	attr := plugin.Attributes(updatedEntry)
+	if !attr.HasSize() && plugin.ReadAction().IsSupportedOn(updatedEntry) {
+		// This Readable entry doesn't know its size, so download its content and set the size
+		// to that. This enables built-in tools like `cat` to work without slowing down `list`.
+		//
+		// TODO: How can we prevent tab-completion from being slow?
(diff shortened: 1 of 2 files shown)


fuse/core.go
```diff
@@ -162,6 +162,18 @@ func (f *fuseNode) Attr(ctx context.Context, a *fuse.Attr) error {
 		return err
 	}
 	attr := plugin.Attributes(updatedEntry)
+	if !attr.HasSize() && plugin.ReadAction().IsSupportedOn(updatedEntry) {
+		// This Readable entry doesn't know its size, so download its content and set the size
+		// to that. This enables built-in tools like `cat` to work without slowing down `list`.
+		//
+		// TODO: How can we prevent tab-completion from being slow?
```
ekinanp: Not sure how to do this. Could we check for tab-completion process by checking for the shell's PID?

Either way, I think moving this stuff over to FUSE is the right move. It seemed weird to include a size sometimes, and not include a size other times -- `plugin.Attributes` should return what the plugin author provided.

ekinanp: Also OK with deferring the fix for later since other users have not yet raised this issue.

MikaelSmith: Yeah, checking for the shell's PID might actually work.

ekinanp: Good news is this fix doesn't slow down `ls` or `find` since those use the API. Right now it's just tab-completion.

PR #617, opened 2019-11-25
Refactor Read/Write primitives

### LINKED ISSUE #601: File lifecycle primitives: Read & Write
Finish file lifecycle primitives so we can read and write files.

FUSE operates on blocks of a file. So every request will be for a Read or Write of a length of bytes at an offset. This maps cleanly to Go's `io.ReaderAt` and `io.WriterAt` interfaces. Read operations via FUSE will only read up to the Size attribute of the file, so we need to know the file size before performing any reads.

This has several use cases our primitives need to satisfy:
- Read a file where I know the size.
- I have read-access to a log or other data source with unknown size that I can only read via a Reader or download fully.
- I'm operating on a file through an actual filesystem.
- Writing to an entry with no persistent content like a message queue.
- Update a file.

Implementing this will be done in multiple subissues
- [x] Clearer interface for Read primitives (#619)
- [x] General Write primitive (#620)

MikaelSmith:
Resolves #601.

### DIFF
--- a/api/fs/file.go
+++ b/api/fs/file.go
@@ -1,9 +1,7 @@
 package apifs
 
 import (
-	"bytes"
 	"context"
-	"io/ioutil"
 	"os"
 
 	"github.com/puppetlabs/wash/plugin"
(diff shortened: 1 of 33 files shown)


MikaelSmith:
This doesn't match what I described in #601 yet, particularly for writing. Reading may be close.

PR #623, opened 2019-11-26
Generalize console control

### LINKED ISSUE #297: Kubernetes config with 'exec' resource where exec requires MFA hangs
A common way to configure AWS EKS in kube config is
```
user:
  exec:
    apiVersion: client.authentication.k8s.io/v1alpha1
    command: aws-iam-authenticator
    args: [token, -i, infracore-test-1, --cache]
    env:
    - name: AWS_PROFILE
      value: <profile>
```

This will cache and re-use an MFA token if using an account with MFA, but if the token needs to be re-requested Wash hangs, then eventually times out and crashes the terminal.

We need to identify and expose that request for an MFA token when it happens.

----

For completeness, the way the AWS CLI tool sets up kube config is with
```
user:
  exec:
    apiVersion: client.authentication.k8s.io/v1alpha1
    command: aws
    args: [eks get-token --cluster-name <cluster>]
    env:
    - name: AWS_PROFILE
      value: <profile>
```
which doesn't cache the token at all, so it attempts to request an MFA token every time you run kubectl.

MikaelSmith:
Sometimes a plugin needs control of the console for a prompt it doesn't make directly. Get control before calling all primitives that take a context (as a proxy for potential network activity or execs).

This isn't entirely safe. If it's backgrounded and prompts for input it may not exit cleanly. Using `plugin.Prompt` guards against this issue.

Fixes #297.

### DIFF
--- a/plugin/interactive.go
+++ b/plugin/interactive.go
@@ -1,9 +1,12 @@
 package plugin
 
 import (
+	"context"
 	"fmt"
 	"os"
+	"os/signal"
 	"sync"
+	"syscall"
 	"unsafe"
 
 	"github.com/mattn/go-isatty"
@@ -35,18 +38,14 @@ func tcSetpgrp(fd int, pgrp int) (err error) {
 	return unix.IoctlSetInt(fd, unix.TIOCSPGRP, int(uintptr(unsafe.Pointer(&v))))
 }
 
-// Only allow one Prompt call at a time. This prevents multiple plugins loading concurrently
-// from messing things up by calling Prompt concurrently.
-var promptMux sync.Mutex
-
-// Prompt prints the supplied message, then waits for input on stdin.
-func Prompt(msg string) (string, error) {
+// withConsole invokes the specified function while Wash has control of console input.
+// The function should save results by writing to captured variables.
+func withConsole(ctx context.Context, fn func(context.Context) error) error {
 	if !IsInteractive() {
-		return "", fmt.Errorf("not an interactive session")
+		// If not interactive, all we can do is call the function and return.
+		// If the function prompts for input without checking whether it can, then it may crash.
+		return fn(ctx)
 	}
-
-	promptMux.Lock()
-	defer promptMux.Unlock()
 
 	// Even if Wash is running interactively, it will not have control of STDIN while another command
 	// is running within the shell environment. If it doesn't have control and tries to read from it,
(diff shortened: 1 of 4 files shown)


ekinanp:
So this handles the case when the plugin's API prompts for something, right?

plugin/interactive.go
```diff
@@ -56,27 +63,39 @@ func Prompt(msg string) (string, error) {
...
 	curGrp := unix.Getpgrp()
 
-	var v string
-	if inGrp == curGrp {
-		// We control stdin
-		fmt.Fprintf(os.Stderr, "%s: ", msg)
-		_, err = fmt.Scanln(&v)
-	} else {
-		// Need to get control, prompt, then return control.
+	if inGrp != curGrp {
```
MikaelSmith: The problem might be that Ctrl-C is terminating the daemon when it has stdin. So I'm going to try catching SIGTERM and telling it to stop the context (meaning fn needs to be more opinionated about its signature).

MikaelSmith:
Yes.

MikaelSmith:
Ok, I think this works better now.

plugin/interactive.go
```diff
@@ -56,27 +55,51 @@ func Prompt(msg string) (string, error) {
...
 	if inGrp == curGrp {
-		// We control stdin
+		return fn(ctx)
+	}
+
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGINT)
+	cancelCtx, cancel := context.WithCancel(ctx)
+	go func() {
+		<-sigCh
```
ekinanp: This won't block forever, will it? Guessing the `Reset` below prevents that?

plugin/methodWrappers.go
```diff
@@ -198,8 +210,10 @@ func Signal(ctx context.Context, s Signalable, signal string) error {
 	}
 
 	// Go ahead and send the signal
-	err = s.Signal(ctx, signal)
-	if err != nil {
+	if err = withConsole(ctx, func(c context.Context) (err error) {
```
ekinanp: Nit: `err =` part on separate line

plugin/methodWrappers.go
```diff
@@ -218,8 +232,10 @@ func Signal(ctx context.Context, s Signalable, signal string) error {
 
 // Delete deletes the given entry.
 func Delete(ctx context.Context, d Deletable) (deleted bool, err error) {
-	deleted, err = d.Delete(ctx)
-	if err != nil {
+	if err = withConsole(ctx, func(c context.Context) (err error) {
```
ekinanp: Nit: `err =` part on separate line

ekinanp (commented):
Most comments are about moving `err :=` stuff to a separate line. I think only one asks a legitimate question. Otherwise, this LGTM.

plugin/methodWrappers.go
MikaelSmith: You and I just disagree on this.

plugin/interactive.go
MikaelSmith: Hmm, it might. I can probably do something about that.

MikaelSmith: Should be addressed now.

ekinanp: Why not just close `sigCh` in a defer? You'll still need to call the cancel func either way

MikaelSmith: Oh, didn't think of it.

ekinanp (approved):

PR #625, opened 2019-11-27
Support multiple arguments for ls, meta, signal, winfo and clear + other improvements

### LINKED ISSUE #598: Most commands should accept multiple arguments
The following commands should accept multiple arguments: `ls`, `meta`, `signal`, `winfo`, `wclear`.

`wexec` would be nice to make work with multiple arguments as well, but we don't have an obvious pattern for how it should look yet.

ekinanp:
Resolves https://github.com/puppetlabs/wash/issues/598

### DIFF
--- a/cmd/clear.go
+++ b/cmd/clear.go
@@ -8,13 +8,12 @@ import (
 func clearCommand() *cobra.Command {
 	use, aliases := generateShellAlias("clear")
 	clearCmd := &cobra.Command{
-		Use:     use + " [<path>]",
+		Use:     use + " [<path>]...",
 		Aliases: aliases,
-		Short:   "Clears the cache at <path>, or current directory if not specified",
+		Short:   "Clears the cache at the specified paths, or current directory if not specified",
 		Long: `Wash caches most operations. If the resource you're querying appears out-of-date, use this
-subcommand to reset the cache for resources at or contained within <path>. Defaults to the current
-directory if <path> is not specified.`,
-		Args: cobra.MaximumNArgs(1),
+subcommand to reset the cache for resources at or contained within the specified paths.
+Defaults to the current directory if no path is provided.`,
 		RunE: toRunE(clearMain),
 	}
 	clearCmd.Flags().BoolP("verbose", "v", false, "Print paths that were cleared from the cache")
@@ -22,9 +21,9 @@ directory if <path> is not specified.`,
 }
 
 func clearMain(cmd *cobra.Command, args []string) exitCode {
-	path := "."
+	paths := []string{"."}
 	if len(args) > 0 {
-		path = args[0]
+		paths = args
 	}
 	verbose, err := cmd.Flags().GetBool("verbose")
 	if err != nil {
@@ -32,19 +31,28 @@ func clearMain(cmd *cobra.Command, args []string) exitCode {
 	}
 
 	conn := cmdutil.NewClient()
-	cleared, err := conn.Clear(path)
-	if err != nil {
-		cmdutil.ErrPrintf("%v\n", err)
-		return exitCode{1}
-	}
 
-	if verbose {
-		for _, p := range cleared {
-			cmdutil.Println("Cleared", p)
+	// Perform the operation. Note that wclear isn't parallelized because
+	// conn.Clear only hits the Wash daemon. Thus, it should be a very fast
+	// request.
+	ec := 0
+	for _, path := range paths {
+		cleared, err := conn.Clear(path)
+		if err != nil {
+			ec = 1
+			cmdutil.ErrPrintf("%v: %v\n", path, err)
+			continue
 		}
-	} else {
-		cmdutil.Println("Cleared", path)
+
+		if verbose {
+			for _, p := range cleared {
+				cmdutil.Println("Cleared", p)
+			}
+		} else {
+			cmdutil.Println("Cleared", path)
+		}
 	}
 
-	return exitCode{0}
+	// Return the exit code
+	return exitCode{ec}
 }
--- a/cmd/ls.go
+++ b/cmd/ls.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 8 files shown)


cmd/signal.go
```diff
@@ -8,8 +8,8 @@ import (
 
 func signalCommand() *cobra.Command {
 	signalCmd := &cobra.Command{
-		Use:   "signal <path> <signal>",
-		Short: "Sends the specified signal to the entry at the specified path",
+		Use:   "signal <signal> [path]...",
```
ekinanp: @MikaelSmith If "make supporting multiple paths easy" was your reason for the original `signal <signal> <path` usage, then I apologize for changing it. I didn't think people would want multiple paths so quickly. If you're OK w/ the reasoning in the commit message, then how do you recommend I update the `signal` blog post?

cmd/ls.go
```diff
@@ -13,24 +15,19 @@ import (
...
+no path is specified. If the -l option is set, then the name,
+last modified time, and supported actions are displayed for
+each child.`,
+		RunE: toRunE(lsMain),
 	}
 	lsCmd.Flags().BoolP("long", "l", false, "List in long format")
 	return lsCmd
 }
 
-func headers() []cmdutil.ColumnHeader {
```
ekinanp: Removing these makes the output cleaner with multiple paths. The built-in `ls -l` also doesn't have any headers for its long format, so I think we should be safe here.

ekinanp:
ls' usage is meant to be as consistent as possible with the built-in `ls`. Here are some examples:

```
wash . ❯ ls docker
containers/
volumes/
wash . ❯ ls -l docker
containers/   <unknown>   list
volumes/      <unknown>   list
wash . ❯ ls docker docker/containers/wash_tutorial_redis_1/log aws foo bar
ls: bar: puppetlabs.wash/plugin-does-not-exist: Plugin bar does not exist
ls: foo: puppetlabs.wash/plugin-does-not-exist: Plugin foo does not exist
docker/containers/wash_tutorial_redis_1/log

aws:
demo_one/
demo_two/

docker:
containers/
volumes/
wash . ❯ ls -l docker docker/containers/wash_tutorial_redis_1/log aws foo bar
ls: bar: puppetlabs.wash/plugin-does-not-exist: Plugin bar does not exist
ls: foo: puppetlabs.wash/plugin-does-not-exist: Plugin foo does not exist
docker/containers/wash_tutorial_redis_1/log   <unknown>   read, stream

aws:
demo_one/                                     <unknown>   list
demo_two/                                     <unknown>   list

docker:
containers/                                   <unknown>   list
volumes/                                      <unknown>   list
```

ekinanp:
`winfo` and `meta` output a map of `<path> => <info>/<metadata>` when given multiple paths. For a single path, their output remains the same

cmd/signal.go
MikaelSmith: I think that is what I was thinking. Exec is still a problem, but I think by it's nature it's going to be hard to make it consistent with everything else.

I'd change the blog post, and add an `Update: ...` note at the top explaining that it's been updated for changes to `signal` in Wash release x.y (once we release).

cmd/ls.go
MikaelSmith: Makes sense. It'd be nice if `<unknown>` somehow reflected that it was expected to be a time, but not sure how to do that without being overly verbose. Maybe `<no mtime>`?

cmd/signal.go
ekinanp: Agreed on exec.

Filed https://github.com/puppetlabs/wash/issues/627 for the blog post update.

cmd/ls.go
```diff
@@ -42,59 +39,163 @@ func cname(entry apitypes.Entry) string {
...
+//   * A "file"   -- entry that does not implement "list"
+//   * A "dir"    -- entry that does implement "list"
+type lsItem struct {
+	path     string
+	entry    apitypes.Entry
+	children []apitypes.Entry
+	err      error
+}
+
+const errorItem = 0
```
MikaelSmith: Common to put these in a single statement to show they're grouped:
```
const (
  errorItem = 0
  fileItem  = 1
  dirItem   = 2
)
```

cmd/ls.go
```diff
@@ -42,59 +39,163 @@ func cname(entry apitypes.Entry) string {
...
-			cmdutil.ErrPrintf("%v\n", err)
-			return exitCode{1}
+	// Fetch the required data
+	var items []lsItem
+	for _, path := range paths {
+		var item lsItem
+		item.path = path
+		item.entry, item.err = conn.Info(path)
+		if item.err == nil && item.Type() == dirItem {
+			item.children, item.err = conn.List(path)
```
MikaelSmith: Can we do this in parallel?

cmd/ls.go
```diff
@@ -42,59 +39,163 @@ func cname(entry apitypes.Entry) string {
...
 
-	if longFormat {
-		cmdutil.Print(formatLSEntries(entries))
-	} else {
-		for _, entry := range entries {
-			cmdutil.Println(cname(entry))
+	// Sort the items to ensure that the output's
+	// printed in the expected "errors", "files",
+	// and "dirs" order.
+	sort.Sort(lsItemSlice(items))
```
MikaelSmith: If all we need is to sort once, then https://golang.org/pkg/sort/#Slice lets you provide just the less function.

cmd/clear.go
```diff
@@ -8,43 +8,51 @@ import (
...
-	cleared, err := conn.Clear(path)
-	if err != nil {
-		cmdutil.ErrPrintf("%v\n", err)
-		return exitCode{1}
-	}
 
-	if verbose {
-		for _, p := range cleared {
-			cmdutil.Println("Cleared", p)
+	// Perform the operation. Note that wclear isn't parallelized because
```
MikaelSmith: Is there some reason you didn't parallelize the rest?

MikaelSmith (changes requested):

cmd/ls.go
ekinanp: `<mtime unknown>` ?

cmd/clear.go
ekinanp: Yeah, because we need to fetch all the data before doing anything else. But I can go ahead and parallelize that part, it's not that much harder and it is a bit of a performance gain.

cmd/ls.go
ekinanp: Sure

cmd/ls.go
ekinanp: Thanks, that's much cleaner.

cmd/ls.go
MikaelSmith: I'm ok with that.

cmd/clear.go
MikaelSmith: I meant, didn't parallelize the other commands in that commit.

ekinanp: Yeah, that's what I'm referring to.

ekinanp:
Ok, updated to clean-up ls' implementation and to parallelize data fetching in ls, winfo and meta.

cmd/util/format.go
```diff
@@ -36,6 +48,13 @@ func Printf(msg string, a ...interface{}) {
 	}
 }
 
+// SafePrintf is a thread-safe wrapper to Printf
```
MikaelSmith: This would probably be fine as the default implementation, locking is relatively cheap and we always want serialized output.

MikaelSmith (approved):

cmd/util/format.go
ekinanp: Meh, let's leave it as-is. Not that much more complicated, and it could be useful to export the locks for bulk writes in the future.

PR #637, opened 2019-12-03
Fix Wash regressions

MikaelSmith:
Fix regressions introduced by #635. Supports #297.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,8 @@ To build it, run `go build`. To test, run `go test ./...`.
 
 > Requires golang 1.12+.
 
+Acceptance tests are in the [test](test) directory. First build with `go build`, then run them using [BATS](https://github.com/bats-core/bats-core) with `PATH=.:$PATH bats test`.
+
 See the [core plugin docs](CORE_PLUGIN_DEVELOPMENT.md) if you're interested in core plugin development.
 
 ## Roadmap
--- a/cmd/rootMain.go
+++ b/cmd/rootMain.go
@@ -59,6 +59,26 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
 	// Set plugin interactivity to false if execfile or rootCommandFlag were specified.
 	plugin.InitInteractive(execfile == "" && rootCommandFlag == "")
 
+	// If interactive and this process is in its own process group, then fork and run the original
+	// command as a new process that's not the leader of its process group. This is specifically to
+	// work with shells (zsh) that try to use their parent's process group, which would no longer be
+	// in the same session. By forking, we keep the original process group in its original session so
+	// the child shell can still modify it.
+	if plugin.IsInteractive() && os.Getpid() == unix.Getpgrp() {
+		comm := exec.Command(os.Args[0], os.Args[1:]...)
+		comm.Stdin = os.Stdin
+		comm.Stdout = os.Stdout
+		comm.Stderr = os.Stderr
+		if err := comm.Run(); err != nil {
+			if exitErr, ok := err.(*exec.ExitError); ok {
+				return exitCode{exitErr.ExitCode()}
+			}
+			cmdutil.ErrPrintf("%v\n", err)
+			return exitCode{1}
+		}
+		return exitCode{0}
+	}
+
 	// TODO: instead of running a server in-process, can we start one in a separate process that can
 	//       be shared between multiple invocations of `wash`?
 	plugins, serverOpts, err := serverOptsFor(cmd)
--- a/cmd/rootMain.go
+++ b/cmd/rootMain.go
@@ -6,7 +6,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
-	"time"
 
 	"golang.org/x/sys/unix"
 
@@ -21,7 +20,16 @@ import (
 // rootCommandFlag is associated with the `-c` option of the root command, set in root.go.
 var rootCommandFlag string
 
-// Start the wash server, then present the default system shell.
+// Start the Wash server then present the default system shell. The server will be running in the
+// current process, while the shell will be in a separate child process. We'd like the server to be
+// able to prompt for input without interupting the shell (as the controller of the terminal) so
+// that plugins can prompt the user for input like e.g. security tokens.
+//
+// To allow prompts, we start the shell process then put the Wash server (daemon) in a new session
+// with `setsid`. As a new session, the daemon has a different controlling terminal and can
+// therefore prompt for input without having to control the shell's terminal. This approach was
+// inspired by https://blog.nelhage.com/2011/02/changing-ctty/.
+//
 // On exit, stop the server and return any errors.
 func rootMain(cmd *cobra.Command, args []string) exitCode {
 	// Configure logrus to emit simple text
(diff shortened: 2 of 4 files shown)


cmd/rootMain.go
```diff
@@ -125,24 +125,41 @@ Try 'help'`)
...
 			}
 		} else {
-			codeCh <- 0
+			exitCodeCh <- 0
 		}
-		close(codeCh)
+		close(exitCodeCh)
 	}()
 
+	// Only switch process group if interactive, where the child starts a new session.
```
ekinanp: Nit: This should probably contain some details about why we're doing this for future maintainers. Thinking of something like

> Only switch process group if interactive, where the child starts a new session. This way, the daemon and the shell share the same TTY, which enables plugins to prompt the user for input like e.g. security tokens.

README.md
```diff
@@ -26,6 +26,8 @@ To build it, run `go build`. To test, run `go test ./...`.
 
 > Requires golang 1.12+.
 
+Additional unit tests are in the [test](test) directory. First built with `go build`, then run them using [BATS](https://github.com/bats-core/bats-core) with `PATH=.:$PATH bats test`.
```
ekinanp: This is a pretty good idea. Only concern is that the stuff in `test` tests the Wash binary, so they're not really unit tests. They seem more like acceptance tests?

MikaelSmith: Ah yes, wrote the wrong thing.

MikaelSmith:
There's one more issue I'm investigating with zsh
```
zsh: can't set tty pgrp: operation not permitted
```

One difference I've found is that running `go run wash.go` starts with Wash in the `go run ...` process group, and when zsh exits it returns the tty to that process group. Where-as starting `wash`, it starts in its own process group which gets reparented to be under the zsh session, so trying to return the tty to the parent process group fails because you're not allowed to switch control that direction.

MikaelSmith:
Updated.

cmd/rootMain.go
```diff
@@ -59,6 +59,25 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
 	// Set plugin interactivity to false if execfile or rootCommandFlag were specified.
 	plugin.InitInteractive(execfile == "" && rootCommandFlag == "")
 
+	// If interactive and this process is in its own process group, then fork and run the original
+	// command as a new process that's not the leader of its process group. This is specifically to
+	// work with shells that try to use the parent's process group, which we change later on. By
```
ekinanp: What does "try to use the parent's process group" mean? That some shells try to use the parent's process group's TTY?

MikaelSmith: In this case I think it's returning the TTY to the parent process group. But I don't fully understand what zsh is doing.

ekinanp: Ok. Now say `P` is the server process in pgrp `A`, `C` is the child shell process in pgrp `B`, and that `C` needs `A` (it is one of those shells you mention in the comment). Without this code, we'd get `P -- B` at the end, which resulted a failure because it didn't match `A`. Does that mean that `C` knows it needs `A` when it boots up (i.e. zsh statically determines the parent's pgrp instead of checking it via a system call)?

It's OK if the answer to this is "I have no idea." Just trying to understand it.

MikaelSmith: I suspect it's actually: `C` asks for its parent process id (`P`) and requests the process group (`A`) for that. It then calls `tcsetpgrp` or something similar in `ioctl` (https://github.com/zsh-users/zsh/blob/ad9f07e66fded8a44adba15d576960cef587f9d4/Src/utils.c#L4643-L4674) to transfer the TTY to that process group. Originally that would've worked because `pgrp must be a (nonempty) process group belonging to the same session as the calling process` (https://linux.die.net/man/3/tcsetpgrp), but now `A` is actually a different process group that's no longer part of the same session as `C` (I think `setsid` does that).

Re-reading https://blog.nelhage.com/2010/01/a-brief-introduction-to-termios-signaling-and-job-control/, it's possible that there's an additional step I need to take after moving the daemon to acquire a controlling terminal.

MikaelSmith: I'm going to keep digging into this, because I think there is a little more setup missing. For example
```
USER           PID  PPID  PGID   SESS JOBC STAT   TT       TIME COMMAND
michaelsmith 25735 25734 25735      0    0 Ss     ??    0:04.33 wash
michaelsmith 25167 25166 25167      0    1 S    s001    0:00.86 -zsh
michaelsmith 25734 25167 25734      0    1 S    s001    0:00.09 wash
michaelsmith 25745 25735 25745      0    0 S+   s001    0:00.18 /bin/zsh
michaelsmith 26275 26274 26275      0    1 S    s003    0:00.71 -zsh
michaelsmith 26436 26275 26436      0    1 S    s003    0:00.12 zsh
michaelsmith 26987 26436 26987      0    1 S+   s003    0:00.11 zsh
```

This shows two process trees:
- Run `wash`: 25167 (login zsh) -> 25734 (wash) -> 25735 (sub-wash) -> 25745 (/bin/zsh)
- Run `zsh`, then `zsh` again: 26275 (login zsh) -> 26436 (zsh) -> 26987 (zsh)

Digging into `ps` (on macOS), here are some notable traits:
- everything in a login shell shares the same TTY (except sub-wash, which doesn't seem to know its TTY)
- everything has a session of 0. This seems weird, I think ps on macOS is just not capturing anything useful there.
- `/bin/zsh` has a job control count of 0; seems like it should be 1
- sub-wash is a session leader (the 2nd lower-case `s` under STAT).

MikaelSmith: Figured out session ID, `ps` on macOS doesn't really query it right. Wrote a small C program that checks it
```
printf("pid: %d, process groupd id: %d, session ID: %d\n", pid, getpgid(pid), getsid(pid));
```
and by default everything shares a single session.

What I've done in this PR seems to work the best, haven't come up with any variations. I'll try to wrap up all the explanation in some code comments.

ekinanp (approved):
The comments are not blockers.

cmd/rootMain.go
ekinanp: Re: https://github.com/puppetlabs/wash/pull/637#discussion_r354017719, you're saying that the `operation not permitted` error that you found on zsh happened because of something like:

* `P` and `C` share the same session, let's say `S1`
* `P` changes its process group to `B`
* `P` changes its session to `S2`
* `C` asks for `P`'s process group, which is `B`.
* `C` attempts to transfer the TTY to `B`, but that doesn't work because `P`'s session is now `S2` instead of `S1`

MikaelSmith: Yes. Previously they were all in the same process group, now the Wash daemon is in a different process group.

I need to specifically check that this works on Linux as well, because it's getting into some low-level classification. I haven't figured out how to have an acceptance test that exercises the interactive code.

cmd/rootMain.go
```diff
@@ -21,7 +21,16 @@ import (
 // rootCommandFlag is associated with the `-c` option of the root command, set in root.go.
 var rootCommandFlag string
 
-// Start the wash server, then present the default system shell.
+// Start the Wash server then present the default system shell. The server will be running in the
```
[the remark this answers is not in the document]
MikaelSmith: Yeah. I think I'd accidentally run my installed `wash` when I thought it worked.

cmd/rootMain.go
```diff
@@ -21,7 +21,16 @@ import (
...
+// current process, while the shell will be in a separate child process. We'd like the server to be
+// able to prompt for input without interupting the shell (as the controller of the terminal) so
+// that plugins can prompt the user for input like e.g. security tokens.
+//
+// To allow prompts, we start the shell process then put the Wash server (daemon) in a new session.
```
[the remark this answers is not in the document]
MikaelSmith: Ok, will update.

cmd/rootMain.go
MikaelSmith: Works on Linux.

MikaelSmith: Ok, the new session was the only important thing. So did find a way to simplify it, and I think the explanation is clearer now too.

cmd/rootMain.go
```diff
@@ -117,79 +145,34 @@ Try 'help'`)
...
-			}
-		} else {
-			codeCh <- 0
-		}
-		close(codeCh)
-	}()
+	// If interactive (when we might prompt the user for input, such as security tokens), create a
+	// new session. If not interactive, calling setsid is pointless and might fail.
+	if plugin.IsInteractive() {
+		if _, err := unix.Setsid(); err != nil {
```
ekinanp: This is a lot simper.

ekinanp (approved):

cmd/rootMain.go
MikaelSmith: Only worked after I started ensuring the daemon wasn't leader of a process group. Put that note in the commit message.

cmd/rootMain.go
```diff
@@ -59,6 +59,26 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
 	// Set plugin interactivity to false if execfile or rootCommandFlag were specified.
 	plugin.InitInteractive(execfile == "" && rootCommandFlag == "")
 
+	// If interactive and this process is in its own process group, then fork and run the original
+	// command as a new process that's not the leader of its process group. This is specifically to
+	// work with shells (zsh) that try to use their parent's process group, which would no longer be
+	// in the same session. By forking, we keep the original process group in its original session so
+	// the child shell can still modify it.
+	if plugin.IsInteractive() && os.Getpid() == unix.Getpgrp() {
```
ekinanp: From what I'm seeing in https://linux.die.net/man/2/setsid, could the reasoning behind this be simplified to "we ensure that the original command is not the leader of its process group so that `setsid` (which creates the new session) works"?

MikaelSmith: Kind of, bet technically the other approach would have also worked (and we wouldn't need to exec).

Issue #619, opened 2019-11-25
Clearer interface for Read primitives


MikaelSmith:
Reads from FUSE always operate on a length and offset within a file. This makes the simplest interface to adapt the `io.ReaderAt` interface. It also maps well to some modes of object access like AWS S3 and Google Cloud Storage APIs, or using `dd` to access parts of a file.

However other things we want to read don't have a pre-determined size, and we have to download the entire thing to find out the size. In those cases we want to save the content when we download it for later block reads. We want to download that data as soon as the file is accessed (`Open` is the earliest clear access) so we can update the size attribute of the file. `Release` could be used to free the buffered data.

So we want to provide a basic Read function for block access
```
type BlockReadable struct {
  Read(ctx context.Context, size int64, offset int64) ([]byte, error)
}
```

Files where you need to download the content to know the size should implement
```
type Readable struct {
  Read(context.Context) ([]byte, error)
}
```

----

External plugins have matching semantics
```
// Block-readable entries
<plugin_script> read <path> <state> <size> <offset> # data written to stdout
// All other readable entries
<plugin_script> read <path> <state> # data written to stdout
```

Here, we still preserve the `methods` array, but we add the following semantics for `read`:
* `read` can be a tuple of `[read, <block_readable>]`. If `block_readable` is true, we use the block-readable calling convention.
  * If `read` isn't a tuple, then assume `<block_readable>` is false. This avoids breaking external plugins.
* For prefetching, `[read, <content>]` we can assume that `<block_readable>` is false.

[`read` prefetching](https://puppetlabs.github.io/wash/docs/external-plugins#entry-json-object) is not supported for block-readable entries.

ekinanp:
Note that the `[read, <block_readable>]` tuple is a specialized case of the more general `[<method>, <signature>]` tuple. The latter could be useful for overloaded actions; overloading's a useful way to capture a common pattern that could be implemented in different ways.

ekinanp:
Updated the issue to change `BlockReadable`'s signature to `size, offset` b/c the latter results in a less wordy comment ("read size bits of content starting at offset")

ekinanp:
https://github.com/puppetlabs/wash/pull/643 was the last PR needed for this work.

Issue #609, opened 2019-11-21
Initial `cat` doesn't display entire content for files with unknown sizes and size > 4096


ekinanp:
Consider
```
wash . ❯ cat gcp/Wash/cloud_functions/function-1/log
LEVEL   EXECUTION_ID      TIME_UTC                         LOG
DEBUG   luob2iy1ss6g      2019-11-19T19:17:28.682161215Z   Function execution started
        luob2iy1ss6g      2019-11-19T19:17:28.688Z         2019/11/19 19:17:28 Function triggered by change to: { projects/wash-247718/databases/(default)/documents/users/bob }
        luob2iy1ss6g      2019-11-19T19:17:28.688Z         2019/11/19 19:17:28 {{<nil>} {<nil>}}
DEBUG   luob2iy1ss6g      2019-11-19T19:17:28.688574170Z   Function execution took 7 ms, finished with status: 'ok'
DEBUG   846149830131923   2019-11-19T19:36:05.815614226Z   Function execution started
        846149830131923   2019-11-19T19:36:05.883Z         2019/11/19 19:36:05 Function triggered by change to: { projects/wash-247718/databases/(default)/documents/users/bob }
        846149830131923   2019-11-19T19:36:05.884Z         2019/11/19 19:36:05 {{map[foo:map[stringValue:bar]]} {map[foo:map[stringValue:bar] baz:map[stringValue:qux]]}}
DEBUG   846149830131923   2019-11-19T19:36:05.884075444Z   Function execution took 69 ms, finished with status: 'ok'
DEBUG   850996199332111   2019-11-21T19:28:07.488097718Z   Function execution started
        850996199332111   2019-11-21T19:28:07.496Z         2019/11/21 19:28:07 Function triggered by change to: { projects/wash-247718/databases/(default)/documents/users/bob }
        850996199332111   2019-11-21T19:28:07.496Z         2019/11/21 19:28:07 {{map[baz:map[stringValue:qux] foo:map[stringValue:bar]]} {map[foo:map[stringValue:bar] a_field:map[stringValue:a_value] baz:map[stringValue:qux]]}}
DEBUG   850996199332111   2019-11-21T19:28:07.496795932Z   Function execution took 10 ms, finished with status: 'ok'
DEBUG   851070818696154   2019-11-21T20:02:46.905081137Z   Function execution started
[...]

PR #638, opened 2019-12-04
Support block-readable entries

ekinanp:
External plugins will be handled in a follow-up PR. We should also file a separate issue for the volume package; I didn't implement BlockReadable for the volume package because it isn't as easy to do as e.g. S3 objects or GCP storage objects.

Supports https://github.com/puppetlabs/wash/issues/619

### DIFF
--- a/plugin/cache.go
+++ b/plugin/cache.go
@@ -208,20 +208,51 @@ func cachedList(ctx context.Context, p Parent) (*EntryMap, error) {
 	return cachedEntries.(*EntryMap), nil
 }
 
-// CachedOpen caches a Readable object's Open method.
-// When using the reader returned by this method, use idempotent read operations
-// such as ReadAt or wrap it in a SectionReader. Using Read operations on the cached
-// reader will change it and make subsequent uses of the cached reader invalid.
-func cachedOpen(ctx context.Context, r Readable) (SizedReader, error) {
-	cachedContent, err := cachedDefaultOp(ctx, OpenOp, r, func() (interface{}, error) {
-		return r.Open(ctx)
+// cachedRead caches an entry's Read method
+func cachedRead(ctx context.Context, e Entry) (entryContent, error) {
+	cachedContent, err := cachedDefaultOp(ctx, ReadOp, e, func() (interface{}, error) {
+		switch signature := ReadAction().signature(e); signature {
+		case defaultSignature:
+			// Both external and core plugin entries that have the default Read signature
+			// implement the Readable interface, so we can go ahead and cast directly.
+			r := e.(Readable)
+			rawContent, err := r.Read(ctx)
+			if err != nil {
+				return nil, err
+			}
+			return newEntryContent(rawContent), nil
+		case blockReadableSignature:
+			// Go doesn't allow overloaded functions, so external plugin entries will
+			// implement the BlockReadable interface via the blockRead method.
+			var readFunc func(ctx context.Context, size int64, offset int64) ([]byte, error)
+			switch t := e.(type) {
+			case externalPlugin:
+				// TODO: external plugin implementation here!
+			case BlockReadable:
+				readFunc = func(ctx context.Context, size int64, offset int64) ([]byte, error) {
+					return t.Read(ctx, size, offset)
+				}
+			default:
+				// We should never hit this code-path
+				panic("attempting to retrieve the content of a non-readable entry")
+			}
+			content := newBlockReadableEntryContent(readFunc)
+			if attr := e.attributes(); attr.HasSize() {
+				content.sz = attr.Size()
+			}
+			return content, nil
+		default:
+			// We should never hit this code-path
+			msg := fmt.Sprintf("unknown signature '%v' for read", signature)
+			panic(msg)
+		}
 	})
 
 	if err != nil {
 		return nil, err
 	}
 
-	return cachedContent.(SizedReader), nil
+	return cachedContent.(entryContent), nil
 }
 
 // cachedMetadata caches an entry's Metadata method
--- a/plugin/cache.go
+++ b/plugin/cache.go
@@ -222,9 +222,11 @@ func cachedRead(ctx context.Context, e Entry) (entryContent, error) {
 			}
 			return newEntryContent(rawContent), nil
 		case blockReadableSignature:
-			// Go doesn't allow overloaded functions, so external plugin entries will
-			// implement the BlockReadable interface via the blockRead method.
-			var readFunc func(ctx context.Context, size int64, offset int64) ([]byte, error)
+			// Go doesn't allow overloaded functions, so the external plugin entry type
+			// cannot implement both BlockReadable#Read and Readable#Read. Thus, external
+			// plugins implement the BlockReadable interface via a separate blockRead
+			// method.
+			var readFunc blockReadFunc
 			switch t := e.(type) {
 			case externalPlugin:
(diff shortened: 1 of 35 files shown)


plugin/methodWrappers.go
```diff
@@ -129,14 +129,41 @@ func List(ctx context.Context, p Parent) (*EntryMap, error) {
...
+// If the offset is >= to the available content size, then data != nil, len(data) == 0,
+// and err == io.EOF. Otherwise if len(data) < size, then err == io.EOF.
+//
+// Note that Read is thread-safe.
+func Read(ctx context.Context, e Entry, size int64, offset int64) (data []byte, err error) {
+	if !ReadAction().IsSupportedOn(e) {
+		panic("plugin.Read called on a non-readable entry")
+	}
+	data = []byte{}
+	if attr := e.attributes(); attr.HasSize() {
```
ekinanp: It feels weird for a BlockReadable entry to have an unknown size attribute (why would the API let you block-read its content without telling you how big that content is?). Should we panic if we run into that case?

Reason I bring it up here is b/c if we do run into that kind of entry, then the plugin API's responsible for implementing `Read`'s semantics. `Readable` entries w/ the default signature don't need to worry about this because the `entryContentImpl` type takes care of all that.

ekinanp: I'm also thinking about changing `size int64` => `size int`. Reason is b/c `len(data)` is an int.

ekinanp: Is it worth returning `io.EOF` errors? When would it be useful for plugins to provide that information (aside from the block-readable entry w/ unknown size attribute case)? Asking b/c we ignore `io.EOF` in FUSE.

ekinanp:
I'll also look into https://github.com/puppetlabs/wash/issues/609 in a separate PR.

plugin/methodWrappers.go
MikaelSmith: I think leaving out `io.EOF` makes sense. I don't think we need it. We would potentially need it if we switched to `OpenDirectIO` mode (as noted in https://godoc.org/bazil.org/fuse/fs#HandleReader), which could help with #609.

plugin/cache.go
```diff
@@ -208,20 +208,51 @@ func cachedList(ctx context.Context, p Parent) (*EntryMap, error) {
...
+			r := e.(Readable)
+			rawContent, err := r.Read(ctx)
+			if err != nil {
+				return nil, err
+			}
+			return newEntryContent(rawContent), nil
+		case blockReadableSignature:
+			// Go doesn't allow overloaded functions, so external plugin entries will
+			// implement the BlockReadable interface via the blockRead method.
+			var readFunc func(ctx context.Context, size int64, offset int64) ([]byte, error)
```
MikaelSmith: Maybe make this a type if we're going to use it here and in the `newBlockReadableEntryContent` call.

plugin/cache_test.go
```diff
@@ -369,13 +373,92 @@ func (suite *CacheTestSuite) TestCachedListSetEntryID() {
...
+		cacheTestsMockEntry: newCacheTestsMockEntry("foo"),
+	}
+	entry.DisableDefaultCaching()
+	entry.SetTestID("/foo")
+
+	ctx := context.Background()
+	expectedRawContent := []byte("some raw content")
+	entry.On("Read", ctx, int64(10), int64(0)).Return(expectedRawContent, nil).Once()
+
+	content, err := cachedRead(ctx, entry)
```
MikaelSmith: I feel like somewhere in here we should be verifying that the length of content we got back matched the requested length.

MikaelSmith: Or at least actualLength <= requestedLength. Because we're probably going to truncate anything longer than the requested length.

MikaelSmith (changes requested):
Just a few minor suggestions.

plugin/cache.go
```diff
@@ -208,20 +208,51 @@ func cachedList(ctx context.Context, p Parent) (*EntryMap, error) {
...
+			// Both external and core plugin entries that have the default Read signature
+			// implement the Readable interface, so we can go ahead and cast directly.
+			r := e.(Readable)
+			rawContent, err := r.Read(ctx)
+			if err != nil {
+				return nil, err
+			}
+			return newEntryContent(rawContent), nil
+		case blockReadableSignature:
+			// Go doesn't allow overloaded functions, so external plugin entries will
```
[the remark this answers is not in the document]
ekinanp: What do you mean?

plugin/cache.go
ekinanp: Like `type readFunc = func(ctx context.Context, size int64, offset int64) ([]byte, error)` ?

MikaelSmith: Yes.

plugin/cache.go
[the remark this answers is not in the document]
MikaelSmith: I mean, is it possible for a type to implement both `Read` signatures for `Readable` and `BlockReadable`? If so, then I don't think I understand your code comment.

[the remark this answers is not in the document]
ekinanp: Ah ok. Yeah, that is impossible. I'll further clarify the comment.

plugin/cache.go
ekinanp: Ok

ekinanp:
Updated to ignore `io.EOF` errors, and to also address the current feedback.

plugin/cache_test.go
ekinanp: Not sure what you mean. I ended up adding a test-case for `content.size`.

I'm thinking any `actualLength <= requestedLength` checks should be done in `plugin.Read`.

MikaelSmith: Yes, I think you're right.

MikaelSmith: My main observation in the test was that you requested a read of size 10, but get back content longer than that. And nothing fails.

plugin/methodWrappers.go
MikaelSmith: Oh, I also think we should require a BlockReadable entry to set the size attribute.

MikaelSmith (approved):

ekinanp:
Going to add a bit more validation to `plugin.Read` before merging this.

ekinanp:
Updated to restore `io.EOF` and to add some more validation to `plugin.Read`.

plugin/methodWrappers.go
```diff
@@ -153,6 +162,8 @@ func Read(ctx context.Context, e Entry, size int64, offset int64) (data []byte,
 			size = contentSize
 			err = io.EOF
 		}
+	} else if ReadAction().signature(e) == blockReadableSignature {
+		activity.Warnf(ctx, "size attribute not set for block-readable entry %v", e.id())
```
ekinanp: A warning seems fine since this will likely come up when the plugin author's testing their plugin.

plugin/methodWrappers.go
```diff
@@ -163,6 +174,9 @@ func Read(ctx context.Context, e Entry, size int64, offset int64) (data []byte,
 	if readErr != nil {
 		err = readErr
 	}
+	if actualSize := int64(len(data)); actualSize > size {
+		return nil, fmt.Errorf("requested %v bytes (input was %v), but plugin's API returned %v bytes", size, inputSize, actualSize)
```
ekinanp: This should also come up during plugin author's testing.

MikaelSmith (approved):

PR #640, opened 2019-12-05
Use os-package signals rather than syscall

MikaelSmith:
Removes a direct dependency that can be platform-dependent in favor of the `os` abstraction for the same signal types.


ekinanp (approved):

MikaelSmith:
Oh, Kill isn't the same as SIGTERM. Nevermind.

PR #642, opened 2019-12-05
Cleanup from Read changes

MikaelSmith:
`SizedReader` is no longer needed, and api.fs was missed in the work. Add static assertions on the intended types for the api.fs resources to catch this in the future.


ekinanp (approved):
Ah, thanks.

PR #644, opened 2019-12-09
Fix CI

MikaelSmith:
Previous PR was merged before CI finished, and failed to cleanup some unused test definitions that relied on `SizedReader`.


ekinanp (approved):
Ah I didn't rebase master, that's why I didn't run into this when running `go test`.

PR #649, opened 2019-12-10
Change `wash version` => `wash --version`, add `verify-install` flag

ekinanp:


ekinanp:
I'll update the website after this stuff is released so that new users won't be confused.

cmd/root.go
```diff
@@ -142,16 +142,17 @@ then starts your system shell with shortcuts configured for wash subcommands.`,
 		// Omit server from embedded cases because a daemon is already running.
 		addServerArgs(rootCmd, "warn")
 		addCommand(rootCmd, serverCommand())
-		// rootCommandFlag is used in rootMain.go.
+		// The *Flag variables are declared in rootMain.go
```
MikaelSmith: `root*Flag` would be more precise.

cmd/rootMain.go
```diff
@@ -59,6 +67,18 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
 	}
 	defer os.RemoveAll(rundir)
 
+	socketpath := filepath.Join(rundir, "api.sock")
+
+	if rootVerifyInstallFlag {
+		srv := server.ForVerifyInstall(mountpath, socketpath)
+		if err := srv.Start(); err != nil {
+			cmdutil.ErrPrintf("Install verification failed: %v\n", err)
```
MikaelSmith: I think we should also print something when it succeeds (on stdout). After stopping the server.

cmd/internal/server/core.go
```diff
@@ -113,19 +135,21 @@ func (s *Server) Start() error {
 	}
 	s.fuse = controlChannels{stopCh: fuseServerStopCh, stoppedCh: fuseServerStoppedCh}
 
-	if s.opts.CPUProfilePath != "" {
-		f, err := os.Create(s.opts.CPUProfilePath)
-		if err != nil {
-			log.Fatal(err)
+	if !s.forVerifyInstall {
+		if s.opts.CPUProfilePath != "" {
```
MikaelSmith: I think we should leave the CPUProfilePath stuff in. It might be interesting to test just that portion of the server, and there's no harm in having it still work.

cmd/internal/server/core.go
```diff
@@ -148,6 +172,10 @@ func (s *Server) stopFUSEServer() {
 }
 
 func (s *Server) shutdown() {
+	if s.forVerifyInstall {
+		return
+	}
+
 	if s.opts.CPUProfilePath != "" {
 		pprof.StopCPUProfile()
 	}
```
MikaelSmith: Wouldn't hurt to still make sure we flush open journals.

MikaelSmith (changes requested):

cmd/internal/server/core.go
ekinanp: Deferring that to when we need it, since `verify-install` (currently) has no way to set the CPU profile.

cmd/internal/server/core.go
ekinanp: Deferring that to when we need it since nothing should be printed to journals during `verify-install`

ekinanp:
Updated to address the feedback (unless otherwise noted).

MikaelSmith (approved):

MikaelSmith:
This should probably have an issue associated with it.

PR #651, opened 2019-12-10
Small fixes

### LINKED ISSUE #650: Opening file with MacVim doesn't create history entry
MacVim isn't added to history when I open a file with it. For example, when the file changes and gets reloaded, I can't find that anywhere in `whistory`.

MikaelSmith:
Fixes #650, #652, and some minor issues that aren't really user-facing.

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -41,13 +41,22 @@ type octetResponse struct {
 	Reader io.Reader
 }
 
-type handler func(http.ResponseWriter, *http.Request) *errorResponse
+type handler struct {
+	fn      func(http.ResponseWriter, *http.Request) *errorResponse
+	logOnly bool
+}
 
 func (handle handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	activity.Record(r.Context(), "API: %v %v", r.Method, r.URL)
+	var record func(msg string, a ...interface{})
+	if handle.logOnly {
+		record = log.Printf
+	} else {
+		record = func(msg string, a ...interface{}) { activity.Record(r.Context(), msg, a...) }
+	}
+	record("API: %v %v", r.Method, r.URL)
 
-	if err := handle(w, r); err != nil {
-		activity.Record(r.Context(), "API: %v %v: %v", r.Method, r.URL, err)
+	if err := handle.fn(w, r); err != nil {
+		record("API: %v %v: %v", r.Method, r.URL, err)
 		w.WriteHeader(err.statusCode)
 
 		// NOTE: Do not set these headers in the middleware because not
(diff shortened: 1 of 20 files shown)


cmd/rootMain.go
```diff
@@ -43,6 +43,34 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
 	// Configure logrus to emit simple text
 	log.SetFormatter(&log.TextFormatter{DisableTimestamp: true})
 
+	var execfile string
```
ekinanp: Will this cause issues with `verify-install`?

MikaelSmith: Still working that out.

MikaelSmith: I'm trying to avoid creating temp directories before we fork, but don't fork for verify install. The only thing that's changed is that `plugin.IsInteractive` will now be false instead of true when running `--verify-install`.

MikaelSmith: We shouldn't have any interactive behavior during `--verify-install`, so I think that's fine.

cmd/rootMain.go
```diff
@@ -104,7 +104,7 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
 			return exitCode{1}
 		}
 		srv.Stop()
-		cmdutil.Printf("Verify install succeeded")
+		cmdutil.Printf("Verify install succeeded\n")
```
ekinanp: Hmm, my shell made it look like the newline was there. Weird.

cmd/rootMain.go
ekinanp: Yeah, that's fine. May want to update the comment on Line 51 though.

MikaelSmith: Done.

api/server.go
```diff
@@ -41,13 +41,22 @@ type octetResponse struct {
 	Reader io.Reader
 }
 
-type handler func(http.ResponseWriter, *http.Request) *errorResponse
+type handler struct {
+	fn     func(http.ResponseWriter, *http.Request) *errorResponse
+	record bool
```
MikaelSmith: This is hiding too much, because most things don't set it (so it defaults to false). Going to switch to something that you can opt-in to.

cmd/rootMain.go
```diff
@@ -43,6 +43,36 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
...
 
+	var execfile string
+	if len(args) > 0 {
+		execfile = args[0]
+	}
+
+	// Set plugin interactivity to false if execfile or rootCommandFlag were specified.
+	// Also set to false if rootVerifyInstallFlag so we don't try to fork right before running
+	// verification; verify-install should have no interactive behavior.
+	plugin.InitInteractive(execfile == "" && rootCommandFlag == "" && !rootVerifyInstallFlag)
```
ekinanp: Sorry, one other nit. The comment talks about the "false" conditions, but the code talks about the "true" conditions. This creates a mis-match in my head when I'm trying to understand why this is correct. Could the comment instead talk about the "true" condition? So something like
```
interactivity is true if a script (execfile) isn't specified, a command isn't specified, and verifyInstall is false
```

MikaelSmith: Done.

ekinanp (approved):

PR #655, opened 2019-12-11
Retry unmount if it failed due to an open file handle

### LINKED ISSUE #654: Wash requires manual intervention even after closing open file
When exiting Wash with a file opened in an editor, I encounter
```
WARN FUSE: Shutdown failed: unmount /Users/michaelsmith/Library/Caches/wash/mnt962968583: resource busy
WARN FUSE: Manual cleanup required: umount /Users/michaelsmith/Library/Caches/wash/mnt962968583
```

Even after I close the open file, it just waits for me to unmount. It shouldn't require explicitly running `umount`.

MikaelSmith:
Rather than wait for the user to run `umount`, retry unmounting  in case the file handle is released.

Fixes #654.

### DIFF
--- a/fuse/core.go
+++ b/fuse/core.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/user"
 	"strconv"
+	"strings"
 	"time"
 
 	"bazil.org/fuse"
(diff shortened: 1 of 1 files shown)


fuse/core.go
```diff
@@ -244,8 +244,14 @@ func ServeFuseFS(
 
 			log.Infof("FUSE: Unmounting %v", mountpoint)
 			if err = fuse.Unmount(mountpoint); err != nil {
-				log.Warnf("FUSE: Shutdown failed: %v", err.Error())
+				log.Warnf("FUSE: Shutdown failed: %v", err)
 				log.Warnf("FUSE: Manual cleanup required: umount %v", mountpoint)
+
+				// Retry in a loop until no longer blocked buy an open handle.
+				// All errors are `*os.PathError`, so we just match a known error string.
+				for ; err != nil && err.Error() == "resource busy"; err = fuse.Unmount(mountpoint) {
```
ekinanp: Is "resource busy" platform agnostic? Also, what are the chances that this could be an infinite loop (i.e. should there be a counter)?

ekinanp (approved):
LGTM, just a comment.

fuse/core.go
MikaelSmith: I'm not sure about platform agnostic, I'll try it out on a linux machine.

I'm pretty sure it's not an infinite loop. Once the fs is unmounted, the error message changes to `invalid argument`.

MikaelSmith: Ah, it's `Resource busy` on Linux :)

MikaelSmith: Updated so it now works on Linux and macOS.

ekinanp: Re: infinite loop, I'm mostly worried about those open handles never closing so resource busy is always returned. Sounds like that's not an issue?

MikaelSmith: Well, the current behavior is it doesn't exit in that case either.

ekinanp: Maybe a warning could be useful? Least let the user know that something went wrong with the unmount?

ekinanp (approved):

fuse/core.go
MikaelSmith: That's what happens 2 lines up.

PR #659, opened 2019-12-13
Prompt the user for a list of enabled core plugins

### LINKED ISSUE #455: AWS errors when you haven't configured it
When I start Wash and haven't used AWS before, I get a warning
```
WARN aws failed to load: could not load any profiles: the /home/casey/.aws/credentials file does not exist 
```

It's not clear what I should do to address that warning. Since I've never used AWS, it's weird that I'd get the warning in the first place.

GCP does the same thing.

Originally seen in #432.

----

Proposal: For a first-time user (someone who doesn't have the `plugins` OR `external-plugins` key defined in their `wash.yaml` file), create the following UX:

```
./wash
Do you use docker (y/n)? y
kubernetes? y
aws? n
gcp? y
The docker, kubernetes, and gcp plugins have been enabled.
You can disable them by modifying the 'plugins' key in your 'wash.yml'
file, then restarting the shell.

Welcome to Wash!
  Wash includes several built-in commands: wexec, find, list, meta, tail.
  See commands run with wash via 'whistory', and logs with 'whistory <id>'.
Try 'help'
wash . >
```

ekinanp:
Resolves #455

### DIFF
--- a/cmd/rootMain.go
+++ b/cmd/rootMain.go
@@ -126,7 +126,7 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
 
 	if plugin.IsInteractive() {
 		cmdutil.Println(`Welcome to Wash!
-  Wash includes several built-in commands: wexec, find, list, meta, tail.
+  Wash includes several built-in commands: ls, wexec, find, meta, tail.
   See commands run with wash via 'whistory', and logs with 'whistory <id>'.
 Try 'help'`)
 	}
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -98,8 +98,7 @@ func serverOptsFor(cmd *cobra.Command) (map[string]plugin.Root, server.Opts, err
 
 	// Check the internal plugins
 	if viper.IsSet("plugins") || viper.IsSet("external-plugins") {
-		enabledPlugins := viper.GetStringSlice("plugins")
-		for _, name := range enabledPlugins {
+		for _, name := range viper.GetStringSlice("plugins") {
 			if plug, ok := server.InternalPlugins[name]; ok {
 				plugins[name] = plug
 			} else {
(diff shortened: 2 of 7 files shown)


ekinanp:
Here's what the new UX looks like (including load errors) for a first-time user
```
Eniss-MacBook-Pro:wash enisinan$ ./wash
Do you use aws?: y
docker?: y
gcp?: y
kubernetes?: n
The aws, docker, gcp core plugins have been enabled.
You can disable them by modifying the 'plugins' key in your config
file (~/.puppetlabs/wash/wash.yaml), and then restarting the shell

WARN aws failed to load: failed to read /Users/enisinan/.aws/credentials: unclosed section: [enis
WARN gcp failed to load: google: error getting credentials using GOOGLE_APPLICATION_CREDENTIALS environment variable: invalid character '"' after object key:value pair
WARN Wash will still expose 'stub' plugin roots for aws, gcp so that you can view their documentation via 'docs <plugin>' (e.g. 'docs aws'). However, you'll need to exit then restart the shell in order to properly reload a failed plugin.
Welcome to Wash!
  Wash includes several built-in commands: wexec, find, list, meta, tail.
  See commands run with wash via 'whistory', and logs with 'whistory <id>'.
Try 'help'
wash . ❯ ls
aws/
docker/
gcp/
```

cmd/util/prompt.go
```diff
@@ -0,0 +1,31 @@
...
+// YesOrNoP parses input representing confirmation. confirmed (bool) is
+// true if the input starts with "y" or "Y".
+var YesOrNoP InputParser = func(input string) (confirmed interface{}, err error) {
+	confirmed = len(input) > 0 && (input[0] == 'y' || input[0] == 'Y')
+	return
+}
+
+// Prompt prints the supplied message, waits for input on stdin,
+// then passes the input over to the supplied parser.
+func Prompt(msg string, parser InputParser) (interface{}, error) {
```
ekinanp: Having this take a function makes it easier to reuse input parsers without having to create functions like "PromptYesOrNo".

cmd/server.go
```diff
@@ -103,31 +94,90 @@ func serverOptsFor(cmd *cobra.Command) (map[string]plugin.Root, server.Opts, err
...
-		return nil, server.Opts{}, fmt.Errorf("failed to unmarshal the external-plugins key: %v", err)
-	}
-
-	// Load internal plugins that are not specifically excluded.
-	enabledPlugins := viper.GetStringSlice("plugins")
 	plugins := make(map[string]plugin.Root)
-	if len(enabledPlugins) > 0 {
+
+	// Check the internal plugins
+	if viper.IsSet("plugins") || viper.IsSet("external-plugins") {
```
ekinanp: External plugins were added here b/c it is possible (and makes sense) for someone to load only external plugins

cmd/rootMain.go
```diff
@@ -126,7 +126,7 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
 
 	if plugin.IsInteractive() {
 		cmdutil.Println(`Welcome to Wash!
-  Wash includes several built-in commands: wexec, find, list, meta, tail.
+  Wash includes several built-in commands: ls, wexec, find, meta, tail.
```
ekinanp: I think we should change this to something like
```
Welcome to Wash!
  Type 'help' to view the available commands, 'whistory' to see the commands run,
  and 'whistory <id>' to see any logs associated with a specific command.
Try 'winfo .'!
```

We can then extend `winfo <entry>` (and `docs <entry>`) to look something like
```
Path: /Users/enisinan/Library/Caches/wash/mnt820574140/aws
Name: aws
CName: aws
Actions:
- list (lets you do 'cd <path>', 'ls <path>', and 'tree <path>')
- exec (lets you do 'wexec <path> <command> <args>')
- read (lets you do 'cat <path>', 'grep <regex> <path>', 'less <path>')
...
Attributes: {}
```

This way, the user
* Becomes familiar w/ `winfo` as the main tool for seeing what you can do w/ an entry
* Figures out that they can `ls` the cwd `.`
* Will likely try `winfo` on the plugins themselves, and `cd`/`ls` into them (and repeat this process for the other entries that they can find).

cmd/validate.go
```diff
@@ -47,6 +47,8 @@ method.`,
 }
 
 func validateMain(cmd *cobra.Command, args []string) exitCode {
+	plugin.InitInteractive(false)
+
 	// Validate that 'plugin' is a valid plugin and load it
 	plugins, serverOpts, err := serverOptsFor(cmd)
```
ekinanp: I'm not sure why `validate` needs all of `wash server`'s arguments.

cmd/rootMain.go
ekinanp: Hmm, `Try 'ls'` also reads well and seems natural.

I guess my main concern is exposing the fundamentals of interacting with Wash as early as possible to the user. Everything derives from `winfo` (once we update actions to also list the commands that they work with), which is why I opted for `Try 'winfo'`

ekinanp: Main theme here is to try to capture as much of "here are all of the things you can do with Wash" as possible in the main UX so first-time users can immediately dive into the tool. Having people use `winfo` first is a good way of doing that.

I think this new `Actions` output could be verbose (and annoying for people already familiar w/ the actions), so may be worth displaying it via `winfo --verbose`/`winfo -v` ?

ekinanp: Also wondering if `winfo` => `winfo .` (maybe `winfo $W`)? Latter makes it clear that `winfo` takes a path.

cmd/validate.go
MikaelSmith: It was an easy way to get access to things we use later. Probably doesn't need everything, but the extra stuff probably doesn't hurt. I'd just look at what's actually used from `plugins` and `serverOpts`. We want `plugins` for the list of available plugins and access to their Root interfaces. We want `serverOpts` for plugin config and whatever `SetupLogging` does.

cmd/server.go
```diff
@@ -103,31 +94,99 @@ func serverOptsFor(cmd *cobra.Command) (map[string]plugin.Root, server.Opts, err
...
+		for name, plug := range server.InternalPlugins {
 			plugins[name] = plug
 		}
+	} else {
+		// Assume first-time user. First, we prompt them to get a list
+		// of plugins that they wish to enable. Next, we write the
+		// enabled plugins back to their specified config file.
+		enabledPlugins := []string{}
+
+		// Prompt them for the list of enabled plugins. This should look something
```
MikaelSmith: Each of these steps would probably be nice as a separate function we could try to mock and test.

MikaelSmith (changes requested):

cmd/validate.go
ekinanp: Ah ok. I'll add a comment explaining what you just said.

cmd/server.go
ekinanp: Sure, I can break it out.

cmd/util/prompt.go
```diff
@@ -0,0 +1,31 @@
...
+	stderrMux.Lock()
+	defer stderrMux.Unlock()
+
+	var input string
+	fmt.Fprintf(Stderr, "%s: ", msg)
```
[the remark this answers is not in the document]
ekinanp: Agreed. Easiest is to probably go with `%s ` so people can include `:` in the message itself.

cmd/rootMain.go
ekinanp: Hmm, I like your idea of a "default, user-friendly" format, and making the structured YAML/JSON output optional (and keep that structured output as it currently is). We could have that "user-friendly" format direct people to "docs" if they'd like more detailed info about the entry, so that way `winfo` also mentions `docs`. I'll think a bit more about how that user-friendly format should look, then file an issue.

Agreed on `Name` and `CName`, though now that the structured output is optional, maybe we can keep as-is?

Re: `winfo .`, agreed. Could probably do that after all the "Make winfo the entry-point to Wash UX" work.

ekinanp:
Sorry, pushed something up on accident. Not yet done addressing feedback.

ekinanp:
Ok, updated to address the feedback.

MikaelSmith (approved):

Issue #662, opened 2019-12-14
`winfo` outputs keys in different order when given multiple paths (YAML format)


ekinanp:
For example,

```
wash . ❯ winfo aws
Name: aws
CName: aws
Actions:
- list
Attributes: {}
wash . ❯ winfo aws docker
aws:
  Actions:
  - list
  Attributes: {}
  CName: aws
  Name: aws
docker:
  Actions:
  - list
  Attributes: {}
  CName: docker
  Name: docker
```

The latter output should have the same order, i.e. `Name`, `CName`, `Actions`, `Attributes` but it doesn't.

Issue #661, opened 2019-12-13
`winfo` should default to cwd and omit the path (because it's redundant)


ekinanp:
Defaulting to cwd enables something like `winfo` to just work, while omitting the outputted path means the output would look something like
```
wash . ❯ winfo
Name: /
CName: '#'
Actions:
- list
Attributes: {}
```

PR #663, opened 2019-12-14
Generalize goyaml.Marshaler + some winfo improvements/fixes

### LINKED ISSUE #483: The textBuilder does not handle objects that aren't maps, arrays or primitive values
If I pass in an arbitrary type to [textBuilder](https://github.com/puppetlabs/wash/blob/master/cmd/util/marshal.go#L79-L117) that isn't an object, array, or primitive value, then https://github.com/puppetlabs/wash/blob/master/cmd/util/marshal.go#L115 gets printed, which isn't what I want. This issue's the reason why `winfo --output text` doesn't work as intended.

ekinanp:
Relevant `winfo` issues are #662 and #661. Generalizing goyaml.Marshaler fixes #483

### DIFF
--- a/cmd/info.go
+++ b/cmd/info.go
@@ -41,7 +41,7 @@ func infoMain(cmd *cobra.Command, args []string) exitCode {
 
 	// Use a sorted map so that we can control how the information's
 	// displayed.
-	infoMap := make(map[string]orderedMap)
+	infoMap := infoResultMap{}
 
 	// Fetch the data
 	ec := 0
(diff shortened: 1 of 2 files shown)


cmd/info.go
```diff
@@ -58,7 +60,6 @@ func infoMain(cmd *cobra.Command, args []string) exitCode {
 			}
 
 			entryMap := orderedMap{linkedhashmap.New()}
-			entryMap.Put("Path", entry.Path)
```
MikaelSmith: We can have multiple entries. Path seems like it could still be useful when we printing more than one entry.

MikaelSmith (changes requested):

cmd/info.go
ekinanp: The output will be a hash of `<path> => <info>`.

MikaelSmith (approved):

Issue #438, opened 2019-07-31
Remove the apifs package


ekinanp:
This package is an adapter for Wash to support local files/directories; it preceded the shell experience. Now that the shell experience is in place, we no longer need this package (the shell experience is supposed to keep users within the mountpoint so they should never interact with local files/dirs).

MikaelSmith:
We discussed this a bit more. There are a couple things going on here that are useful context, although it's probably fine still to remove this code.

Wash's `list` and `tail` commands currently work with all local files. However, navigating to local files from Wash is confusing; it breaks out of the Wash conceptual context of remote systems. Most other commands don't really have useful behavior outside of Wash. So it's probably not important to keep this.

However, one use case that does make sense is copying files from a remote system to the local system. Wash doesn't have a clear way to do that. Copying to `~/some_dir` works now, and will continue to work after removing this. However that still seems to confuse people. Once we add more write operations to the plugin system, it may be useful to model the local filesystem as a plugin at `/local` in Wash's shell.

ekinanp:
@MikaelSmith One thing I was thinking is to move apifs code over to the client. So have a wrapper around the client that returns the right API responses when given a local file path (we can assume the latter if we get an entry-not-found error). That would also make everything work with local files in the remote server case (which is hard to do with a `/local` plugin). I could file a separate issue for that work.

PR #664, opened 2019-12-16
Implement updated Writable spec

### LINKED ISSUE #620: General Write primitive
We added a `Write` primitive to Wash to implement sending a message to a message queue. We'd like to generalize that to be able to make changes to any remote file.

We have a couple use cases in mind for writing:
- I'm operating on a file through an actual filesystem. It makes sense to mirror Open, WriteAt, Sync, Close so we get consistent behavior.
- Writing to an entry with no persistent content like a message queue. This just requires a write; it can mirror WriteAt and ignore the specified offset.
- Update an object store. Most of these replace the whole object, so we want to keep a local copy and only update when it would be synchronized so that we avoid uploading new copies more than necessary.

The file lifecycle has several points that could be useful
- Open: buffer data to compute its size, open a file descriptor. Additionally we need to take into account whether an open for write is truncating the file.
- ReadAt: read a block of data from a file or data source.
- WriteAt: write a block of data to a file or data source.
- Sync: batch buffered writes.
- Close: release a file descriptor or cached copy.
- Setattr: update the size of a file.

We have several things we expect to work:
- `echo 'some text' >> file` appends to the file. If that requires rewriting the file, it should do so with a single upload.
- `echo 'some text' > file` overwrites the file.
- Editing a file should only result in uploading anything when you explicitly Save the file.

----

Implement support for replacing a file. Plugins implement the `Writable` interface as
```
Write(context.Context, []byte) error
```

Wash then aggregates all file updates, and calls `Write` when changes are flushed to the filesystem.

External plugins:
```
<plugin_script> write <path> <state> # data read from stdin
```

### LINKED ISSUE #656: Linux VIM/VSCode doesn't work with files with unknown content size
These applications interpret the file as an empty file thanks to https://github.com/puppetlabs/wash/issues/609. However, this isn't urgent since other tools like `cat`/`less`/`grep` still work, and those kinds of files aren't (commonly) editable.

MikaelSmith:
Implement the new Writable interface (defined to mirror the Readable interface). Implement Write on simple file interfaces: local, AWS S3, and Google Cloud Storage.

Resolves #620. Also fixes #656.

### DIFF
--- a/api/fs/root.go
+++ b/api/fs/root.go
@@ -1,0 +1,52 @@
+package apifs
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/puppetlabs/wash/plugin"
+)
+
+// Root of the local filesystem plugin
+type Root struct {
+	dir
+}
+
+// Init for root
+func (r *Root) Init(cfg map[string]interface{}) error {
+	var basepath string
+	if pathI, ok := cfg["basepath"]; ok {
+		basepath, ok = pathI.(string)
+		if !ok {
+			return fmt.Errorf("local.basepath config must be a string, not %s", pathI)
+		}
+	} else {
+		basepath = "/tmp"
+	}
+
+	finfo, err := os.Stat(basepath)
+	if err != nil {
+		return err
+	}
+
+	r.dir.fsnode = newFSNode(finfo, basepath)
+
+	r.EntryBase = plugin.NewEntry("local")
+	r.DisableDefaultCaching()
+	return nil
+}
+
+// Schema returns the root's schema
+func (r *Root) Schema() *plugin.EntrySchema {
+	return plugin.
+		NewEntrySchema(r, "local").
+		SetDescription(rootDescription).
+		IsSingleton()
+}
+
+var _ = plugin.Root(&Root{})
+
+const rootDescription = `
+This plugin exposes part of the local filesystem. The path is mounted based on the path specified
+in the WASH_LOCALFS environment variable.
+`
--- a/cmd/internal/server/core.go
+++ b/cmd/internal/server/core.go
@@ -23,6 +23,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// InternalPlugins lists the plugins enabled by default in Wash.
 var InternalPlugins = map[string]plugin.Root{
 	"aws":        &aws.Root{},
 	"docker":     &docker.Root{},
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -264,9 +264,12 @@ echo "Hello, world!"
 ### write
 The `write` action lets you write data to an entry. Thus, any command that writes a file also works with these entries.
 
-Note that Wash distinguishes between file-like and non-file-like entries. An entry is file-like if it's readable and writable and defines its size; you can edit it like a file. If it doesn't define a size then it's non-file-like, and trying to open it with a ReadWrite handle will error; reads from it may not return data you previously wrote to it.
+Note that Wash distinguishes between file-like and non-file-like entries. An entry is file-like if it's readable and writable and defines its size; you can edit it like a file.
+
+If it doesn't define a size then it's non-file-like, and trying to open it with a ReadWrite handle will error; reads from it may not return data you previously wrote to it. You should check its documentation with the `docs` command for that entry's write semantics. We also recommend not using editors with these entries to avoid weird behavior.
 
 #### Examples
+Modifying a file stored in Google Cloud Storage
 ```
 wash . ❯ echo 'exit 1' >> gcp/Wash/storage/some-wash-stuff/an\ example\ folder/static.sh
 wash . ❯ cat gcp/Wash/storage/some-wash-stuff/an\ example\ folder/static.sh
@@ -274,6 +277,16 @@ wash . ❯ cat gcp/Wash/storage/some-wash-stuff/an\ example\ folder/static.sh
 
 echo "Hello, world!"
 exit 1
+```
+
+Writing a message to a hypothetical message queue where each write publishes a message and each read consumes a message
+```
+wash > echo 'message 1' >> myqueue
+wash > echo 'message 2' >> myqueue
+wash > cat myqueue
+message 1
+wash > cat myqueue
+message 2
 ```
 
 ### stream
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -264,6 +264,8 @@ echo "Hello, world!"
 ### write
 The `write` action lets you write data to an entry. Thus, any command that writes a file also works with these entries.
 
+Note that Wash distinguishes between file-like and non-file-like entries. An entry is file-like if it's readable and writable and defines its size; you can edit it like a file. If it doesn't define a size then it's non-file-like, and trying to open it with a ReadWrite handle will error; reads from it may not return data you previously wrote to it.
+
 #### Examples
 ```
 wash . ❯ echo 'exit 1' >> gcp/Wash/storage/some-wash-stuff/an\ example\ folder/static.sh
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -29,6 +29,7 @@ title: Docs
 * [Actions](#actions)
   * [list](#list)
   * [read](#read)
+  * [write](#write)
   * [stream](#stream)
   * [exec](#exec)
   * [delete](#delete)
@@ -245,7 +246,7 @@ gcp/Wash/storage/some-wash-stuff
 ```
 
 ### read
-The `read` action lets you view an entry’s content. Thus, any command that reads a file also works with these entries.
+The `read` action lets you read data from an entry. Thus, any command that reads a file also works with these entries.
 
 #### Examples
 ```
@@ -258,6 +259,19 @@ echo "Hello, world!"
 ```
 wash . ❯ grep "Hello" gcp/Wash/storage/some-wash-stuff/an\ example\ folder/static.sh
 echo "Hello, world!"
+```
+
+### write
+The `write` action lets you write data to an entry. Thus, any command that writes a file also works with these entries.
+
+#### Examples
+```
+wash . ❯ echo 'exit 1' >> gcp/Wash/storage/some-wash-stuff/an\ example\ folder/static.sh
+wash . ❯ cat gcp/Wash/storage/some-wash-stuff/an\ example\ folder/static.sh
+#!/bin/sh
+
+echo "Hello, world!"
+exit 1
 ```
 
 ### stream
--- a/fuse/dir.go
+++ b/fuse/dir.go
@@ -2,6 +2,7 @@ package fuse
 
 import (
 	"context"
+	"os"
 
 	"bazil.org/fuse"
 	"bazil.org/fuse/fs"
@@ -13,7 +14,7 @@ import (
 // ==== FUSE Directory Interface ====
 
 type dir struct {
-	*fuseNode
+	fuseNode
 }
 
 var _ fs.Node = (*dir)(nil)
@@ -93,4 +94,30 @@ func (d *dir) ReadDirAll(ctx context.Context) ([]fuse.Dirent, error) {
 	})
 	activity.Record(ctx, "FUSE: Listed in %v: %+v", d, res)
 	return res, nil
+}
+
+func (d *dir) Attr(ctx context.Context, a *fuse.Attr) error {
+	// FUSE caches nodes for a long time, meaning there's a chance that
+	// f's attributes are outdated. 'refind' requests the entry from its
+	// parent to ensure it has updated attributes.
+	entry, err := d.refind(ctx)
+	if err != nil {
+		activity.Warnf(ctx, "FUSE: Attr errored %v, %v", d, err)
+		return err
+	}
+	// NOTE: We could set f.entry to entry, but doing so would require
+	// a separate mutex which may hinder performance. Since updating f.entry
+	// is not strictly necessary for the other FUSE operations, we choose to
+	// leave it alone.
+
+	applyAttr(a, plugin.Attributes(entry), os.ModeDir|0550)
+	// Attr is not a particularly interesting call and happens a lot. Log it to debug like other
+	// activity, but leave it out of activity because it introduces history entries for lots of
+	// miscellaneous shell activity.
+	if d.parent == nil {
+		log.Tracef("FUSE: Attr %v", d)
+	} else {
+		log.Debugf("FUSE: Attr %v: %+v", d, *a)
+	}
+	return nil
 }
--- a/fuse/file.go
+++ b/fuse/file.go
@@ -25,6 +25,8 @@ type file struct {
 	data []byte
 	// Size for content during writing or with unspecified size attribute
 	size uint64
+	// Tracks whether we still need to query plugin.Size for the entry's size before using size
+	sizeValid bool
 }
 
 func newFile(p *dir, e plugin.Entry) *file {
@@ -78,7 +80,14 @@ func getFileMode(entry plugin.Entry) os.FileMode {
 
 var _ = fs.NodeOpener(&file{})
 
-// Open a file for reading or writing.
+// Open an entry for reading or writing. Several patterns exist for how to interact with entries.
+// - An entry that only supports Read can only be opened ReadOnly.
+// - An entry that supports both Read and Write can be opened in any mode.
+// - An entry that only supports Write can only be opened WriteOnly.
+//
+// When writing and flushing a file, we may call Read on the entry (if it supports Read) even if
+// opened WriteOnly. That only happens when performing a partial write, or when its Size attribute
+// isn't set and there are no calls to Setattr to define the file size.
 func (f *file) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.OpenResponse) (fs.Handle, error) {
 	f.mux.Lock()
 	defer f.mux.Unlock()
@@ -96,19 +105,18 @@ func (f *file) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.OpenR
 	writable := plugin.WriteAction().IsSupportedOn(f.entry)
 	switch {
 	case req.Flags.IsReadOnly() && !readable:
-		activity.Record(ctx, "FUSE: Open read-only unsupported on %v", f)
+		activity.Warnf(ctx, "FUSE: Open read-only unsupported on %v", f)
 		return nil, fuse.ENOTSUP
 	case req.Flags.IsWriteOnly() && !writable:
-		activity.Record(ctx, "FUSE: Open write-only unsupported on %v", f)
+		activity.Warnf(ctx, "FUSE: Open write-only unsupported on %v", f)
 		return nil, fuse.ENOTSUP
 	case req.Flags.IsReadWrite() && (!readable || !writable):
-		activity.Record(ctx, "FUSE: Open read-write unsupported on %v", f)
+		activity.Warnf(ctx, "FUSE: Open read-write unsupported on %v", f)
 		return nil, fuse.ENOTSUP
 	}
 
 	if req.Flags.IsReadOnly() || req.Flags.IsReadWrite() {
-		attr := plugin.Attributes(entry)
-		if !attr.HasSize() {
+		if attr := plugin.Attributes(entry); !attr.HasSize() {
 			// The entry's content size is unknown so open the file in direct IO mode. This enables FUSE
 			// to still read the entry's content so that built-in tools like cat and grep still work.
 			resp.Flags |= fuse.OpenDirectIO
@@ -119,8 +127,16 @@ func (f *file) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.OpenR
 			if err != nil {
 				return nil, err
 			}
+		} else {
+			f.size = attr.Size()
 		}
+
+		f.sizeValid = true
+	} else if !readable {
+		// Reported size is irrelevant for write-only entries.
+		f.sizeValid = true
 	}
+
 	return f, nil
 }
 
@@ -128,6 +144,7 @@ var _ = fs.HandleReleaser(&file{})
 
 func (f *file) Release(ctx context.Context, req *fuse.ReleaseRequest) error {
 	if req.ReleaseFlags&fuse.ReleaseFlush != 0 {
+		activity.Record(ctx, "FUSE: Invoking Flush for Release on %v", f)
 		err := f.Flush(ctx, &fuse.FlushRequest{
 			Header:    req.Header,
 			Handle:    req.Handle,
@@ -135,6 +152,18 @@ func (f *file) Release(ctx context.Context, req *fuse.ReleaseRequest) error {
 		})
 		if err != nil {
 			return err
+		}
+	}
+
+	if _, ok := f.writers[req.Handle]; ok {
+		delete(f.writers, req.Handle)
+
+		if len(f.writers) == 0 {
+			// If we just released the last writer, release the data buffer to conserve memory and
+			// invalidate cache on the entry and its parent so we get updated content and size on the
+			// next request. Leave size so we can use it for future attribute requests.
+			f.data = nil
+			plugin.ClearCacheFor(plugin.ID(f.entry), true)
 		}
 	}
 
@@ -187,7 +216,7 @@ func (f *file) Write(ctx context.Context, req *fuse.WriteRequest, resp *fuse.Wri
 	if newLen := int(newLen); newLen > len(f.data) {
 		f.data = append(f.data, make([]byte, newLen-len(f.data))...)
 	}
-	// Write-only entries are assumed not to have a size.
+	// Size is irrelevant for write-only entries.
 	if plugin.ReadAction().IsSupportedOn(f.entry) && f.size < uint64(newLen) {
 		f.size = uint64(newLen)
 	}
@@ -199,7 +228,7 @@ func (f *file) Write(ctx context.Context, req *fuse.WriteRequest, resp *fuse.Wri
 
 func (f *file) load(ctx context.Context, start, end int64) ([]byte, error) {
 	if !plugin.ReadAction().IsSupportedOn(f.entry) {
-		activity.Record(ctx, "FUSE: Non-contiguous writes (at %v) unsupported on %v", start, f)
+		activity.Warnf(ctx, "FUSE: Non-contiguous writes (at %v) unsupported on %v", start, f)
 		return nil, fuse.ENOTSUP
 	}
 
@@ -214,7 +243,6 @@ func (f *file) Flush(ctx context.Context, req *fuse.FlushRequest) error {
 	f.mux.Lock()
 	defer f.mux.Unlock()
 	activity.Record(ctx, "FUSE: Flush %v: %+v", f, *req)
-	var releasedWriter bool
 
 	// If this handle had an open writer, write current data.
 	if _, ok := f.writers[req.Handle]; ok {
@@ -224,28 +252,36 @@ func (f *file) Flush(ctx context.Context, req *fuse.FlushRequest) error {
 			panic("Size was not kept up-to-date with changes to data.")
 		}
 
-		if uint64(dataLen) < f.size {
-			// Missing some data, load the remainder before writing.
-			data, err := f.load(ctx, dataLen, int64(f.size))
+		if !f.sizeValid {
+			// If this is a readable entry that's opened WriteOnly and hasn't been truncated, we haven't
+			// determined its size. Get it now so we can buffer as needed.
+			size, err := plugin.Size(ctx, f.entry)
 			if err != nil {
 				return err
 			}
+			f.size = size
+			f.sizeValid = true
+		}
+
+		if uint64(dataLen) < f.size {
+			// Missing some data, load the remainder before writing.
+			data, err := f.load(ctx, dataLen, int64(f.size))
+			if err != nil && err != io.EOF {
+				return err
+			}
 			f.data = append(f.data, data...)
+
+			// If still too small then something increased the size beyond the original.
+			// Fill with null characters.
+			if sz := uint64(len(f.data)); sz < f.size {
+				f.data = append(f.data, make([]byte, f.size-sz)...)
+			}
 		}
 
 		if err := plugin.WriteWithAnalytics(ctx, f.entry.(plugin.Writable), f.data); err != nil {
+			activity.Warnf(ctx, "FUSE: Error writing %v: %v", f, err)
 			return err
 		}
-		delete(f.writers, req.Handle)
-		releasedWriter = true
-	}
-
-	if len(f.writers) == 0 && releasedWriter {
-		// Ensure data is released. Leave size so we can use it for future attribute requests.
-		f.data = nil
-
-		// Invalidate cache on the entry and its parent. Need to get updated content and size.
-		plugin.ClearCacheFor(plugin.ID(f.entry), true)
 	}
 	return nil
 }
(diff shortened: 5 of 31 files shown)


api/fs/root.go
```diff
@@ -0,0 +1,39 @@
+package apifs
```
ekinanp: Why does this need to be a plugin?

MikaelSmith:
This is still a work in progress, but I think the approach is about right. Includes a rewrite of fuse/file to broadly mimic https://github.com/bazil/bolt-mount/blob/master/file.go, as suggested in https://github.com/bazil/fuse/issues/225.

Remaining work:
- [ ] Remove or make the `local` plugin disabled by default. It was very useful for testing.
- [ ] Implement `Write` on AWS S3 objects.
- [ ] Add more tests for fuse/file.
- [ ] Revisit TODOs on reading data when file is opened for writing.

api/fs/root.go
MikaelSmith: Just added a comment addressing this and other work that still needs to be done.

ekinanp: Ah, for testing FUSE stuff. Should probably be a developer thing then, i.e. users should not be able to enable it.

plugin/types.go
```diff
@@ -195,12 +195,10 @@ type Readable interface {
 	Read(context.Context) ([]byte, error)
 }
 
-// Writable is an entry that can write data directly to the entry. It mirrors
-// the WriterAt interface with an added context for handling the lifecycle of
-// remote write operations.
+// Writable is an entry that overwrites the prior contents
```
ekinanp: Do you plan on updating the `Writable` docs separately? If not, then this should mention the last bullet point in https://github.com/puppetlabs/wash/issues/620#issuecomment-564192168 since the current comment is not true for GCP topics.

plugin/gcp/storageObject.go
```diff
@@ -45,6 +45,16 @@ func (s *storageObject) Read(ctx context.Context, size int64, offset int64) ([]b
...
 }
 
+func (s *storageObject) Write(ctx context.Context, p []byte) error {
+	wr := s.ObjectHandle.NewWriter(ctx)
+	if _, err := wr.Write(p); err != nil {
+		wr.Close()
+		return err
+	}
+
+	return wr.Close()
```
ekinanp: If `wr.Close()` fails, does that mean the data isn't written (whereby returning a Write error makes sense)? If not, then this should probably be logged to the activity journal or something similar.

cmd/internal/server/core.go
```diff
@@ -23,12 +24,19 @@ import (
...
-}
+// InternalPlugins lists the plugins enabled by default in Wash.
+var InternalPlugins = func() map[string]plugin.Root {
+	plugins := map[string]plugin.Root{
+		"aws":        &aws.Root{},
+		"docker":     &docker.Root{},
+		"gcp":        &gcp.Root{},
+		"kubernetes": &kubernetes.Root{},
+	}
+	if os.Getenv("WASH_LOCALFS") != "" {
```
ekinanp: Now that https://github.com/puppetlabs/wash/issues/455 is resolved, I think this should move over to https://github.com/puppetlabs/wash/blob/master/cmd/server.go#L137. That way, you can just say "You can enable localfs by setting WASH_LOCALFS to true" or something w/o having people mess with their `plugins` key.

fuse/dir.go
```diff
@@ -66,7 +67,9 @@ func (d *dir) Lookup(ctx context.Context, req *fuse.LookupRequest, resp *fuse.Lo
 	}
 
 	log.Debugf("FUSE: Found file %v/%v", d, cname)
-	return newFile(d, entry), nil
+	f := newFile(d, entry)
+	f.fillAttr(&resp.Attr)
```
ekinanp: What's the benefit of doing this here?

ekinanp (commented):
Still reviewing, just left a few initial comments.

plugin/gcp/pubsubTopic.go
```diff
@@ -141,14 +141,11 @@ func (t *pubsubTopic) Stream(ctx context.Context) (io.ReadCloser, error) {
 	return t.newPubsubTopicWatcher(ctx)
 }
 
-func (t *pubsubTopic) Write(ctx context.Context, _ int64, b []byte) (int, error) {
+func (t *pubsubTopic) Write(ctx context.Context, b []byte) error {
```
ekinanp: Similar Q. to https://github.com/puppetlabs/wash/pull/664#discussion_r358519024 regarding result.Get

api/fs/root.go
```diff
@@ -0,0 +1,55 @@
...
+func (r *Root) Schema() *plugin.EntrySchema {
+	return plugin.
+		NewEntrySchema(r, "local").
+		SetDescription(rootDescription).
+		IsSingleton()
+}
+
+var _ = plugin.Root(&Root{})
+
+const rootDescription = `
```
ekinanp: Per https://github.com/puppetlabs/wash/pull/664#discussion_r358520344, you could probably capture the config in the LOCALFS environment variable.

plugin/types.go
MikaelSmith: I'll make some updates. I will need to write updates for the docs related to external plugins.

plugin/gcp/storageObject.go
MikaelSmith: Yes, it means the data wasn't successfully written.

fuse/dir.go
MikaelSmith: May avoid a separate call to Attr, but it's probably not necessary.

plugin/gcp/pubsubTopic.go
MikaelSmith: Yes.

api/fs/root.go
MikaelSmith: So some of the discussion in https://github.com/puppetlabs/wash/issues/438 pointed to making apifs just another plugin. That's why I'm implementing it so closely to one.

ekinanp: You could still keep `Init` as-is, you'd just create the config in `server.go` from the env. var.

MikaelSmith: Done.

MikaelSmith:
I think this is ready for review again.

fuse/file.go
```diff
@@ -3,103 +3,295 @@ package fuse
...
+	}
 
-	// Initiate content request and return a channel providing the results.
-	if plugin.ReadAction().IsSupportedOn(updatedEntry) {
-		activity.Record(ctx, "FUSE: Opened %v", f)
-		if attr := plugin.Attributes(updatedEntry); !attr.HasSize() {
-			// The entry's content size is unknown so open the file in direct
-			// IO mode. This enables FUSE to still read the entry's content so
-			// that built-in tools like cat and grep still work.
+	if req.Flags.IsReadOnly() || req.Flags.IsReadWrite() {
```
MikaelSmith: I'm not sure these are the right conditions. We definitely don't want to call `plugin.Size` on an entry that doesn't support Read. We definitely do want the size if opening it ReadOnly, to address how Vim opens it.

If you try to do a write in the middle of a file that doesn't have a Size attribute defined, then you may end up truncating the file (because we won't realize some data is missing that we need to read before writing, see the implementation of `Flush`). So if the file implements ReadAction but is opened WriteOnly, maybe we still want to determine the size. Or maybe we want to error if you try to open a file for writing and haven't explicitly set the Size attribute.

ekinanp (commented):
Some initial comments, still reviewing

fuse/file.go
```diff
@@ -3,103 +3,295 @@ package fuse
...
+	defer f.mux.Unlock()
+	activity.Record(ctx, "FUSE: Flush %v: %+v", f, *req)
+	var releasedWriter bool
 
-func (fh fileHandle) Release(ctx context.Context, req *fuse.ReleaseRequest) error {
-	activity.Record(ctx, "FUSE: Release %v", fh.id)
-	if closer, ok := fh.r.(io.Closer); ok {
-		return closer.Close()
+	// If this handle had an open writer, write current data.
+	if _, ok := f.writers[req.Handle]; ok {
```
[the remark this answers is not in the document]
MikaelSmith: It's not uncommon to call Flush on a handle opened for reading (not sure why). It would also not have an open writer if you opened it for writing but didn't do any calls `write` calls.

fuse/file.go
```diff
@@ -3,103 +3,295 @@ package fuse
...
+		if err := plugin.WriteWithAnalytics(ctx, f.entry.(plugin.Writable), f.data); err != nil {
+			return err
+		}
+		delete(f.writers, req.Handle)
+		releasedWriter = true
 	}
 
-	if closer, ok := fh.w.(io.Closer); ok {
-		return closer.Close()
+	if len(f.writers) == 0 && releasedWriter {
```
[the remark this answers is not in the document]
MikaelSmith: I couldn't think of a way to identify when we'd closed the last open writer. But I could probably delay deleting it from `f.writers` until release as well. We might get some redundant writes if someone decides to call `flush` multiple times, but that's what they asked for.

api/fs/root.go
```diff
@@ -0,0 +1,52 @@
+package apifs
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/puppetlabs/wash/plugin"
+)
+
+// Root of the GCP plugin
```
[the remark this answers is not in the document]
MikaelSmith: The rebase was just to fix this. All other changes are in a new commit at the end.

ekinanp (commented):
Per the [conversation on Slack](https://puppetcommunity.slack.com/archives/CHMCR172S/p1576703313033800), we want a way to distinguish between `file-like` and `non-file-like` entries. 

* `File-like` entries are entries with content, such that Read returns that content and Write updates it. `Non-file-like` entries are entries without content, so that "Read" data could represent one thing and the "Written" data could represent another.

* **For plugin author docs**: `file-like` entries are entries with a `Size` attribute (implicitly entry is also Readable). So an entry has content iff it has a `Size` attribute. Thus, people should not set size for `non-file-like` entries.
    * Reason: `Size` attribute means you can filter those things with `find`. When would filtering on `Size` make sense for Docker container logs, cloud function logs, etc?
        * We could always change these semantics later to a `is-file-like` key in the entry JSON on user demand

* **Implementation**:
    * `File-like` entries are `Readable`, `Writable`, and have a `Size` attribute.
        * Reason: Up to now, we tell people “set size if you know it” for Readable entries. We also do that for things like `metadata.json` entries, which are not `file-like`. These semantics ensure that those entries don't accidentally become `file-like`.


    * For `file-like` entries, `f.data` represents the content. For `non-file-like` entries, `f.data` represents what's about to be written.
    
    * For `file-like` entries, the first `Write` call loads the entire data (because we're going to overwrite it anyways on a `Flush`). `f.data` is then updated appropriately with the written block.
    
    * `Flush` should just write whatever `f.data` is.
    
    * For `Read`, `f.data` should only be returned for `file-like` entries (and if it exists). Otherwise, we should delegate to `plugin.ReadWithAnalytics`. That's because `f.data` represents the content for `file-like` entries while for other entries, it's separate from what could be read.

ekinanp:
One example of a non-file like entry that may have Read/Write could be a GCP log, where read could return its most recent 1000 entries while write could write a new entry.

MikaelSmith:
Updated.

fuse/file.go
```diff
@@ -3,103 +3,370 @@ package fuse
...
+
+		if uint64(dataLen) < f.readSize {
+			// Missing some data, load the remainder before writing.
+			data, err := f.load(ctx, dataLen, int64(f.readSize))
+			if err != nil && err != io.EOF {
+				return err
+			}
+			f.data = append(f.data, data...)
+
+			// If still too small then something increased the size beyond the original.
```
ekinanp: When/how could this happen? Should this be a known issue/something we document?

fuse/file.go
```diff
@@ -3,103 +3,370 @@ package fuse
...
+
+	if req.Valid.Size() {
+		if !req.Valid.Handle() {
+			// No guarantee we'll ever write the change. If this is ever necessary, we could update it
+			// to immediately do a plugin.Write.
+			return fuse.ENOTSUP
+		}
+
+		// Ensure handle is in list of writers.
+		f.writers[req.Handle] = struct{}{}
```
ekinanp: Might be helpful to mention somewhere that this will likely be (only) called to truncate the file for writes. That helps clarify why we're including the handle in the list of writers.

fuse/file.go
```diff
@@ -3,103 +3,370 @@ package fuse
...
+	activity.Record(ctx, "FUSE: Attr %v: %+v", f, *a)
+	return nil
 }
 
-// Open a file for reading or writing.
+func (f *file) fillAttr(a *fuse.Attr) {
+	attr := plugin.Attributes(f.entry)
+	applyAttr(a, attr, getFileMode(f.entry))
+
+	if f.useLocalState() || !attr.HasSize() {
```
ekinanp: I'd mention that `!attr.HasSize()` should only be true for non-file-like entries

fuse/file.go
```diff
@@ -3,103 +3,370 @@ package fuse
...
+	if f.isFileLike() && f.readSize < uint64(newLen) {
+		f.readSize = uint64(newLen)
+	}
+
+	resp.Size = copy(f.data[req.Offset:], req.Data)
+	activity.Record(ctx, "FUSE: Write %v/%v bytes starting at %v from %v", resp.Size, len(req.Data), req.Offset, f)
+	return nil
+}
+
+func (f *file) load(ctx context.Context, start, end int64) ([]byte, error) {
```
ekinanp: This should probably panic if called on a non-file-like entry.

fuse/file.go
```diff
@@ -3,103 +3,370 @@ package fuse
...
+	applyAttr(a, attr, getFileMode(f.entry))
+
+	if f.useLocalState() || !attr.HasSize() {
+		// Use whatever size we know locally. Retrieving content can be expensive so we settle for
+		// including size only when it's been retreived previously by opening the file.
+		a.Size = f.readSize
+	}
+}
+
+func getFileMode(entry plugin.Entry) os.FileMode {
```
ekinanp: Nit: `getFileMode` => `defaultMode`. Latter matches the corresponding param. in applyAttr and it seems to be what this is calculating.

fuse/file.go
```diff
@@ -3,103 +3,370 @@ package fuse
...
-			// The entry's content size is unknown so open the file in direct
-			// IO mode. This enables FUSE to still read the entry's content so
-			// that built-in tools like cat and grep still work.
-			resp.Flags |= fuse.OpenDirectIO
+	f.mux.Lock()
+	defer f.mux.Unlock()
+	activity.Record(ctx, "FUSE: Open %v: %+v", f, *req)
+
+	if !f.useLocalState() {
+		// Check for an updated entry in case it has static state.
```
ekinanp: Why do we need to refresh the entry here?

fuse/file.go
```diff
@@ -3,103 +3,370 @@ package fuse
...
+	return &file{fuseNode: newFuseNode("f", p, e), writers: make(map[fuse.HandleID]struct{})}
+}
+
+func (f *file) isFileLike() bool {
+	attr := plugin.Attributes(f.entry)
+	return attr.HasSize()
+}
+
+// If currently writing a file-like object, we should use local state to fulfil many requests.
+func (f *file) useLocalState() bool {
```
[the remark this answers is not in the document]
ekinanp: I'd also comment that this returning false means "entry is not file-like OR we're not writing to it". Helps clarify some of the `!useLocalState` uses elsewhere in the code.

plugin/externalPluginEntry.go
```diff
@@ -390,6 +391,30 @@ func (e *externalPluginEntry) blockRead(ctx context.Context, size int64, offset
 	return inv.stdout.Bytes(), nil
 }
 
+func (e *externalPluginEntry) Write(ctx context.Context, p []byte) error {
```
ekinanp: We could DRY this up by re-factoring invocation to something like
```
type invocation struct {
  internal.Command // So we get the "SetStdout", "SetStderr", "SetStdin" stuff for free
  // All the other stuff
}

// Add an InvokeAndWait helper to invocation and modify pluginScript's InvokeAndWait to call that.
// May also need to store context in invocation to record to activity?
```

Then in Write, this would read as
```
inv := e.script.NewInvocation(ctx, "write", e).SetStdout(...).SetStderr(...).SetStdin(...)
err := inv.InvokeAndWait()
etc.
```

ekinanp (commented):
Reviewing the tests next

plugin/externalPluginEntry.go
ekinanp: Maybe we could move over `SetStdout`, `SetStderr` to invocation#InvokeAndWait. Idea is `invocation#InvokeAndWait` would fill invocation#stdout, invocation#stderr.

docs/docs/index.md
```diff
@@ -260,6 +261,21 @@ wash . ❯ grep "Hello" gcp/Wash/storage/some-wash-stuff/an\ example\ folder/sta
 echo "Hello, world!"
 ```
 
+### write
+The `write` action lets you write data to an entry. Thus, any command that writes a file also works with these entries.
+
+Note that Wash distinguishes between file-like and non-file-like entries. An entry is file-like if it's readable and writable and defines its size; you can edit it like a file. If it doesn't define a size then it's non-file-like, and trying to open it with a ReadWrite handle will error; reads from it may not return data you previously wrote to it.
+
+#### Examples
+```
```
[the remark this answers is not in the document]
ekinanp: Actually, would be cool to also include a vim example with a file-like entry (just to cement the fact that it would work with vim).

fuse/file_test.go
```diff
@@ -0,0 +1,653 @@
...
+
+	m := plugintest.NewMockBase()
+	mr := plugintest.NewMockRead()
+	mr.Attributes().SetSize(1)
+	mw := plugintest.NewMockWrite()
+	mw.Attributes().SetSize(1)
+	mrw := plugintest.NewMockReadWrite()
+	mrw.Attributes().SetSize(1)
+
+	// expectErrs corresponds to expectations on fields in flags.
```
ekinanp: This took me a bit to figure out what you mean, so I think an example would be useful here.

fuse/file_test.go
```diff
@@ -0,0 +1,653 @@
...
+
+func (suite *fileTestSuite) SetupTest() {
+	plugin.SetTestCache(datastore.NewMemCache())
+}
+
+func (suite *fileTestSuite) TearDownTest() {
+	plugin.UnsetTestCache()
+}
+
+func (suite *fileTestSuite) TestOpen_File() {
```
ekinanp: Nit: `File` => `FileLikeEntry` ? Latter's more precise and matches what's in the code.

fuse/file.go
```diff
@@ -3,103 +3,370 @@ package fuse
...
+	// Size of readable content, necessary for *non-file-like* entries
+	readSize uint64
+}
 
 func newFile(p *dir, e plugin.Entry) *file {
-	return &file{newFuseNode("f", p, e)}
+	return &file{fuseNode: newFuseNode("f", p, e), writers: make(map[fuse.HandleID]struct{})}
+}
+
+func (f *file) isFileLike() bool {
```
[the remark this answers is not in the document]
ekinanp: Hmm, maybe `isFileLike` => `isFileLikeEntry` ? It's slightly weird seeing `file#isFileLike`.

fuse/file.go
```diff
@@ -3,103 +3,370 @@ package fuse
...
+		return nil, fuse.ENOTSUP
+	case req.Flags.IsReadWrite() && (!readable || !writable):
+		activity.Warnf(ctx, "FUSE: Open read-write unsupported on %v", f)
+		return nil, fuse.ENOTSUP
 	}
 
-	if plugin.WriteAction().IsSupportedOn(updatedEntry) {
-		fh.w = updatedEntry.(plugin.Writable)
+	if !f.isFileLike() && req.Flags.IsReadWrite() {
+		// Error ReadWrite on non-file-like entries because it probably won't work well.
```
ekinanp: So with this code, if I have a non-file-like entry that implements both Read/Write, and I do something like `vim <entry>`, I'm going to see what's returned by read b/c vim opens ReadOnly. Then I do some edits and write what I see via `:w`. That's going to be written because vim opens WriteOnly (but what's written may also include the Read data, which can cause some weird issues to happen). I don't think there's anything we can do to prevent that, so might be worth adding a warning for users to be careful when editing non-file-like entries (and preferably that they _shouldn't_ edit non-file-like entries).

api/fs/root.go
```diff
@@ -0,0 +1,52 @@
...
+
+	r.EntryBase = plugin.NewEntry("local")
+	r.DisableDefaultCaching()
+	return nil
+}
+
+// Schema returns the root's schema
+func (r *Root) Schema() *plugin.EntrySchema {
+	return plugin.
+		NewEntrySchema(r, "local").
```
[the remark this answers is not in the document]
MikaelSmith: I think I'm going to skip this one, not an issue for now.

fuse/file.go
```diff
@@ -3,103 +3,370 @@ package fuse
...
 	"github.com/puppetlabs/wash/activity"
 	"github.com/puppetlabs/wash/plugin"
 )
 
 // ==== FUSE file Interface ====
 
+// file's implementation differs primarily on two axes: whether `Size` is set, and whether we're
+// currently writing.
+// - a *file-like* entry declares a `Size` in its `Attributes`
+//   - reads and writes are symmetric; the kernel page cache will be used, and the file size
```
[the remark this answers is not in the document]
MikaelSmith: I disagree slightly because size does not necessary correspond to state. It's possible to grow a file with setattr and then write the content. But it's close enough that I'm fine calling it `content`.

fuse/file.go
```diff
@@ -3,103 +3,370 @@ package fuse
...
+//   - when writing, `data` represents the local state of the file; changes to its size will be
+//     reflected in `data` and reads will be served from it. The file's size will be `len(data)`.
+//     Unchanged sections will be filled from `plugin.Read` as needed.
+//   - when not writing, data is read directly from `plugin.Read`
+// - a *non-file-like* entry has `Size` unset
+//   - read always pulls from `plugin.Read` and writes are buffered independently
+//   - `data` stores only the data to be written, and is not initialized from `plugin.Read`
+//   - the file's size will be reported as its readable size; it will not reflect calls to `write`
+//
+// Note that writes only result in calling `plugin.Write` when a file handle is closed. Writing
```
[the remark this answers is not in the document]
MikaelSmith: No. This is because FUSE's semantics are confusing. When the system call `close` is called on a file handle, FUSE calls the Flush callback. When the last copy of a file handle is closed, FUSE calls the Release callback. See the description at https://libfuse.github.io/doxygen/structfuse__operations.html#ad4ec9c309072a92dd82ddb20efa4ab14. Would a link to that help?

fuse/file.go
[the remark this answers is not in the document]
MikaelSmith: I don't think we should restrict the definition to only things that are Writable. I think `metadata.json` specifically is file-like, because it has a consistent size.

fuse/file.go
MikaelSmith: Would've been a call to Setattr. Rewriting this comment.

fuse/file.go
MikaelSmith: I'll elaborate in the comment. It's not only used to truncate the file (to size 0), you could directly call `truncate` on it and supply any size.

fuse/file.go
MikaelSmith: Preloaded external plugin content. I'll mention that in the comment.

fuse/file_test.go
MikaelSmith: And `TestOpen_NonFile` to `TestOpen_NonFileLikeEntry`? FileLikeEntry is a mouthful, I was using an obvious short-hand.

fuse/file.go
MikaelSmith: Editing with vim will work, it just may do something you don't expect. Not sure what to say beyond what's already in the docs.

The comment here is because if you have a ReadWrite handle, you expect what you write to be directly reflected in read operations, and things like `seek` to make sense. They won't, because reads and writes are not symmetric for non-file-like entries.

plugin/externalPluginEntry.go
MikaelSmith: I can work on refactoring things a bit.

fuse/file.go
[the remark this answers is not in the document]
ekinanp: Yes, that would help.

fuse/file.go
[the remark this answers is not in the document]
ekinanp: "current local content" => "current local content size". I'd also mention what you just said, so something like "If writing, then this is typically len(data); the atypical case is when the file's truncated via a setattr call"

fuse/file.go
```diff
@@ -3,103 +3,377 @@ package fuse
...
+	f.fillAttr(a)
+	activity.Record(ctx, "FUSE: Attr %v: %+v", f, *a)
+	return nil
+}
+
+func (f *file) fillAttr(a *fuse.Attr) {
+	attr := plugin.Attributes(f.entry)
+	applyAttr(a, attr, defaultMode(f.entry))
+
+	if f.useLocalContent() || !f.isFileLikeEntry() {
```
ekinanp: Yeah, this makes more sense. Thanks.

fuse/file.go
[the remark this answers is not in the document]
MikaelSmith: Reworked that section a bit.

fuse/file.go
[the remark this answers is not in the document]
MikaelSmith: Done.

fuse/file_test.go
ekinanp: Doesn't look too bad, but it is a nit so am OK if you want to leave it as-is.

ekinanp: Actually, it does help make the other tests a bit easier to read for me. Like `Read_File` is better understood as `Read_FileLikeEntry`. So would be nice to make that change.

ekinanp: It'd also help to name this test `TestOpen_FileLikeEntry_OpenFlags` since you're testing that Open does the right thing for all combos of openflag and entry. Similar case for the other one.

MikaelSmith: Done.

docs/docs/index.md
```diff
@@ -260,6 +261,34 @@ wash . ❯ grep "Hello" gcp/Wash/storage/some-wash-stuff/an\ example\ folder/sta
 echo "Hello, world!"
 ```
 
+### write
+The `write` action lets you write data to an entry. Thus, any command that writes a file also works with these entries.
+
+Note that Wash distinguishes between file-like and non-file-like entries. An entry is file-like if it's readable and writable and defines its size; you can edit it like a file.
+
+If it doesn't define a size then it's non-file-like, and trying to open it with a ReadWrite handle will error; reads from it may not return data you previously wrote to it. You should check its documentation with the `docs` command for that entry's write semantics.
```
ekinanp: I'd mention the stuff about `vim` here, like "We also recommend not using editors with these entries to avoid weird behavior".

fuse/file.go
ekinanp: Oh, the code makes sense. I only commented what I said here b/c that's when I thought of it.

See https://github.com/puppetlabs/wash/pull/664/files#r362674541 re: where to add a note about unexpected things w/ editors.

plugin/types.go
```diff
@@ -15,7 +15,15 @@ or referenced via the API - and tools for controlling how its data is cached.
 Implementing the Parent interface displays that resource as a directory on the filesystem.
 Anything that does not implement Parent will be displayed as a file.
 
-The Readable interface gives a file its contents when read via the filesystem.
+The Readable interface allows reading data from an entry via the filesystem. The Writable
+interface allows sending data to the entry.
+
+Wash distinguishes between two different patterns for things you can read and write. It considers
+a "file-like" entry to be one with a defined size (so the `size` attribute is set when listing the
```
ekinanp: OK, I agree about using `size` alone for file-like things. But I think we should remove `size` from the `metadata/data.json` files because those aren't really "files" (editing them makes no sense) so they shouldn't be treated as such.

plugin/types.go
```diff
@@ -183,24 +191,33 @@ type Streamable interface {
...
+// Writable is an entry that we can write new data to. What that means can be
+// implementation-specific; it could be overwriting a file, submitting a
+// configuration change to an API, or writing data to a queue. It doesn't
+// support a concept of a partial write.
+//
+// Writable can be implemented with or without Readable/BlockReadable. If an
+// entry is only Writable, then only full writes (starting from offset 0) are
+// allowed, anything else initiated by the filesystem will result in an error.
+//
+// It's up to the implementer to decide how much data integrity to guarantee.
```
ekinanp: What do you mean by "data integrity"?

plugin/types.go
```diff
@@ -183,24 +191,33 @@ type Streamable interface {
...
 	Read(context.Context) ([]byte, error)
 }
 
-// Writable is an entry that can write data directly to the entry. It mirrors
-// the WriterAt interface with an added context for handling the lifecycle of
-// remote write operations.
+// Writable is an entry that we can write new data to. What that means can be
+// implementation-specific; it could be overwriting a file, submitting a
+// configuration change to an API, or writing data to a queue. It doesn't
+// support a concept of a partial write.
```
ekinanp: Should we add a reminder to plugin authors to include write-semantics for non-file-like entries in the description? Also holds for the external plugin docs.

plugin/externalPluginEntry_test.go
```diff
@@ -661,6 +661,109 @@ func (suite *ExternalPluginEntryTestSuite) TestListReadWithMethodResults() {
...
+}
+func (m *mockCommand) StderrPipe() (io.ReadCloser, error) {
+	args := m.Called()
+	return args.Get(0).(io.ReadCloser), args.Error(1)
+}
+func (m *mockCommand) ExitCode() int { return m.Called().Int(0) }
+
+var _ = internal.Command(&mockCommand{})
+
+func (suite *ExternalPluginEntryTestSuite) TestWrite() {
```
ekinanp: Why does Write have so many tests? Seems like it should be symmetric with other `InvokeAndWait` tests.

plugin/types.go
MikaelSmith: I think editing metadata does make sense for some things: the metadata is a reflection of a mix of state and configuration, and you may be able to make changes to that configuration. Editing those values to change the configuration would make sense. This is especially true for everything in Kubernetes.

plugin/types.go
MikaelSmith: Primarily around whether writing volumes makes sure an fsync happens so that data is serialized to persistent storage before returning. Not sure how to summarize that.

plugin/externalPluginEntry_test.go
MikaelSmith: I'll have to spend a little time on this tomorrow. I'm not sure the other tests are verifying all the assertions we expect. They also mock at a higher level (the script's InvokeAndWait call) than Write had available originally.

plugin/types.go
ekinanp: Right, but what you write isn't going to strictly match what's read -- there isn't a 1:1 symmetry between the content because we output pretty-printed JSON. But I don't have strong feelings about keeping size around either so up to you.

plugin/types.go
ekinanp: Ah ok. Sounds like an issue for file-like entries, which could be summed up as `plugin.Write(data); data == plugin.Read` (ensure that the next Read from the API will fetch the written data). Don't know if that implies "always call fsync" for volumes.

plugin/types.go
MikaelSmith: Ok, that makes sense.

plugin/types.go
MikaelSmith: Done.

ekinanp (approved):

plugin/types.go
MikaelSmith: I ended up removing this because I don't think any plugins should explicitly implement something like calling `sync` after writing. It represents a possible loss of data, but the machine would have to die right after and we'd likely get a connection error if that happened. Things like vim and databases will call Fsync, and we're currently lying to them in some cases that it did something, but for now I'd suggest don't run a database on Wash's filesystem.

PR #668, opened 2019-12-17
Ensure ClearCache removes attributes in parent list as well

### LINKED ISSUE #667: wclear doesn't refresh entry attributes
When calling `wclear path`, the cached attributes for the entry at `path` aren't reset. I have to clear the parent to get updated attributes.

MikaelSmith:
Makes ClearCache more thorough about clearing data about the entry. Attributes and metadata frequently come from the parent's List command, so clear that cache as well.

Fixes #667.

### DIFF
--- a/plugin/cache.go
+++ b/plugin/cache.go
@@ -88,13 +88,30 @@ func allOpKeysIncludingChildrenRegex(path string) *regexp.Regexp {
 }
 
 // ClearCacheFor removes entries from the cache that match or are children of the provided path.
-// If successful, returns an array of deleted keys.
+// If successful, returns an array of deleted keys. Optionally clear the list operation for the
+// parent to remove any attributes related to the specified entry.
 //
 // TODO: If path == "/", we could optimize this by calling cache.Flush(). Not important
 // right now, but may be worth considering in the future.
-func ClearCacheFor(path string) []string {
+func ClearCacheFor(path string, clearParentList bool) []string {
 	rx := allOpKeysIncludingChildrenRegex(path)
-	return cache.Delete(rx)
+	deleted := cache.Delete(rx)
+
+	if clearParentList {
+		parentID, _ := splitID(path)
+		listOpName := defaultOpCodeToNameMap[ListOp]
+		deleted = append(deleted, cache.Delete(opKeyRegex(listOpName, parentID))...)
+	}
+
+	return deleted
+}
+
+// returns (parentID, cname)
+func splitID(entryID string) (string, string) {
+	segments := strings.Split(entryID, "/")
+	parentID := strings.Join(segments[:len(segments)-1], "/")
+	cname := segments[len(segments)-1]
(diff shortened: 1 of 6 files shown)


plugin/cache.go
```diff
@@ -88,13 +88,33 @@ func allOpKeysIncludingChildrenRegex(path string) *regexp.Regexp {
...
 
 // ClearCacheFor removes entries from the cache that match or are children of the provided path.
-// If successful, returns an array of deleted keys.
+// If successful, returns an array of deleted keys. Optionally clear the list operation for the
+// parent to remove any attributes related to the specified entry.
 //
 // TODO: If path == "/", we could optimize this by calling cache.Flush(). Not important
 // right now, but may be worth considering in the future.
-func ClearCacheFor(path string) []string {
+func ClearCacheFor(path string, clearParent bool) []string {
```
ekinanp: Nit: `clearParent` => `clearParentList` (might as well document that in the param name)

plugin/methodWrappers.go
```diff
@@ -254,35 +247,21 @@ func Delete(ctx context.Context, d Deletable) (deleted bool, err error) {
...
-	parentID, cname := splitID(d.id())
-	listOpName := defaultOpCodeToNameMap[ListOp]
-	entries, _ := cache.Get(listOpName, parentID)
-	if entries == nil {
-		return
-	}
+	//
+	// If !deleted, the entry will eventually be deleted. However it's likely that
+	// Delete did update the entry (e.g. on VMs, Delete causes a state transition).
+	// Thus we clear the parent's cached list result to ensure fresh data.
```
ekinanp: Nit: "Thus we clear" => "Thus we also clear". Took me a bit to figure that out.

plugin/cache.go
ekinanp: It might be nice to document that `clearParent == false` would be useful if the caller plans on modifying the cached entry map instead.

ekinanp (commented):
Looks good overall, just some comments.

plugin/cache.go
MikaelSmith: Modifying the cached entry map is only possible within the plugin module. So probably not useful to document generally.

ekinanp: Hmm, if that's the case, when would `clearParent == false` be useful to people outside the plugin package? Maybe `ClearCacheFor` could remain as-is, and we just have an internal `clearCacheFor` wrapper that does the real work? Not a blocker, just a thought.

ekinanp: Nevermind, I see it used in the `volume` package. So I guess for testing it makes it so you don't have to mock the parent's cache. Still feels weird.

ekinanp (approved):

plugin/cache.go
MikaelSmith: There might be other cases, but it does feel fairly special-case. Maybe we remove it later.

PR #679, opened 2020-01-10
Structure docs to be entry-point for a first time Wash user

ekinanp:
Made this PR at master so the website changes can easily be viewed via my fork.


cmd/rootMain.go
```diff
@@ -118,17 +119,20 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
...
-  Wash includes several built-in commands: ls, wexec, find, meta, tail.
-  See commands run with wash via 'whistory', and logs with 'whistory <id>'.
-Try 'help'`)
+	if plugin.IsInteractive() && successfullyLoadedPlugins {
+		var randomPlugin string
+		for plugin := range plugins {
+			randomPlugin = plugin
+			break
+		}
+		cmdutil.Println(fmt.Sprintf("Welcome to Wash! Try 'ls' then 'docs <child>' (e.g. 'docs %v')", randomPlugin))
```
ekinanp: Reason for this instead of `docs .` is so they get used to doing `ls` => `docs <child>` when navigating through Wash (and being comfortable w/ `docs` as the main source of truth for entries). It is also a useful way to introduce `stree <plugin_root>`. If this seems OK, I can update the registry's description.

cmd/rootMain.go
```diff
@@ -118,17 +119,20 @@ func rootMain(cmd *cobra.Command, args []string) exitCode {
...
 		return exitCode{1}
 	}
 	defer srv.Stop()
 
-	if plugin.IsInteractive() {
-		cmdutil.Println(`Welcome to Wash!
-  Wash includes several built-in commands: ls, wexec, find, meta, tail.
-  See commands run with wash via 'whistory', and logs with 'whistory <id>'.
-Try 'help'`)
+	if plugin.IsInteractive() && successfullyLoadedPlugins {
```
ekinanp: Reason for the "successfullyLoadedPlugins" check is b/c it'd be weird for a first-time user to see a WARNING message being printed and then the "Welcome to Wash!" prompt showing up afterwards as if everything was successful.

I could also just omit the "Try 'ls'" part, and leave the Welcome message as-is.

docs/_layouts/default.html
```diff
@@ -80,7 +80,7 @@
 
           <li>
             <span class="nav-header">
-              <a href="{{ '/tutorials' | relative_url }}" class="center-vertical">Tutorials</a>
+              <a href="{{ '/tutorials' | relative_url }}" class="center-vertical">Tutorials (optional)</a>
```
ekinanp: Am also considering changing "Tutorials" to "Hands-on Wash" so the title is less intimidating. Latter's a bit more involved though.

docs/_tutorials/index.md
```diff
@@ -3,7 +3,7 @@ title: Tutorials
 ---
 **Note:** Make sure you've [installed Wash]({{ '/getting_started' | relative_url }}) before going any further.
 
-This series of hands-on tutorials introduces you to Wash. Each tutorial assumes that you have the [test environment set-up](00_test_environment), and that you have an understanding of Wash’s key abstractions. If you're unfamiliar with Wash, start with the [Fundamentals](01_fundamentals/plugins-actions-entries). You can complete the tutorials in any order. If you're not sure where to start, here are a few good options: 
+This series of (optional) hands-on tutorials introduces you to Wash. Each tutorial assumes that you have the [test environment set-up](00_test_environment), and that you have an understanding of Wash’s key abstractions. If you're unfamiliar with Wash, start with the [Fundamentals](01_fundamentals/plugins-actions-entries). You can complete the tutorials in any order. If you're not sure where to start, here are a few good options:
```
ekinanp: I still really do want to remove "Fundamentals" b/c I think the changes here better (and more concisely) captures all of that material.

ekinanp: Still need to update some prompts with the new "Welcome" message so want to make sure we figure that out first.

cmd/rootMain.go
ekinanp: Here's an example first-time UX:
```
Eniss-MacBook-Pro:wash enisinan$ ./wash
Do you use aws (y/n)? y
docker? y
gcp? y
kubernetes? y
The aws, docker, gcp, kubernetes core plugins have been enabled.
You can disable them by modifying the 'plugins' key in your config
file (~/.puppetlabs/wash/wash.yaml), and then restarting the shell

Welcome to Wash! Try 'ls' then 'docs <child>' (e.g. 'docs aws')
wash . ❯ ls
aws/
docker/
gcp/
kubernetes/
wash . ❯ docs aws
This is the AWS plugin root. The AWS plugin reads the AWS_SHARED_CREDENTIALS_FILE
environment variable or $HOME/.aws/credentials and AWS_CONFIG_FILE environment
variable or $HOME/.aws/config to find profiles and configure the SDK. See
https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials_profiles.html
for more details on how to setup AWS profiles.

You can limit the loaded profiles by adding

aws:
  profiles: [profile_1, profile_2]

to Wash’s config file.

The AWS plugin currently supports EC2 and S3. IAM roles are supported when configured
as described here. Note that currently region will also need to be specified with the
profile.

If using MFA, Wash will prompt for it on standard input. Credentials are valid for 1 hour.
They are cached under wash/aws-credentials in your user cache directory so they can be
re-used across server restarts. Wash may have to re-prompt for a new MFA token in response
to navigating the Wash environment to authorize a new session.

SUPPORTED ACTIONS
* list
    - ls aws
        Type 'docs <child>' to view an ls'ed child's documentation
    - cd aws
        The 'W' environment variable contains the Wash root. Use 'cd $W' to
        return to it.
    - stree aws
        Gives you a high-level overview of the kinds of things that you will
        encounter when you 'cd' and 'ls' through this entry.
    - find aws <expression>
        Prints all children that satisfy <expression>. Type 'find --help' to
        see all the things that you can filter on.
    - (anything else that works with directories [e.g. 'tree'])

Type 'help' to see a list of all the built-in Wash commands.
```

ekinanp: Here's an example of an entry that includes the `SUPPORTED ATTRIBUTES` section:
```
wash . ❯ docs aws/wash/resources/ec2/instances/i-04621c13583930e6c/
This is an EC2 instance. Its Exec action uses SSH. It will look up port, user,
and other configuration by exact hostname match from default SSH config files.
If present, a local SSH agent will be used for authentication. Lots of SSH
configuration is currently omitted, such as global known hosts files, finding
known hosts from the config, identity file from config… pretty much everything
but port and user from config as enumerated in
https://github.com/kevinburke/ssh_config/blob/0.5/validators.go. The known hosts
file will be ignored if StrictHostKeyChecking=no, such as in

Host *.compute.amazonaws.com
  StrictHostKeyChecking no

SUPPORTED ATTRIBUTES
* mtime (last modified time) -- 2020-01-01 22:15:53 +0000 UTC
* crtime (creation time) -- 2020-01-01 22:15:52 +0000 UTC

The attributes are a subset of the entry's metadata, which contains everything
you ever need to know about the entry. You can use
    meta aws/wash/resources/ec2/instances/i-04621c13583930e6c/
to view the metadata and
    meta --attribute aws/wash/resources/ec2/instances/i-04621c13583930e6c/
to view the (partial) metadata.

SUPPORTED ACTIONS
* delete
    - delete aws/wash/resources/ec2/instances/i-04621c13583930e6c/
* exec
    - wexec aws/wash/resources/ec2/instances/i-04621c13583930e6c/ <command> <args...>
        e.g. wexec aws/wash/resources/ec2/instances/i-04621c13583930e6c/ uname
* list
    - ls aws/wash/resources/ec2/instances/i-04621c13583930e6c/
        Type 'docs <child>' to view an ls'ed child's documentation
    - cd aws/wash/resources/ec2/instances/i-04621c13583930e6c/
        The 'W' environment variable contains the Wash root. Use 'cd $W' to
        return to it.
    - stree aws/wash/resources/ec2/instances/i-04621c13583930e6c/
        Gives you a high-level overview of the kinds of things that you will
        encounter when you 'cd' and 'ls' through this entry.
    - find aws/wash/resources/ec2/instances/i-04621c13583930e6c/ <expression>
        Prints all children that satisfy <expression>. Type 'find --help' to
        see all the things that you can filter on.
    - (anything else that works with directories [e.g. 'tree'])
* signal
    - signal <signal> aws/wash/resources/ec2/instances/i-04621c13583930e6c/
        e.g. signal start aws/wash/resources/ec2/instances/i-04621c13583930e6c/

Type 'help' to see a list of all the built-in Wash commands.

SUPPORTED SIGNALS
* start
    Starts the EC2 instance
* stop
    Stops the EC2 instance
* hibernate
    Hibernates the EC2 instance
* restart
    Reboots the EC2 instance
* terminate
    Terminates the EC2 instance
```

cmd/docs.go
```diff
@@ -67,24 +97,145 @@ func docsMain(cmd *cobra.Command, args []string) exitCode {
...
+			actionDescriptionLines = []string{
+				fmt.Sprintf("- signal <signal> %s", path),
+				fmt.Sprintf("    e.g. signal start %s", path),
+			}
+		}
+		for _, line := range actionDescriptionLines {
+			supportedActions.WriteString(fmt.Sprintf("    %v\n", line))
+		}
+	}
+	supportedActions.WriteString("\nType 'help' to see a list of all the built-in Wash commands.")
```
ekinanp: May be worth mentioning tab-completion and globbing somewhere, but not sure.

cmd/rootMain.go
ekinanp: Hmm, I guess one advantage of just saying `Try 'docs .'` is that people could see all the `list` examples, so they'd be exposed to `stree` a bit earlier (without having to scrounge through detailed plugin root description). But I'm not sure how likely a long description would deter people from reading the `list` stuff -- my guess "not very" since I think it is expected that a `docs` command's output could be verbose.

cmd/docs.go
```diff
@@ -67,24 +97,145 @@ func docsMain(cmd *cobra.Command, args []string) exitCode {
...
+			actionDescriptionLines = []string{
+				fmt.Sprintf("- ls %s", path),
+				fmt.Sprintf("    Type 'docs <child>' to view an ls'ed child's documentation"),
+				fmt.Sprintf("- cd %s", path),
+				fmt.Sprintf("    The 'W' environment variable contains the Wash root. Use 'cd $W' to"),
+				fmt.Sprintf("    return to it."),
+				fmt.Sprintf("- stree %s", path),
+				fmt.Sprintf("    Gives you a high-level overview of the kinds of things that you will"),
+				fmt.Sprintf("    encounter when you 'cd' and 'ls' through this entry."),
+				fmt.Sprintf("- find %s <expression>", path),
```
ekinanp: Thinking of moving over `find`'s introduction to the "SUPPORTED ATTRIBUTES" section since attributes/metadata filtering is probably `find`'s biggest use-case.

docs/getting_started.md
```diff
@@ -31,7 +31,6 @@ Wash is distributed as a single binary; the only prerequisite is `libfuse`. Here
     * If anything fails, then check out the [known issues page]({{ '/known_issues' | relative_url }}) to see if the failure(s) correspond to any of the known issues. Otherwise, please don't hesitate to ask us on [slack](https://puppetcommunity.slack.com/app_redirect?channel=wash) for help (or file an [issue](https://github.com/puppetlabs/wash/issues))! Note that you can use `wash --verify-install` to test any fixes.
 
 * You're good to go! Try running `wash` to start up the shell[^1]
-    * If you plan on extending Wash, then check out the [external plugin docs]({{ '/docs/external-plugins' | relative_url }}).
```
ekinanp: Removed this b/c I already include something like this in an error message for a first-time user who does not enable any of the core plugins.

cmd/docs.go
MikaelSmith: Hopefully people just try it. I wouldn't add it without more reason.

cmd/docs.go
```diff
@@ -67,24 +97,145 @@ func docsMain(cmd *cobra.Command, args []string) exitCode {
...
 	}
 
+	cmdutil.Println(docs.String())
 	return exitCode{0}
 }
 
-func printSignalSet(setName string, signals []apitypes.SignalSchema) {
-	cmdutil.Println()
-	cmdutil.Printf("%v\n", setName)
+func stringifySupportedAttributes(path string, entry apitypes.Entry) string {
```
MikaelSmith: These could be good functions to unit test.

cmd/docs.go
MikaelSmith: Seems reasonable.

cmd/rootMain.go
MikaelSmith: I'd like to include `help` still too. It's not obvious that that should work.

docs/getting_started.md
MikaelSmith: I don't feel like they're mutually exclusive.

plugin/registry.go
```diff
@@ -129,3 +129,8 @@ func (r *stubRoot) ChildSchemas() []*EntrySchema {
 func (r *stubRoot) List(context.Context) ([]Entry, error) {
 	return []Entry{}, nil
 }
+
+const registryDescription = `
+Welcome to Wash, a UNIX-like shell that lets you manage all your things (entries) as
```
MikaelSmith: I'm undecided on using `thing` all over the place. I think it's ok to use entry consistently, and maybe describe what that means in one place.

docs/getting_started.md
ekinanp: It's unnecessary redundancy. A first-time, external plugin-only user is going to start the shell, going to say no to all the core plugins, and are then going to see the external plugin error message which already points to the docs.

docs/_tutorials/index.md
ekinanp: Also, much of "Fundamentals" is the type of stuff that you'll immediately "get" upon using Wash. Maybe something about Attributes/Metadata could be useful, but outside of that I don't think it has much value.

cmd/rootMain.go
ekinanp: I think I'm going to go back to "Try 'docs .'" --  biggest reason is b/c `find`'s being moved over to the Attributes/Metadata section so a first-time user won't see it being mentioned for a while. I'll mention `help` as part of the registry's description (which I'll also reword per your suggestion). Does that work? I like that better b/c `docs` is still the entry-point, but `help` is also mentioned early-on.

cmd/docs.go
ekinanp: I'll add some tests for them.

docs/_layouts/default.html
ekinanp: @MikaelSmith What are your thoughts on this?

docs/_tutorials/index.md
ekinanp: @MikaelSmith Thoughts on this? I think at the very least "Fundamentals" => "Attributes/Metadata".

docs/_layouts/default.html
MikaelSmith: I think I'd stick with Tutorials, not clear to me that Hands-on Wash is really better.

docs/_tutorials/index.md
MikaelSmith: Changing to Attributes/Metadata and slimming it down (if some parts seem redundant) seem reasonable.

cmd/rootMain.go
ekinanp: Ok, updated. You'll find `help` at the end of `docs <registry>`'s output

docs/_layouts/default.html
ekinanp: Tutorials it is.

docs/_tutorials/index.md
ekinanp: Done.

cmd/docs.go
ekinanp: Done.

plugin/registry.go
ekinanp: Updated.

ekinanp:
There's a few more updates I'll need to do:

* Update the "Welcome to Wash!" message in tutorials once these changes are released
* Link to the Attributes and Find tutorials in find's help text (and esp. the meta primary tutorial in meta primary's help text). I think that'd be a useful thing to expose to people.

MikaelSmith (approved):

Issue #658, opened 2019-12-12
Reduce downloading the (latest) Wash binary tgz to a directly runnable copy-paste command


ekinanp:
Based on the instructions in https://puppetlabs.github.io/wash/installing_wash, to download the latest Wash binary, users need to:
* Copy-paste the `curl` command
* Click a link that takes them to https://github.com/puppetlabs/wash/releases/latest to figure out the value of `{WASH_VERSION}`
* Substitute `{WASH_VERSION}` in their copy-pasted `curl` command
* Run the command

Ideally, users should be able to copy-paste the appropriate curl command and run it. So something like

```
curl -sfLO https://github.com/puppetlabs/wash/releases/download/latest/wash-latest-x86_64-unknown-linux.tgz
```

This requires changing the name of the generated `.tgz` file.

PR #680, opened 2020-01-10
Generate version-agnostic release TARs

ekinanp:
This will make it easier for people to install the project from source
when getting started. The website will be updated after the next Wash
release.

We continue to maintain the versioned TARs since that seems to be a
common convention in projects.

Supports #658

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,14 +43,16 @@ jobs:
     script:
     - GOOS=darwin GOARCH=amd64 go build -ldflags="$LINKFLAGS"
     - tar czf ${PROJECT_NAME}-${TRAVIS_TAG}-x86_64-apple-darwin.tgz wash
+    - cp ${PROJECT_NAME}-${TRAVIS_TAG}-x86_64-apple-darwin.tgz ${PROJECT_NAME}-x86_64-apple-darwin.tgz
     - GOOS=linux GOARCH=amd64 go build -ldflags="$LINKFLAGS"
     - tar czf ${PROJECT_NAME}-${TRAVIS_TAG}-x86_64-unknown-linux.tgz wash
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -43,14 +43,16 @@ jobs:
     script:
     - GOOS=darwin GOARCH=amd64 go build -ldflags="$LINKFLAGS"
     - tar czf ${PROJECT_NAME}-${TRAVIS_TAG}-x86_64-apple-darwin.tgz wash
+    - tar czf ${PROJECT_NAME}-latest-x86_64-apple-darwin.tgz wash
```
MikaelSmith: This could be a copy of the previous tar. I think it probably should be; somehow tar isn't necessarily deterministic.

ekinanp: You're right. Updated.

.travis.yml
```diff
@@ -43,14 +43,16 @@ jobs:
     script:
     - GOOS=darwin GOARCH=amd64 go build -ldflags="$LINKFLAGS"
     - tar czf ${PROJECT_NAME}-${TRAVIS_TAG}-x86_64-apple-darwin.tgz wash
+    - cp ${PROJECT_NAME}-${TRAVIS_TAG}-x86_64-apple-darwin.tgz ${PROJECT_NAME}-latest-x86_64-apple-darwin.tgz
```
MikaelSmith: I'd drop `latest` from the name, I don't think it provides any useful information. Unless making it have the same number of dashes is useful...

MikaelSmith (approved):

.travis.yml
ekinanp: Makes sense. Updated again.

MikaelSmith (approved):

Issue #684, opened 2020-01-13
Change `meta` attribute to partial metadata


ekinanp:
This is basically removing https://github.com/puppetlabs/wash/blob/master/plugin/entryAttributes.go#L63 from EntryAttributes and moving it over to EntryBase as a `PartialMetadata` field. Reason for the change is b/c "partial" metadata makes a lot more sense, and is what we've been telling people anyways re: the `meta` attribute.

We'll also need to:

- [x] Update api with the changes
- [x] Update find's implementation
- [x] In external plugins, add `partial_metadata` key to entry JSON; change `meta_attribute_schema` => `partial_metadata_schema` in entry schema JSON.
- [x] Change `meta --attribute` => `meta --partial`. Update `info` and `docs`.
- [x] Update find's help text, especially the meta primary's.

AFTER RELEASE:
- [x] Update website with these changes. These include external plugin docs, blog posts, command documentation, tutorials etc. External plugin docs should highlight difference between partial metadata and metadata.
- [x] Update the wash-ruby gem, and any (known) deps. This is currently just puppetwash. We should also implement `write` for wash-ruby since we haven't done that yet.

ekinanp:
This was done a while ago.

PR #685, opened 2020-01-14
(Code) Change meta attribute => partial metadata

ekinanp:
Supports #684.

There's a few more changes that need to be made (see the issue). These changes only make sense after the code here's released.

### DIFF
--- a/cmd/internal/find/primary/meta.go
+++ b/cmd/internal/find/primary/meta.go
@@ -70,58 +70,19 @@ const metaDetailedDescription = `
 NOTE: https://puppetlabs.github.io/wash/tutorials/02_find/meta-primary
 contains a hands-on tutorial of the meta primary.
 
-The meta primary constructs a predicate on the entry's metadata. By
-default, this is the meta attribute. If you'd like to construct the
-predicate on the entry's full metadata, then set the "fullmeta" option.
-Be careful when you do this, because find will make O(N) API requests
-to retrieve this information (N = the number of visited entries). As a
-precaution, the "fullmeta" option is only supported on entries with
-schemas.
+NOTE: The EXAMPLES section contains some real-world examples of the meta
+primary in action.
 
-Meta is a specialized filter, so you should only use it if you need to
-filter your entries on a property that isn't captured by the common Wash
-attributes. For example, the meta primary can be used to filter EC2
-instances on a specific tag. It can be used to filter Docker containers
-with a specified label. In general, the meta primary can be used to filter
-on any property that's specified in the entry's meta attribute. See the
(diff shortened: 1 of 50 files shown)


cmd/internal/find/primary/meta.go
```diff
@@ -70,58 +70,19 @@ const metaDetailedDescription = `
 NOTE: https://puppetlabs.github.io/wash/tutorials/02_find/meta-primary
 contains a hands-on tutorial of the meta primary.
 
-The meta primary constructs a predicate on the entry's metadata. By
```
ekinanp: This was all deleted b/c much of this stuff will be in the meta primary tutorial (seems like it makes sense to put it there). Am open to other ideas of restructuring this.

MikaelSmith: Ok, I agree most of that isn't needed as a reference, just the first time you encounter the concept.

plugin/entryBase.go
```diff
@@ -55,61 +55,34 @@ func NewEntry(name string) EntryBase {
...
 		e.SetTTLOf(defaultOpCode(op), 15*time.Second)
 	}
 	return e
 }
 
+func (e *EntryBase) partialMetadata() JSONObject {
+	if e.specifiedPartialMetadata != nil {
+		return e.specifiedPartialMetadata
+	}
+	return e.attributes.ToMap()
```
MikaelSmith: Why do we return attributes instead?

plugin/entryBase.go
```diff
@@ -55,61 +55,34 @@ func NewEntry(name string) EntryBase {
...
+}
+
 // ENTRY INTERFACE
 
-// Metadata returns the entry's meta attribute (see plugin.EntryAttributes).
-// Override this if e has additional metadata.
+// Metadata returns the entry's partial metadata. Override this if e has
```
MikaelSmith: It's worth calling out somewhere that it's normal to set partial metadata and override Metadata, and why that's the case.

MikaelSmith (changes requested):

plugin/entryBase.go
ekinanp: That's the whole "attributes subset of metadata" thing. Edge-case is if someone specifies just the attributes, but they don't set partial metadata, that likely means the attributes are the metadata. I'll add a comment.

ekinanp: Updated.

plugin/entryBase.go
ekinanp: Updated.

MikaelSmith (approved):

PR #690, opened 2020-01-15
Properly negate the default schemaP for schema-agnostic primaries

ekinanp:


cmd/internal/find/parser/parseExpression_test.go
```diff
@@ -171,6 +171,10 @@ func (s *ParseExpressionTestSuite) TestParseExpressionSchemaPEval() {
 	s.RNSTC("! -true", "", schema)
 	s.RNSTC("! ( -true -o -false )", "", schema)
 	s.RSTC("! ( -true -a -false )", "", schema)
+	// Test that the schemaP for schema-agnostic primaries will always return
+	// true even if the primary's negated
+	s.RSTC("-name 'foo'", "", schema)
```
MikaelSmith: I feel like we should have another test that these matches actually match what we expect.

ekinanp: What do you mean?

MikaelSmith: This is saying that both expressions will always be valid to compare against any entry, right? Is there a way we can test the macro level as well, that `-name 'foo'` matches an entry named `foo`, and that `! -name 'foo'` doesn't match `foo` but matches an entry not named `foo`?

ekinanp: Yes, it's meant to work for any schema-agnostic primary. `name` was chosen as the representative example. I'll further clarify the comment.

We probably could, but I prefer to separate those as primaries that are just plain old entry functions (where negating them is easy -- you return `!p(e)`) and primaries that implement their own negation semantics (like `meta`). The former's accomplished by the `true` and `false` primaries, while there's already tests for the latter. `name` is a plain old entry function, so its negation is already tested (and the test for `-name 'foo'` itself is in https://github.com/puppetlabs/wash/blob/master/cmd/internal/find/primary/name_test.go -- that test makes sure that the "function" is correct).

If you like, I can add primary-specific negation tests, but I'd like to do that as part of the RQL work if that is OK. Note that the bug here isn't related to the `name` function being wrong, it is b/c the schemaP isn't quite right.

ekinanp: Updated the comment.

MikaelSmith (approved):

PR #691, opened 2020-01-16
Move the external plugin implementation to its own package

MikaelSmith:
Separates the external plugin implementation into a separate package from `plugin` so that it can make use of other packages in Wash that themselves depend on `plugin` (like `volume` and `transport`).

The separation is possible by making `MethodSignature` a public type, and adding a hidden `externalPlugin` interface for "dynamic" plugin functionality ( where an entry's features may not be known at compile-time).


plugin/types.go
```diff
@@ -235,3 +237,25 @@ type Signalable interface {
 	Entry
 	Signal(context.Context, string) error
 }
+
+// This interface exists to break the circular dependency between plugin and external.
+// The external plugin implementation is in its own module so it can use other modules
+// that implement new features and have dependencies on this module.
+type externalPlugin interface {
+	MethodSupported(string) MethodSignature
```
ekinanp: Why not `MethodSignature` ? `MethodSupported` seems like a weird name.

plugin/entrySchema.go
```diff
@@ -123,6 +132,12 @@ func NewEntrySchema(e Entry, label string) *EntrySchema {
 // Note that UnmarshalJSON is not implemented since that is not
 // how plugin.EntrySchema objects are meant to be used.
 func (s EntrySchema) MarshalJSON() ([]byte, error) {
+	if s.entry == nil {
```
ekinanp: What caused this change?

plugin/entryBase.go
```diff
@@ -99,6 +99,12 @@ func (e *EntryBase) Name() string {
 	return e.name
 }
 
+// ID returns the entry's ID. It won't panic on an empty
```
ekinanp: Yeah, this makes sense. Do you think plugin authors would find this useful? If not, may be worth telling them to ignore this.

ekinanp (commented):
I'm not sure I like letting external plugins use `volume` and `transport` packages because those are part of the Go plugin library. Ruby/Python/other plugin libraries could provide their own versions of those packages. The duplication sucks, but I also worry that letting all external plugins access Go plugin library stuff could lead to a messy external plugin interface. Hopefully what you come up with solves both the duplication problem while also keeping the interface clean. But I definitely lean towards duplication if it means keeping the (existing) external plugin docs clean.

Ideally, it'd be nice if we could auto-generate external plugin library code from a specified "Go" template. E.g. some declarative thing like "These are the supported attributes, this is their type. These are the schema methods ..." and some code could generate the Ruby/Python/other language version of all that stuff (like auto-generate wash-ruby). Kind of like how AWS generates all their SDKs. That'd be pretty hard to extend for `volume` and `transport` though, but maybe not impossible?

Anyways, I think we both agree that the external plugin stuff should be in its own package, so glad this PR will take care of that.

plugin/types.go
MikaelSmith: Inherited from SupportedMethods, but MethodSignature seems reasonable too.

plugin/entryBase.go
MikaelSmith: Probably don't need to use it. We use it only to determine the plugin's name, which isn't a general problem. I guess we should either clarify what ID is (I'd just point to the description of the ID wrapper method) or note that it's primarily for implementing the external plugin support. I think I'll do the latter.

plugin/entrySchema.go
MikaelSmith: We used to use the private `entrySchema`, which is just a plain-old-data (POD in C++ parlance) struct. We would serialize that as part of the graph. I started using `EntrySchema` instead, but we only fill in the parts from `entrySchema`. So this method would hit a null pointer when we tried to marshal it because `entry` (and `graph`) weren't set.

I'm open to suggestions. I could probably mess with public/private nesting more to have a public POD class we can use for external but hide it from `EntrySchema`'s docs.

MikaelSmith:
I'd like to re-use transport in particular because SSH libraries in different languages can be quite different to setup, and that's a substantial burden to get right. Also anything we can pull into core Wash will perform better. I think the interface I have in mind will be pretty easy to work with.

plugin/types.go
ekinanp: Let's go with that then.

plugin/entrySchema.go
ekinanp: I'm still confused. I guess the specific scenario that led to `s.entry == nil` would help.

MikaelSmith: Pushed up a separate commit that shows what I proposed instead of adding that check.

plugin/entrySchema.go
```diff
@@ -67,7 +75,9 @@ func schema(e Entry) (*EntrySchema, error) {
 	}
 }
 
-type entrySchema struct {
+// EntrySchemaData is the raw data we intend to serialize. You should
```
MikaelSmith: I could update this comment to reflect that we use this as part of constructing nodes in the external plugin graph, and it needs to be kept in-sync with data we store in EntrySchema.

plugin/external/pluginEntry.go
```diff
@@ -270,8 +277,8 @@ func (e *externalPluginEntry) schema() (*EntrySchema, error) {
 		// As a sanity check, ensure that the methods specified in the entry's schema
```
MikaelSmith: I don't think this test is working right yet. Trying to start Wash with an external plugin, I got
```
{"Boltwash":{"children":["Group"],"description":null,"label":"bolt","metadata_schema":null,"methods":["list","schema"],"partial_metadata_schema":null,"signals":null,"singleton":true},"Group":{"children":["Target"],"description":null,"label":"group","metadata_schema":null,"methods":["list","schema"],"partial_metadata_schema":null,"signals":null,"singleton":null},"Target":{"description":null,"label":"target","metadata_schema":null,"methods":["list","exec","schema"],"partial_metadata_schema":null,"signals":null,"singleton":null}}
expected something like:
{
	"type_id_one":{
		"label": "one",
		"methods": ["list"],
		"children": ["type_id_two"]
	},
	"type_id_two":{
		"label":"two",
		"methods": ["read"]
	}
}
```

plugin/entrySchema.go
ekinanp: Oh, I think I see what's going on. For external plugins, the graph nodes are `EntrySchema` objects but `s.entry == nil` (makes sense because `s.entry` is meant for core plugin entry schemas). `s.entry == nil` means those nodes just contain `entrySchema` fields, so call `entrySchema#MarshalJSON` which is what you're doing here.

That makes sense. I prefer a comment along those lines instead of the `EntrySchemaData` type b/c I don't think the comment is that confusing, while the data type is.

plugin/external/pluginEntry_test.go
```diff
@@ -181,19 +180,8 @@ func (suite *ExternalPluginEntryTestSuite) TestDecodeExternalPluginEntryWithStat
 	}
 }
 
-func (suite *ExternalPluginEntryTestSuite) TestDecodeExternalPluginEntryWithCacheTTLs() {
```
MikaelSmith: Restore this test.

plugin/external/pluginEntry_test.go
```diff
@@ -298,58 +286,58 @@ func (suite *ExternalPluginEntryTestSuite) TestDecodeExternalPluginEntryWithSche
 	entry, err := decodedEntry.toExternalPluginEntry(context.Background(), true, false)
 	if suite.NoError(err) {
 		suite.True(entry.schemaKnown)
-		suite.Equal("foo", rawTypeID(entry))
+		suite.Equal("foo", entry.RawTypeID())
 	}
 }
 
-func (suite *ExternalPluginEntryTestSuite) TestDecodeExternalPluginEntryWithInaccessibleReason() {
```
MikaelSmith: Restore this test.

plugin/external/pluginEntry_test.go
```diff
@@ -298,58 +286,58 @@ func (suite *ExternalPluginEntryTestSuite) TestDecodeExternalPluginEntryWithSche
...
-		suite.True(entry.isInaccessible)
-	}
-}
+// 	entry, err := decodedEntry.toExternalPluginEntry(context.Background(), false, false)
+// 	if suite.NoError(err) {
+// 		suite.True(entry.isInaccessible)
+// 	}
+// }
 
-func (suite *ExternalPluginEntryTestSuite) TestSetCacheTTLs() {
```
MikaelSmith: Restore this test.

plugin/entrySchema.go
MikaelSmith: Ok. I'll revert it and clarify the comment.

plugin/external/pluginEntry.go
MikaelSmith: Oh, this is happening because I defined `list` to return an empty list, so there are no child classes to declare for the schema yet. Works fine if I comment out the `list` implementation.

plugin/external/pluginEntry_test.go
MikaelSmith: Have thoughts on how I should test that ttl and isInaccessible?

MikaelSmith: Exposing getters for those seem fine.

plugin/entrySchema.go
MikaelSmith: Having EntrySchema and EntrySchemaData was also in danger of doing the wrong type cast, this is a lot clearer.

MikaelSmith:
Ok, I think this is ready again.

plugin/external/pluginEntry_test.go
ekinanp: Yeah, getters are fine. I'd add them to methodWrappers though instead of EntryBase.

plugin/external/pluginEntry.go
ekinanp: Ah yeah, that makes sense.

plugin/external/pluginEntry_test.go
MikaelSmith: I don't think they're useful generally, they seem fine as part of EntryBase.

ekinanp: Was thinking that'd be a nice way to avoid a "plugin authors ignore this stuff" comment, but I think adding that is enough too.

MikaelSmith: Not sure we need to tell people to ignore it. It's pretty basic access, if someone wants to use it they can. I just don't have a reason to make it available more generally.

ekinanp (approved):

PR #692, opened 2020-01-18
External volume and transport features for external plugins

MikaelSmith:
Add features that external plugins can request Wash provide them from its core capabilities. Implement `volume.FS` and `transport.ExecSSH` as features.

### DIFF
--- a/plugin/external/pluginEntry.go
+++ b/plugin/external/pluginEntry.go
@@ -26,10 +26,13 @@ type decodedCacheTTLs struct {
 }
 
 // decodedExternalPluginEntry describes a decoded serialized entry.
+// Methods can be a string, or a tuple of [<method>, <value>]. <value> can take different forms
+// depending on the method. json.RawMessage lets us delay parsing these so we can parse them into
+// method-specific structs, which greatly simplifies validation.
 type decodedExternalPluginEntry struct {
 	TypeID             string                 `json:"type_id"`
 	Name               string                 `json:"name"`
-	Methods            []interface{}          `json:"methods"`
+	Methods            []json.RawMessage      `json:"methods"`
 	SlashReplacer      string                 `json:"slash_replacer"`
 	CacheTTLs          decodedCacheTTLs       `json:"cache_ttls"`
 	InaccessibleReason string                 `json:"inaccessible_reason"`
@@ -38,45 +41,83 @@ type decodedExternalPluginEntry struct {
 	State              string                 `json:"state"`
 }
 
-const entryMethodTypeError = "each method must be a string or tuple [<method>, <result_or_signature>], not %v"
+type methodTuple struct {
+	Method string
+	Value  json.RawMessage
+}
 
-func mungeToMethods(input []interface{}) (map[string]methodInfo, error) {
+func (m *methodTuple) UnmarshalJSON(data []byte) error {
+	var list []json.RawMessage
+	if err := json.Unmarshal(data, &list); err != nil {
+		return err
+	}
+
+	if len(list) != 2 {
+		return fmt.Errorf("each method must be a string or tuple [<method>, <value>], not %v", string(data))
+	}
+
+	if err := json.Unmarshal(list[0], &m.Method); err != nil {
+		return err
+	}
+	m.Value = list[1]
+	return nil
+}
+
+func (e decodedExternalPluginEntry) getMungedMethods() (map[string]methodInfo, error) {
 	methods := make(map[string]methodInfo)
-	for _, val := range input {
-		switch data := val.(type) {
-		case string:
-			methods[data] = methodInfo{
-				signature: plugin.DefaultSignature,
-			}
-		case []interface{}:
-			if len(data) != 2 {
-				return nil, fmt.Errorf(entryMethodTypeError, data)
-			}
-			name, ok := data[0].(string)
-			if !ok {
-				return nil, fmt.Errorf(entryMethodTypeError, data)
-			}
-			info := methodInfo{
-				signature: plugin.DefaultSignature,
-			}
-			switch name {
-			default:
-				info.result = data[1]
-			case "read":
-				// Check if we have ["read", <block_readable?>] or ["read", <result>].
-				// The latter implies <block_readable> == false.
-				if blockReadable, ok := data[1].(bool); ok {
-					if blockReadable {
-						info.signature = plugin.BlockReadableSignature
-					}
-				} else {
-					info.result = data[1]
-				}
-			}
-			methods[name] = info
-		default:
-			return nil, fmt.Errorf(entryMethodTypeError, data)
+	for _, raw := range e.Methods {
+		// Try to unmarshal to a string. If that doesn't work, unmarshal to a method-specific tuple.
+		var name string
+		if err := json.Unmarshal(raw, &name); err == nil {
+			methods[name] = methodInfo{signature: plugin.DefaultSignature}
+			continue
 		}
+
+		var tuple methodTuple
+		if err := json.Unmarshal(raw, &tuple); err != nil {
+			return nil, err
+		}
+
+		// Assume a default signature. Then validate prefetched values for specific methods and update
+		// methodInfo as needed.
+		info := methodInfo{signature: plugin.DefaultSignature}
+		switch tuple.Method {
+		case "read":
+			// Check if we have ["read", <block_readable?>] or ["read", <content>].
+			// The latter implies <block_readable> == false.
+			var blockReadable bool
+			if err := json.Unmarshal(tuple.Value, &blockReadable); err == nil {
+				if blockReadable {
+					info.signature = plugin.BlockReadableSignature
+				}
+				break
+			}
+
+			var content string
+			if err := json.Unmarshal(tuple.Value, &content); err != nil {
+				return nil, fmt.Errorf("Read method must provide a string, not %v", string(tuple.Value))
+			}
+			info.tupleValue = []byte(content)
+		case "list":
+			var decodedEntries []decodedExternalPluginEntry
+			if err := json.Unmarshal(tuple.Value, &decodedEntries); err != nil {
+				return nil, fmt.Errorf("implementation of list must conform to %v, not %v", listFormat, string(tuple.Value))
+			}
+			info.tupleValue = decodedEntries
+		case "schema":
+			graph, err := unmarshalSchemaGraph(e.Name, e.TypeID, tuple.Value)
+			if err != nil {
+				return nil, fmt.Errorf(
+					"could not decode schema from stdout: %v\nreceived:\n%v\nexpected something like:\n%v",
+					err,
+					string(tuple.Value),
+					schemaFormat,
+				)
+			}
+			info.tupleValue = graph
+		}
+
+		methods[tuple.Method] = info
 	}
 	return methods, nil
 }
@@ -89,7 +130,7 @@ func (e decodedExternalPluginEntry) toExternalPluginEntry(ctx context.Context, s
 		return nil, fmt.Errorf("the entry's methods must be provided")
 	}
 
-	methods, err := mungeToMethods(e.Methods)
+	methods, err := e.getMungedMethods()
 	if err != nil {
 		return nil, err
 	}
@@ -108,7 +149,7 @@ func (e decodedExternalPluginEntry) toExternalPluginEntry(ctx context.Context, s
 			return nil, fmt.Errorf("entry %v (%v) implements schema, but the plugin root doesn't", e.Name, e.TypeID)
 		}
 		// schemaKnown || isRoot
-		if !isRoot && info.result != nil {
+		if !isRoot && info.tupleValue != nil {
 			return nil, fmt.Errorf(
 				"entry %v (%v) prefetched its schema. Only plugin roots support schema prefetching",
 				e.Name,
@@ -125,8 +166,8 @@ func (e decodedExternalPluginEntry) toExternalPluginEntry(ctx context.Context, s
 
 	// If read content is static, it's likely it's not coming from a source that separately provides
 	// the size of that data. If not provided, update it since we know what it is.
-	if content, ok := methods["read"].result.(string); ok && !e.Attributes.HasSize() {
-		e.Attributes.SetSize(uint64(len(content)))
+	if val := methods["read"].tupleValue; val != nil && !e.Attributes.HasSize() {
+		e.Attributes.SetSize(uint64(len(val.([]byte))))
 	}
 
 	entry := &pluginEntry{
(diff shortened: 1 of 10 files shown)


MikaelSmith:
Demonstrated with https://github.com/puppetlabs/boltwash/pull/1 using https://github.com/puppetlabs/wash-ruby/pull/27. Still need to write lots of testing in various places, but this should be useful to get feedback on how you set it up.

ekinanp:
I think exporting `volume.FS` and `transport` makes sense, I'm just not sure if a separate `features` key is the best way to do it. Here's why I say that:

* `features` needs to be specified in the entry schema JSON. However, `features` doesn't contribute to understanding an entry's high-level nature, which is what the entry schema's for. For example, why would a Wash user care about what core plugin features an entry's using? But I also get why you're including it (so that `stree <ancestor>` shows `volume::fs`). That reason just feels hacky, and hacky means that it may not be the right way to think about this.

* `transport` as-it-currently-is just seems to be a Wash user's way of saying "Hey, I'd like to implement my `exec` using your transport package". It seems like a natural way of extending the method tuple concept to `[<method>, <method_specific_info>]` where for `list` it's a prefetched result, for `read` it's a prefetched result or signature, and for `exec`, it's an implementation, in this case using stuff in the `transport` package. So I don't see why we can't stick the `transport` JSON there.

* `volume.fs` is "I want to re-use this core-plugin entry". Could we achieve that via something like:
    * Check if `list`'s return contains a core plugin entry JSON. A core plugin entry JSON is an entry JSON that needs (all other keys are ignored):
        * `type_id` (this is what tells Wash that we're re-using a core-plugin entry). For `volume.fs`, this could be `__volume::fs__` (or w/e type ID that's likely to minimize conflict).
        * `name` -- For the entry's name
        * `state` -- Representing the entry-specific config. This is where `maxdepth` and all that could go.
For entry schema, we can just check if they provide the core plugin entry type ID in the `children` key. That does make it harder to override specific schema keys (e.g. if you want to change the description or if you want to change the `name` [makes singleton entries easier to document if you change `fs` entry's name for example]), but there's probably a clean way to support that too (have some ideas on that).

MikaelSmith:
Those make sense to me. I'll work on the details a bit today.

plugin/external/pluginEntry.go
```diff
@@ -29,7 +29,7 @@ type decodedCacheTTLs struct {
 type decodedExternalPluginEntry struct {
 	TypeID             string                 `json:"type_id"`
 	Name               string                 `json:"name"`
-	Methods            []interface{}          `json:"methods"`
+	Methods            []json.RawMessage      `json:"methods"`
```
ekinanp: A comment explaining why this is a RawMessage would be useful.

plugin/external/pluginEntry.go
```diff
@@ -40,42 +40,65 @@ type decodedExternalPluginEntry struct {
 
 const entryMethodTypeError = "each method must be a string or tuple [<method>, <result_or_signature>], not %v"
 
-func mungeToMethods(input []interface{}) (map[string]methodInfo, error) {
+type methodTuple struct {
+	Method            string
+	ResultOrSignature json.RawMessage
```
ekinanp: This should be consistent with what we'll be documenting in the external plugin docs. I'm thinking `Value`'s more generic. Then I explain things as "value of the method tuple depends on the method. For list, it is list's result (prefetching). For read, it is read's signature OR its result. For exec, it is the transport ..."

Nit: May be worth changing `Method` => `MethodName` since external plugin docs we say `(<method_name>, ...)` when talking about the method tuple.

ekinanp: `methodInfo` should probably change `result` => `tupleValue` so it is more generic.

plugin/external/pluginEntry.go
```diff
@@ -40,42 +40,65 @@ type decodedExternalPluginEntry struct {
...
-			case "read":
-				// Check if we have ["read", <block_readable?>] or ["read", <result>].
-				// The latter implies <block_readable> == false.
-				if blockReadable, ok := data[1].(bool); ok {
-					if blockReadable {
-						info.signature = plugin.BlockReadableSignature
-					}
+	for _, raw := range input {
+		// Try to unmarshal to a string. If that doesn't work, unmarshal to a method-specific tuple.
+		var data string
```
ekinanp: Nit: `data` => `name`. Latter explicitly indicates what we're unmarshalling.

ekinanp (commented):
RawMessage's a good improvement, makes our reasoning about the unmarshaling explicit in the code. I'm guessing it will also let us validate the tuple value in the future when we're unmarshaling the entry instead of @ method-call time like what's currently going on?

plugin/external/pluginEntry.go
MikaelSmith: I said ResultOrSignature because that's explicitly what was used on the error string when you don't match the expected form.

ekinanp: Yeah, that error string should probably be changed then.

plugin/external/pluginEntry.go
```diff
@@ -125,7 +179,13 @@ func (e decodedExternalPluginEntry) toExternalPluginEntry(ctx context.Context, s
 
 	// If read content is static, it's likely it's not coming from a source that separately provides
 	// the size of that data. If not provided, update it since we know what it is.
-	if content, ok := methods["read"].result.(string); ok && !e.Attributes.HasSize() {
+	if sig := methods["read"]; sig.tupleValue != nil && !e.Attributes.HasSize() {
```
ekinanp: It feels weird to have content validation here. Could we move that over to where the method tuples are being unmarshaled?  I think it's OK if this part just sets the size attribute.

plugin/external/pluginEntry.go
```diff
@@ -370,15 +451,13 @@ func (e *pluginEntry) List(ctx context.Context) ([]plugin.Entry, error) {
...
 	return entries, nil
 }
 
 func (e *pluginEntry) Read(ctx context.Context) ([]byte, error) {
-	if impl := e.methods["read"].result; impl != nil {
-		if content, ok := impl.(string); ok {
-			return []byte(content), nil
-		}
-		return nil, fmt.Errorf("Read method must provide a string, not %v", impl)
+	if impl := e.methods["read"].tupleValue; impl != nil {
```
ekinanp: The tupleValue should already be parsed by this point right? If so, isn't it enough to just cast to content or byte or whatever we're returning?

plugin/external/pluginEntry.go
```diff
@@ -535,6 +614,19 @@ func (e *pluginEntry) Stream(ctx context.Context) (io.ReadCloser, error) {
 }
 
 func (e *pluginEntry) Exec(ctx context.Context, cmd string, args []string, opts plugin.ExecOptions) (plugin.ExecCommand, error) {
+	if result := e.methods["exec"].tupleValue; result != nil {
+		impl, ok := result.(execImpl)
```
ekinanp: This validation feels redundant but maybe that's just me.

plugin/external/pluginEntry.go
MikaelSmith: How can it get the correct size unless it knows the type?

MikaelSmith: Ah, yeah I think it makes sense to move all the parsing to mungeToMethods.

plugin/external/pluginEntry.go
```diff
@@ -361,6 +433,15 @@ func (e *pluginEntry) List(ctx context.Context) ([]plugin.Entry, error) {
 
 	entries := make([]plugin.Entry, len(decodedEntries))
 	for i, decodedExternalPluginEntry := range decodedEntries {
+		if decodedExternalPluginEntry.TypeID == volumeFSTypeID {
+			var opts volumeFSOpts
```
ekinanp: You could generalize this to "Check if user's asking for a core plugin entry" via a map of `<type_id> => <constructor>` where `<constructor>` could take the external plugin entry as its argument. Then this would read as:
```
if createEntry, ok := corePluginEntryTypeID(e.TypeID) {
  entry, err := createEntry(e)
  ...
}
// External plugin entry so rest of code is here
```

I think this generalization seems easy enough that we might as well do it here. Also helps clarify the code.

It might even be worth modifying the map to something like `<type_id>` => `<core_plugin_entry_info>` where `<core_plugin_entry_info>` includes the constructor and the corresponding schema. Then you can do a similar thing in EntrySchema.

**Note:** We should come up with a better name than core plugin entry so that it still persists even when some core plugins become separate Wash implementations . Maybe "wash-provided entry"? Don't really have good ideas here.

plugin/external/pluginEntry.go
MikaelSmith: Will be after a refactor a bit.

plugin/external/pluginEntry.go
MikaelSmith: It is now.

plugin/external/pluginEntry.go
MikaelSmith: Ah yeah, I understand now. Dropped checking the cast.

plugin/external/pluginEntry.go
MikaelSmith: Done.

plugin/external/pluginEntry.go
```diff
@@ -337,18 +397,37 @@ func (e *pluginEntry) SchemaGraph() (*linkedhashmap.Map, error) {
 
 const listFormat = "[{\"name\":\"entry1\",\"methods\":[\"list\"]},{\"name\":\"entry2\",\"methods\":[\"list\"]}]"
 
+type coreEntryCreator interface {
```
ekinanp: Yeah, `coreEntry` seems like a good enough name for now. Some suggestions:

* `coreEntryCreator` => `coreEntry`, `create` => `createInstance`, `coreEntryTypeID` => `coreEntries`. I think all of that reads quite well when put together. It would also help if `createInstance`'s signature included the param names, because it took me a bit to see that `pluginEntry` represented the parent (so `create(parent, decodedEntry)`)

* Maybe a separate `external/coreEntry.go` file? This could be where all the core entries go.

MikaelSmith: Done.

ekinanp:
Cool, pending tests the changes here look good.

MikaelSmith:
Added tests, I think just docs remaining.

MikaelSmith:
Added docs. I can make that a separate PR if needed.

docs/docs/external-plugins.md
```diff
@@ -298,43 +404,33 @@ This section describes the JSON object representing a serialized entry. An entry
...
-
-   **Note:** `read`'s method-tuple can also be specified as `["read", <block_readable?>]` where `<block_readable?>` is a Boolean value. Entries that implement `read`'s block-readable calling convention _must_ specify `read` as the method-tuple `["read", true]`. Similarly, entries that implement `read`'s default signature _can_ specify `read` as the method-tuple `["read", false]`, but this is not required. Finally a prefetched `read` result, i.e. a method-tuple of `["read", <content>]`, implies that `<block_readable?>` is false.
+* `methods` is an array specifying the entry's implemented methods. Each element in the array can be a string representing the method's name, or a method-tuple of `[<method_name>, <method_result>]` indicating a prefetched result. The result should have the same format as `<method_name>`'s output in its calling convention. Note that prefetching is a useful way to avoid unnecessary plugin script invocations. Note that only some methods support prefetched results, and some have unique values of `<method_result>` that enable additional functionality. See the `prefetch` section of each method for specific details.
```
ekinanp: Naming the method-tuple sections as `prefetch` feels weird even if it is the most common tuple value (I can't help but think "why not call it what it is"). I suggest that this be re-worded to something like

> `methods` is an array specifying the entry's implemented methods. Each element in the array can be a string representing the method's name, or a method-tuple of `[<method_name>, <value>]`, where `<value>` is method-specific. See the `Tuple Value` section of each method for specific details.

I think it's OK to repeat prefetch details in the relevant sections. Or maybe we could add note here indicating that some tuple values represent a prefetched result? Not sure. Either way, I don't think `Prefetch` should be the name of the sections describing the method-tuple value -- we should try to keep that generic to account for `exec`.

docs/docs/external-plugins.md
```diff
@@ -298,43 +404,33 @@ This section describes the JSON object representing a serialized entry. An entry
...
+  ```
+  [
+    "list",
+    "exec"
+  ]
+  ```
 
-   Notice that `list`'s `<method_result>` matches what's outputted by a `list` invocation. Similarly, `read`'s `<method_result>` matches what's outputted by a `read` invocation. Also, Wash knows that `<block_readable?>` is false for this entry.
+  ```
+  # With prefetching
```
ekinanp: This should probably be "With method-tuples".

docs/docs/external-plugins.md
```diff
@@ -138,6 +152,48 @@ bash-3.2$ /path/to/myplugin.rb list /myplugin/foo ''
...
 
+### Prefetch
+
+`list` can prefetch other it's entries. The output would be an array of other entries, identical to the output from running the script in the previous examples.
+
+Additionally Wash supports _core entries_ that provide built-in functionality that plugins can take advantage of.
+
+_core entries_ can be used by returning an entry object with `type_id` set to the _core entries_ name surrounded by double underscores (`__core::entry__`) and the `name` field to identify the entry. You can also specify options for a _core entry_ using the `state` field containing serialized JSON of the options. If your plugin use _schemas_, then when using a _core entry_ you must still specify the _core entry_ as a child of the entry that lists it in the schema by it's `type_id`.
+
+* `volume::fs`: a representation of a filesystem that uses your entry's `exec` method to access it. _Options_:
```
ekinanp: Nit: representation of your entry's filesystem that uses its `exec` method to access it? The current wording makes it seem more abstract than what it is (e.g. what filesystem is it representing?)

ekinanp (commented):
Some comments on the docs, otherwise this looks good.

docs/docs/external-plugins.md
```diff
@@ -241,15 +305,53 @@ bash-3.2$ echo "$?"
 0
 ```
 
+### Tuple Value
+
+`exec`'s value can only be used to specify using a built-in _transport_ provided by Wash. The method tuple must be
```
ekinanp: We should mention that `exec`'s tuple-value represents an implementation (currently the only possible implementation is provided by the _transport_ package). That (important) fact is implicit here.

I think the first sentence could be replaced by something like

> `exec`'s value represents an implementation of `exec`. Wash will use this implementation to handle all `exec` calls, so you do _not_ have to implement `exec`'s plugin script invocation for this entry.

> Currently, only implementations provided by the _transport_ package are supported.  The method tuple must be \<rest of docs\>

docs/docs/external-plugins.md
```diff
@@ -241,15 +305,53 @@ bash-3.2$ echo "$?"
...
+  "exec",
+  {
+    "transport": <transport>,
+    "options": { ... }
+  }
+]
+```
+where the object in the `options` field specifies transport-specific options. Supported transports are:
+
+* `ssh`: connect to an SSH server. It will look up port, user, and other configuration by exact hostname match from default SSH config files (global SSH config is currently ignored). If present, a local SSH agent will be used for authentication. The known hosts file will be ignored if StrictHostKeyChecking=no in your SSH config. _Options_ (string values unless otherwise specified):
```
ekinanp: I'd indicate somewhere that the `ssh` transport uses SSH to implement your Exec (unless that's what "connect to an SSH server" means).

ekinanp (approved):

MikaelSmith:
I'm going to split the docs change into a separate PR so we can merge it around release time.

PR #694, opened 2020-01-27
Add option to disable host key check

MikaelSmith:
The boltwash plugin implements Bolt's `host-key-check: false` option by passing a null known hosts file and using the default of accepting new keys. Provide tests that this technique will continue to work.

### DIFF
--- a/transport/ssh_test.go
+++ b/transport/ssh_test.go
@@ -21,9 +21,11 @@ import (
 
 type SSHTestSuite struct {
 	suite.Suite
-	s                        gssh.Server
-	m                        mock.Mock
-	knownHosts, identityFile string
+	s             gssh.Server
+	m             mock.Mock
+	knownHosts    string
+	badKnownHosts string
+	identityFile  string
 }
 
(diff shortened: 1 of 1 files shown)


ekinanp (approved):

transport/ssh.go
```diff
@@ -212,6 +213,8 @@ type Identity struct {
 	// Retries can be set to a non-zero value to retry every 500ms for that many times.
 	Retries uint `json:"retries"`
 	Port    uint `json:"port"`
+	// If true, skip all host key checking.
+	DisableHostKeyCheck bool `json:"disable_host_key_check"`
```
MikaelSmith: I might be able to do this by making KnownHosts=/dev/null.

ekinanp: Whichever way's the one that people are most familiar with works for me.

MikaelSmith: I dislike this approach and hope Bolt moves away from it someday. I think I'm going to remove this option andhack around it with KnownHosts.

PR #698, opened 2020-01-29
Introduce the Bolt plugin for Wash

### LINKED ISSUE #702: Post about using Bolt + Wash together

MikaelSmith:
Adds a post introducing the plugin.

Resolves #702.

### DIFF
--- a/docs/_posts/2020-02-07-boltwash.md
+++ b/docs/_posts/2020-02-07-boltwash.md
@@ -63,6 +63,36 @@ Administrator/
 Public/
 ```
 
+Boltwash includes target configuration as metadata, so you can also view that
+```
+wash bolt/my-linux-nodes > meta node1
+alias: []
+config:
+  ssh:
+    connect-timeout: 10
+    disconnect-timeout: 5
+    load-config: true
+    run-as: root
+    tty: false
+    user: ubuntu
+  transport: ssh
+facts: {}
+features: []
+name: node1
+plugin_hooks: {}
+uri: node1
+vars:
+  datacenter: west coast
+```
+and use `find` to search for nodes based on that metadata
+```
+wash bolt > find . -meta .vars.datacenter 'west coast'
+all/node1
+all/node5
+my-linux-nodes/node1
+my-windows-nodes/node5
+```
+
 ## Debugging with Bolt + Wash
 
 Wash can be helpful debugging all sorts of failures. As an example, suppose you have a cluster of machines (in the `webservers` group) that are having problems
--- a/docs/_posts/2020-02-07-boltwash.md
+++ b/docs/_posts/2020-02-07-boltwash.md
@@ -1,0 +1,129 @@
+---
+title: "Bolt + Wash"
+description: "Navigate your Bolt inventory in Wash"
+author: michaelsmith
+---
+
+If you're not using Puppet Bolt, [check it out](https://puppetlabs.github.io/bolt/).
+
+If you're using Bolt, then we created the [Boltwash](https://github.com/puppetlabs/boltwash) plugin that lets you navigate your inventory. Let's check it out
+```
+wash . > stree bolt
+bolt
+└── [group]
+    └── [target]
+        └── fs
+```
+
+## Exploring Boltwash
+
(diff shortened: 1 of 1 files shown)


docs/_posts/2020-01-29-boltwash.md
```diff
@@ -0,0 +1,65 @@
...
+
+If you're not using a cloud provider but you are using [Puppet Bolt](https://puppetlabs.github.io/bolt), then you've already created groups of targets, assigned properties to those targets, and configured how to connect to them. We created the [boltwash](https://github.com/puppetlabs/boltwash) plugin to leverage your existing Bolt inventory in Wash.
+
+The Bolt plugin for Wash presents a view of Bolt targets organized by groups, and allows you to navigate their filesystems (currently only with SSH)
```
ekinanp: I think this narrative is good enough for the Boltwash repo since people will have likely visited that repo after having some idea of what Boltwash does, but it should be further fleshed out for a Wash blog post/possible Puppet blog post where people have no idea what Boltwash does or what Wash is. Whatever narrative that's here should show off as much of the plugin as possible and only leave implicit details if we can assume that 99% of readers will be able to connect the dots and have that "Aha" moment.

As an example, I'm thinking of the following narrative

"If you're using Bolt, then we created the Boltwash plugin that lets you navigate your inventory... Let's check it out"

_Paste stree output here_

As you can see, the Bolt plugin organizes Bolt targets as groups, blah blah blah

Let's try ls'ing it.

```
ls bolt
```

(We can see the groups)

Let's cd into a group and ls its targets

```
cd group ls target
```

Now let's cd into a target and navigate its filesystem

```
<cd target, ls>
<paste output showing you ls'ing and cd'ing through things, cat'ing/tail'ing logs>
```

Cool, right? Now what's really great about Boltwash is you can use it to debug any failures. Here's an example

```
<Walkthrough your narrative here then link to some other interesting Boltwash use-cases>
```

I think embedding a screencast in the blog post that summarizes all of this stuff might be useful so people can see everything we just talked about.

Also may be worth mentioning the configurable Boltdir stuff and show off some target metadata (and possible filtering with `find` if the current metadata's useful for that).

I think Boltwash is an interesting enough idea that we could get away with a somewhat longer Blog post.

MikaelSmith: This is good feedback. It'll take me a bit to work through it.

docs/_posts/2020-02-07-boltwash.md
```diff
@@ -0,0 +1,128 @@
...
+  {
+    "status": "MainPID=21936,LoadState=loaded,ActiveState=active"
+  }
+Successful on 3 targets: server1,server2,server3
+Ran on 3 targets in 1.91 sec
+```
+
+You can imagine other scenarios as well. Suppose you ran a Bolt plan across several targets and want to get some files back from them. You can use Wash to directly copy them with something like
+```
+wash bolt > for f in group/*/fs/path/to/file; do cp -v "$f" ~/dest/"${f//\//_}"; done
```
ekinanp: I'd maybe add a note here saying that the sed stuff is just to remove `fs` entry from the paths.

docs/_posts/2020-02-07-boltwash.md
```diff
@@ -0,0 +1,128 @@
...
+home/
+...
+```
+
+Great, I can see the files on that system, and I didn't have to do any additional setup. I can go look at config or watch logs
+```
+wash bolt/my-linux-nodes/node1/fs > cat etc/hosts
+127.0.0.1 localhost
+...
+wash bolt/my-linux-nodes/node1/fs > tail -f var/log/syslog
```
ekinanp: Think an example with `find` would be useful? Maybe if logs generated by a plan had a certain regex?

MikaelSmith: Maybe, but kind of out of juice for the day.

MikaelSmith: I don't have any obvious ideas for that that seem better than the grep example.

ekinanp: No worries, not super important.

ekinanp (approved):

ekinanp (approved):

PR #700, opened 2020-01-30
Add the initial RQL architecture

ekinanp:
There's still some more work that needs to be done so this should be targeted at a separate `rql` branch (didn't want to create one now when everyone's asleep). Some of this includes

* Testing the `ast.Query` type
* Implementing the walker and stree pruning (and add the HTTP endpoint)
* Finish implementing the `meta` primary
* Support metadata schemas
* Improve error-handling
* Document everything (after release)

The PR's big, but lots of it is boilerplate.

### DIFF
--- a/api/rql/ast/predicateExpression.go
+++ b/api/rql/ast/predicateExpression.go
@@ -1,0 +1,47 @@
+package ast
+
+import (
+	"time"
+
+	"github.com/puppetlabs/wash/api/rql"
+	"github.com/puppetlabs/wash/api/rql/internal/predicate"
+	"github.com/puppetlabs/wash/api/rql/internal/predicate/expression"
+	"github.com/puppetlabs/wash/plugin"
+	"github.com/shopspring/decimal"
+)
+
+// PE_Primary returns a node representing a predicate expression (PE)
+// of primaries
+func PE_Primary() rql.Primary {
+	return expression.New("Primary", func() rql.ASTNode {
+		return Primary()
+	}).(rql.Primary)
+}
+
+// PE_ActionPredicate returns a node representing PE ActionPredicate
+func PE_ActionPredicate() rql.ActionPredicate {
+	return expression.New("ActionPredicate", func() rql.ASTNode {
+		return predicate.Action(plugin.Action{})
+	}).(rql.ActionPredicate)
+}
+
+// PE_StringPredicate returns a node representing PE StringPredicate
+func PE_StringPredicate() rql.StringPredicate {
+	return expression.New("StringPredicate", func() rql.ASTNode {
+		return predicate.String()
+	}).(rql.StringPredicate)
+}
+
+// PE_TimePredicate returns a node representing PE TimePredicate
+func PE_TimePredicate() rql.TimePredicate {
+	return expression.New("TimePredicate", func() rql.ASTNode {
+		return predicate.Time("", time.Time{})
+	}).(rql.TimePredicate)
+}
+
+// PE_UnsignedNumericPredicate returns a node representing PE UnsignedNumericPredicate
+func PE_UnsignedNumericPredicate() rql.NumericPredicate {
+	return expression.New("UnsignedNumericPredicate", func() rql.ASTNode {
+		return predicate.UnsignedNumeric("", decimal.Decimal{})
+	}).(rql.NumericPredicate)
+}
--- a/api/rql/interfaces.go
+++ b/api/rql/interfaces.go
@@ -1,0 +1,97 @@
+package rql
+
+import (
+	"time"
+
+	"github.com/puppetlabs/wash/plugin"
+	"github.com/shopspring/decimal"
+)
+
+// Primary is the interface implemented by all the primaries. A primary's
+// domain is the set of all entries that it applies to. The domain can be
+// specified at the instance-level (EntryInDomain), the schema-level
+// (EntrySchemaInDomain) or both.
+//
+// Note that a primary can either be an EntryPredicate, an EntrySchemaPredicate,
+// both or neither. In practice, the RQL will use EvalEntrySchema when pruning the
+// stree and EvalEntry when filtering the entries. EntryInDomain and EntrySchemaInDomain
+// are only needed to correctly negate the primaries (for without it, strict negation
+// would return true for entries that are outside the primary's domain).
+//
+// For a given primary p, here are the possible scenarios for EvalEntrySchema
+// and EvalEntry (including negation). Note that p.EntrySchemaInDomain and
+// p.EvalEntrySchema can assume that s != nil, and that evaluation order is from
+// left-to-right (so EntrySchemaInDomain first then EvalEntrySchema) with
+// appropriate '&&' short-circuiting.
+//   * If p implements EntrySchemaPredicate, then for RQL
+//         EvalEntrySchema(s) == p.EntrySchemaInDomain(s) && p.EvalEntrySchema(s)
+//         NOT(EvalEntrySchema(s)) == p.EntrySchemaInDomain(s) && !p.EvalEntrySchema(s)
+//     otherwise
+//         EvalEntrySchema(s) == p.EntrySchemaInDomain(s)
+//         NOT(EvalEntrySchema(s)) == p.EntrySchemaInDomain(s)
+//
+//   * If p implements EntryPredicate, then for RQL
+//         EvalEntry(e) == p.EntryInDomain(e) && p.EvalEntry(e)
+//         NOT(EvalEntry(e)) == p.EntryInDomain(e) && !p.EvalEntry(e)
+//     otherwise
+//         EvalEntry(e) == p.EntryInDomain(e)
+//         NOT(EvalEntry(e)) == p.EntryInDomain(e)
+//
+type Primary interface {
+	ASTNode
+	EntryInDomain(Entry) bool
+	EntrySchemaInDomain(*EntrySchema) bool
+}
+
+// EntryPredicate represents a predicate on an entry
+type EntryPredicate interface {
+	Primary
+	EvalEntry(Entry) bool
+}
+
+// EntrySchemaPredicate represents a predicate on an entry schema object
+type EntrySchemaPredicate interface {
+	Primary
+	EvalEntrySchema(*EntrySchema) bool
+}
+
+// ValuePredicate represents a predicate on a metadata (JSON) value. Its
+// domain is the set of all value types that the predicate applies to.
+//
+// In practice, the RQL will use EvalValue; ValueInDomain's only needed
+// to correctly negate the predicates (like EntryInDomain/EntrySchemaInDomain).
+// Here are the semantics for a given predicate. Note that evaluation order
+// is from left-to-right (so ValueInDomain first then EvalValue) with appropriate
+// '&&' short-circuiting.
+//     EvalValue(v) == p.ValueInDomain(v) && p.EvalValue(v)
+//     NOT(EvalValue(v)) == p.ValueInDomain(v) && !p.EvalValue(v)
+type ValuePredicate interface {
+	ASTNode
+	ValueInDomain(interface{}) bool
+	EvalValue(interface{}) bool
+}
+
+// StringPredicate represents a predicate on a string value
+type StringPredicate interface {
+	ASTNode
+	EvalString(string) bool
+}
+
+// NumericPredicate represents a predicate on a numeric value. The
+// decimal.Decimal type lets us handle arbitrarily large numbers.
+type NumericPredicate interface {
+	ASTNode
+	EvalNumeric(decimal.Decimal) bool
+}
+
+// TimePredicate represents a predicate on a time value.
+type TimePredicate interface {
+	ASTNode
+	EvalTime(time.Time) bool
+}
+
+// ActionPredicate represents a predicate on a Wash action.
+type ActionPredicate interface {
+	ASTNode
+	EvalAction(plugin.Action) bool
+}
--- a/api/rql/internal/errz/matchError.go
+++ b/api/rql/internal/errz/matchError.go
@@ -1,0 +1,38 @@
+package errz
+
+import "fmt"
+
+// MatchError represents the case when the input tokens did not
+// match a given node
+type MatchError struct {
+	reason string
+}
+
+func (m MatchError) Error() string {
+	return m.reason
+}
+
+// MatchErrorf creates a new MatchError object
+func MatchErrorf(format string, a ...interface{}) MatchError {
+	return MatchError{fmt.Sprintf(format, a...)}
+}
+
+// IsMatchError returns true if err is a MatchError,
+// false otherwise.
+func IsMatchError(err error) bool {
+	_, ok := err.(MatchError)
+	return ok
+}
+
+// IsSyntaxError returns true if err is a syntax error, false otherwise.
+func IsSyntaxError(err error) bool {
+	if err == nil {
+		return false
+	}
+	switch err.(type) {
+	case MatchError:
+		return false
+	default:
+		return true
+	}
+}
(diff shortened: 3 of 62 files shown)


api/rql/internal/predicate/expression/expression.go
```diff
@@ -0,0 +1,142 @@
...
+Note that the key difference between a PE and an RPE is that the NOT operator
+in an RPE can only be associated with Atoms instead of other RPEs. As an example,
+given the following PE
+	AND(OR(A1, NOT(A2)), NOT(OR(NOT(AND(A3, A4))), A5))
+Its corresponding RPE is
+	AND(OR(A1, NOT(A2)), AND(AND(A3, A4), NOT(A5)))
+where we used De'Morgan's law to distribute the NOT inside the OR, and noted
+that NOT(NOT(p)) == p.
+*/
+func reduce(exp rql.ASTNode) expressionNode {
```
ekinanp: Note that all the custom negation and De'Morgan's law crap in `wash find` was effectively doing this to the parsed predicate expression.

api/rql/internal/predicate/action_test.go
```diff
@@ -0,0 +1,65 @@
...
+	s.EATTC(a, plugin.ExecAction())
+}
+
+func (s *ActionTestSuite) TestExpression_AtomAndNot() {
+	expr := expression.New("action", func() rql.ASTNode {
+		return Action(plugin.Action{})
+	})
+
+	s.MUM(expr, "exec")
+	s.EAFTC(expr, plugin.ListAction())
```
ekinanp: Note that when testing `Not`, this becomes `EATTC`. This symmetry (switching `F` and `T` around) is common in all the other predicate and primary tests.

api/rql/internal/predicate/boolean_test.go
```diff
@@ -0,0 +1,120 @@
...
+		expr,
+		asttest.StringPredicateC,
+		asttest.NumericPredicateC,
+		asttest.TimePredicateC,
+		asttest.ActionPredicateC,
+	)
+
+	s.MUM(expr, []interface{}{"NOT", true})
+	s.EVTTC(expr, false)
+	s.EVFTC(expr, true, "foo")
```
ekinanp: Notice that `NOT` still returns false for a mis-typed value b/c that value's outside the predicate's domain.

api/rql/internal/predicate/numeric.go
```diff
@@ -0,0 +1,136 @@
...
+type numericValue struct {
+	numeric
+}
+
+func (p *numericValue) Marshal() interface{} {
+	return []interface{}{"number", p.numeric.Marshal()}
+}
+
+func (p *numericValue) Unmarshal(input interface{}) error {
+	if !matcher.Array(matcher.Value("number"))(input) {
```
ekinanp: I'm open to ideas on better centralizing all this redundant unmarshal validation. On my end, I haven't really been able to come up with a clean name for some helper/base type that does this, which is why I stuck with `matcher` as the (horrible) compromise.

api/rql/ast/predicateExpression.go
```diff
@@ -0,0 +1,46 @@
...
+	"time"
+
+	"github.com/puppetlabs/wash/api/rql"
+	"github.com/puppetlabs/wash/api/rql/internal/predicate"
+	"github.com/puppetlabs/wash/api/rql/internal/predicate/expression"
+	"github.com/puppetlabs/wash/plugin"
+	"github.com/shopspring/decimal"
+)
+
+// PE_Primary returns a node representing PE Primary
```
MikaelSmith: PE wasn't obvious to me, maybe write out `predicate expression primary` in the comment.

MikaelSmith: Just needs to be spelled out the first time it's used in a file.

api/rql/interfaces.go
```diff
@@ -0,0 +1,93 @@
...
+// both or neither. In practice, the RQL will use EvalEntrySchema when pruning the
+// stree and EvalEntry when filtering the entries. EntryInDomain and EntrySchemaInDomain
+// are only needed to correctly negate the primaries (for without it, strict negation
+// would return true for entries that are outside the primary's domain).
+//
+// For a given primary p, here are the possible scenarios for EvalEntrySchema
+// and EvalEntry (including negation). Note that p.EntrySchemaInDomain and
+// p.EvalEntrySchema can assume that s != nil.
+//   * If p implements EntrySchemaPredicate, then for RQL
+//         EvalEntrySchema(s) == p.EntrySchemaInDomain(s) && p.EvalEntrySchema(s)
```
MikaelSmith: I'm not clear what the evaluation order is in this expression.

api/rql/internal/errz/matchError.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// MatchErrorf creates a new MatchError object
+func MatchErrorf(format string, a ...interface{}) MatchError {
+	return MatchError{fmt.Sprintf(format, a...)}
+}
+
+// IsMatchError returns true if err is a MatchError,
+// false otherwise.
+func IsMatchError(err error) bool {
```
MikaelSmith: You can just use `errors.As(err, MatchError{})` in Go 1.13.

api/rql/interfaces.go
ekinanp: Should be `InDomain` first then the `Eval` method w/ appropriate short-circuiting. I'll clarify that.

api/rql/ast/predicateExpression.go
ekinanp: Agreed. I'll update.

api/rql/internal/nonterminalNode.go
```diff
@@ -0,0 +1,64 @@
...
+// more other nodes when it is unmarshaled. MatchedNode returns
+// the matched node
+type NonterminalNode interface {
+	rql.ASTNode
+	MatchedNode() rql.ASTNode
+	SetMatchedNode(rql.ASTNode) NonterminalNode
+	SetMatchErrMsg(msg string) NonterminalNode
+}
+
+func NewNonterminalNode(n rql.ASTNode, ns ...rql.ASTNode) NonterminalNode {
```
MikaelSmith: Why two arguments? I guess that makes it so you need at least 1?

api/rql/internal/predicate/expression/atom.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"time"
+
+	"github.com/puppetlabs/wash/api/rql"
+	"github.com/puppetlabs/wash/plugin"
+	"github.com/shopspring/decimal"
+)
+
+func Atom(p rql.ASTNode) rql.ASTNode {
```
MikaelSmith: I'm not clear what an Atom is in this context.

api/rql/interfaces.go
ekinanp: Updated.

api/rql/ast/predicateExpression.go
ekinanp: Updated.

api/rql/interfaces.go
MikaelSmith: Parentheses would have also worked.

api/rql/internal/predicate/numeric.go
```diff
@@ -0,0 +1,136 @@
...
+		return fmt.Errorf("missing the numeric predicate")
+	}
+	if err := p.numeric.Unmarshal(array[1]); err != nil {
+		return fmt.Errorf("%w", err)
+	}
+	return nil
+}
+
+func (p *numericValue) ValueInDomain(v interface{}) bool {
+	// TODO: Support stringified numbers?
```
MikaelSmith: I don't have a reason to do so.

MikaelSmith: Or was this a reference to SI suffixes? If so, please provide an example.

api/rql/internal/predicate/string.go
```diff
@@ -0,0 +1,234 @@
...
+
+func String() rql.StringPredicate {
+	p := &stringP{
+		NonterminalNode: internal.NewNonterminalNode(
+			StringGlob(""),
+			StringRegex(nil),
+			StringEqual(""),
+		),
+	}
+	p.SetMatchErrMsg("must be formatted as either ['glob', <glob>], ['regex', <regex>], or ['=', <str>]")
```
MikaelSmith: Why not use `EQL` here as well? You went to the effort of making it a constant in lines 123 and 127.

api/rql/internal/predicate/time.go
```diff
@@ -0,0 +1,121 @@
...
+		op: op,
+		t:  t,
+	}
+}
+
+// tm => time. Have to name it as such to avoid conflicting with
+// the 'time' package
+type tm struct {
+	op ComparisonOp
+	t  time.Time
```
MikaelSmith: So this goes with absolute time rather than `time.Duration` from a reference point?

api/rql/internal/primary/atime_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/puppetlabs/wash/api/rql/internal/predicate"
+	"github.com/puppetlabs/wash/api/rql/internal/predicate/expression"
+	"github.com/stretchr/testify/suite"
+)
+
+type AtimeTestSuite struct {
+	asttest.Suite
+}
+
+func (s *AtimeTestSuite) TestMarshal() {
```
MikaelSmith: Could be worth a mini test suite for atime/ctime/mtime/crtime to avoid duplicating all the same testing.

MikaelSmith: You could probably make it a test suite that you instantiate with the particular primary name and type.

MikaelSmith (changes requested):

MikaelSmith:
A few tests seem to be failing.

api/rql/internal/nonterminalNode.go
ekinanp: Yup. I'll comment that.

api/rql/internal/predicate/expression/atom.go
ekinanp: Hmm, not sure what you mean. Are you looking for a comment like

> Atom wraps p into a type that works with predicate expressions. The RQL will use this to call p's appropriate Eval* methods, where each Eval* method implements its interface-specific semantics. For example, if p is a Primary _and_ EntryPredicate, then the returned atom's EvalEntry method will be evaluated as a.EvalEntry(e) == p.EntryInDomain(e) && p.EvalEntry(e). Front-end interfaces to the RQL should _always_ use the Atom type when testing their parsed predicates to ensure correct evaluation semantics.

Note that I plan on exporting functions like `Atom`, `Not` in the `ast` package once I get to refactoring `find`'s parsing.

api/rql/internal/predicate/numeric.go
ekinanp: Nah, just stringified numbers. Like "1", "2", "3". Could probably wait until a user asks for it.

api/rql/internal/predicate/string.go
ekinanp: Went back to just using "=" because comparisonOp caused some annoying test failures. The constant was originally so `Time`, `Numeric` would get the right comparisonOp; that security is not needed since string equal has a helper, `StringEqual` that provides that.

api/rql/internal/predicate/time.go
ekinanp: Yes. I feel like that's more idiomatic with other filtering languages (e.g. I found "return log files that were modified after time X" to be more common).

`wash find` can still implement is past/future time semantics. We'd just create a wrapper to a TimePredicate that would add `[AND, [<=, <ref_time>], <time_predicate>]` to whatever predicate was parsed. Negation can also do a similar thing.

api/rql/internal/primary/atime_test.go
ekinanp: Yeah, I think that's a good idea. I'll go ahead and do that.

api/rql/internal/predicate/time.go
ekinanp: If past/future time semantics become common in other front-end interfaces, RQL could probably be updated to support that so people don't have to keep writing wrappers around it.

api/rql/internal/errz/matchError.go
ekinanp: Hmm, that doesn't quite work because it returns true if `err`'s chain contains first instance of MatchError. Some parts of the RQL code wrap match errors, see e.g. https://github.com/ekinanp/wash/blob/rql/api/rql/internal/predicate/numeric.go#L121, so that they could be returned as syntax errors. There, `errors.As` will return true.

I think `errors.As`/`errors.Is` should have been named to something like `errors.Contains`. Since `Contains` is different from `IsMatchError`, I'm opting to keep the helper.

I did find that `IsSyntaxError` is not being used anywhere, so I'll go ahead and remove that.

ekinanp: Updated.

api/rql/internal/nonterminalNode.go
ekinanp: Updated

api/rql/internal/predicate/expression/atom.go
ekinanp: Updated since I feel like that comment would be useful anyways

api/rql/internal/predicate/string.go
ekinanp: Updated

api/rql/internal/primary/atime_test.go
ekinanp: Updated

api/rql/internal/predicate/expression/atom.go
MikaelSmith: I'm not sure what an atom is supposed to represent.

api/rql/internal/predicate/time.go
MikaelSmith: I think that makes sense.

api/rql/internal/predicate/expression/expression.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/puppetlabs/wash/api/rql/internal/errz"
+	"github.com/puppetlabs/wash/plugin"
+	"github.com/shopspring/decimal"
+)
+
+type PtypeGenerator func() rql.ASTNode
+
+/*
+New returns a new predicate expression of 'ptype' predicates (PE).
+The AtomGenerator should generate "emtpy" structs representing
```
MikaelSmith: Typo in `emtpy` (should be `empty`?)

api/rql/internal/predicate/string.go
```diff
@@ -0,0 +1,234 @@
...
+	s string
+}
+
+func (p *stringEqual) Marshal() interface{} {
+	return []interface{}{"=", p.s}
+}
+
+func (p *stringEqual) Unmarshal(input interface{}) error {
+	if !matcher.Array(matcher.Value("="))(input) {
+		return errz.MatchErrorf("must be formatted as ['%v', <str>]", "=")
```
MikaelSmith: There's no reason to do string substitution here.

api/rql/internal/predicate/string.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	array := input.([]interface{})
+	if len(array) > 2 {
+		return fmt.Errorf("must be formatted as ['%v', <str>]", "=")
```
MikaelSmith: Same here.

api/rql/internal/predicate/expression/atom.go
MikaelSmith: Ok, I guess I kind of understand.

MikaelSmith (approved):
Just very minor comments remaining.

ekinanp:
Updated.

MikaelSmith (approved):

PR #706, opened 2020-02-03
Fix some bugs in docs and external plugin entry schemas

ekinanp:


cmd/docs.go
```diff
@@ -57,11 +58,14 @@ func docsMain(cmd *cobra.Command, args []string) exitCode {
...
-	//   SUPPORTED_ATTRIBUTES
+	//   SUPPORTED ATTRIBUTES
 	//     * <attribute> (<full_name_of_attribute>)
 	//
 	//   <description that talks about attributes/metadata and shows off 'meta'>
-	if len(entry.Attributes.ToMap()) > 0 {
+	//
+	// if the entry has any supported attributes. Otherwise, it prints an appropriate
+	// note then prints the attributes/metadata description.
+	if len(entry.Metadata) > 0 {
```
MikaelSmith: Could this go wrong if the entry has some attributes but no metadata?

MikaelSmith: It looks like if that's true, we'll return attributes as metadata. And if there is metadata, you specifically get attributes. So seems like this will work but is slightly confusing.

MikaelSmith (approved):

cmd/docs.go
ekinanp: Ah yeah, I'll add a clarifying comment. Thanks for bringing it up,

ekinanp: Updated.

PR #707, opened 2020-02-03
Go back to show-casing screencasts one-by-one

ekinanp:

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -28,61 +28,47 @@ $.get("https://api.github.com/repos/puppetlabs/wash/releases/latest", function(d
 <p class="intro-large">See for yourself</p>
 
 <!-- Display the demos -->
-<div class="demo-select">
-<select id="demo">
-  <option value="aws">AWS</option>
-  <option value="gcp">GCP</option>
-  <option value="kubernetes">Kubernetes</option>
-  <option value="docker">Docker</option>
-  <option value="external-plugin">Other</option>
-</select>
-</div>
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -28,61 +28,47 @@ $.get("https://api.github.com/repos/puppetlabs/wash/releases/latest", function(d
...
-<select id="demo">
-  <option value="aws">AWS</option>
-  <option value="gcp">GCP</option>
-  <option value="kubernetes">Kubernetes</option>
-  <option value="docker">Docker</option>
-  <option value="external-plugin">Other</option>
-</select>
-</div>
-<div id="aws-demo" style="display: none;">
+# **AWS**
```
ekinanp: I showed two of my friends the website (one of them isn't in tech) and they both couldn't figure out that changing the screencast means clicking the dropdown. So that means the dropdown isn't that obvious.

We could try the tabulated format, but I worry that could cause the same problem -- lazy user doesn't realize that clicking the tab changes the video.

Anyways, I went back to listing the videos one after the other, but can understand the concerns of doing that so made this a draft PR. For what it's worth, it doesn't look too bad.

MikaelSmith:
This seems ok. With the videos at the end they can just scroll to the one they want and we don't have to worry about them missing something at the bottom.

ekinanp:
Ok, made it an official PR.

MikaelSmith (approved):

PR #708, opened 2020-02-03
Test with -race in CI

MikaelSmith:
Attempt to detect race conditions in CI. Make a test timeout longer to account for slower speed with the `-race` flag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
 jobs:
   include:
   - name: Test Go stable
+    script: go test -race ./...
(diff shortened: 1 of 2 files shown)


ekinanp:
Didn't we try this before and ended up running into some weird issues? Curious what changed your mind.

MikaelSmith:
I added it back because an issue came up that `-race` detected. I had this branch lying around and couldn't remember why we hadn't added it.

ekinanp (approved):

PR #709, opened 2020-02-05
Support commands against Windows targets in external plugins

### LINKED ISSUE #699: Interact with Windows systems, part 1
The Exec primitive itself assumes SSH for VMs, and the tools we've built that run commands over Exec - list filesystems and processes - all assume POSIX environments. We should expand this to work with Windows VMs.
- [x] Add WinRM support to boltwash (https://github.com/puppetlabs/boltwash/pull/3)
- [ ] Provide a way to identify whether a target accepts PowerShell or POSIX Shell (i.e. bash/ash/busybox) everywhere we run shell commands, and provide Powershell equivalents to current behavior (#709)
- [ ] Update boltwash to identify WinRM targets as using Powershell (https://github.com/puppetlabs/boltwash/pull/4)
- [ ] Document `os.login_shell` attribute and how it affects `ps`/`volume.FS` in external plugins

MikaelSmith:
Adds support for executing powershell commands instead of POSIX when the `os.login_shell` attribute is set to `powershell`.

Resolves #699.

### DIFF
--- a/cmd/ps.go
+++ b/cmd/ps.go
@@ -122,16 +122,16 @@ func psMain(cmd *cobra.Command, args []string) exitCode {
 			if err != nil {
 				cmdutil.ErrPrintf("errored on %v: %v\n", k, err)
 			}
-			var shell plugin.LoginShell
-			if entry.Attributes.HasLoginShell() {
-				shell = entry.Attributes.LoginShell()
+			var loginShell plugin.Shell
+			if entry.Attributes.HasOS() {
+				loginShell = entry.Attributes.OS().LoginShell
 			}
-			if shell == plugin.UnknownShell {
+			if loginShell == plugin.UnknownShell {
 				// Assume posix if unknown
-				shell = plugin.POSIXShell
+				loginShell = plugin.POSIXShell
 			}
 
-			dispatcher := dispatchers[shell]
+			dispatcher := dispatchers[loginShell]
 			ch, err := dispatcher.execPS(conn, k)
 			if err != nil {
 				cmdutil.ErrPrintf("errored on %v: %v\n", k, err)
--- a/cmd/ps.go
+++ b/cmd/ps.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"bufio"
+	"encoding/csv"
 	"fmt"
 	"os"
 	"strconv"
@@ -11,8 +12,10 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/puppetlabs/wash/api/client"
 	apitypes "github.com/puppetlabs/wash/api/types"
 	cmdutil "github.com/puppetlabs/wash/cmd/util"
+	"github.com/puppetlabs/wash/plugin"
 )
 
 func psCommand() *cobra.Command {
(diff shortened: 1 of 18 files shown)


volume/fs.go
```diff
@@ -95,7 +95,7 @@ func exec(ctx context.Context, executor plugin.Execable, cmdline []string, tty b
 
 // VolumeList satisfies the Interface required by List to enumerate files.
 func (d *FS) VolumeList(ctx context.Context, path string) (DirMap, error) {
-	cmdline := StatCmd(path, d.maxdepth)
+	cmdline := d.selectShellCommand(StatCmd(path, d.maxdepth), ChildItemsCmd(path, d.maxdepth))
```
MikaelSmith: TODO: These changes should have tests associated with them.

cmd/ps.go
```diff
@@ -112,45 +114,80 @@ func parseEntry(line string) (psresult, error) {
 	return psresult{pid: pid, active: activeTime, command: command}, nil
 }
 
-func parseLines(node string, chunk string) []psresult {
+func parseStatLines(chunk string) ([]psresult, error) {
```
ekinanp: Might help to re-arrange some of these functions as something like

```
// PS POSIX
<posix stuff>

// PS POWERSHELL
<powershell stuff>
```

so a maintainer can immediately see what's what.

cmd/ps.go
```diff
@@ -183,27 +220,61 @@ func psMain(cmd *cobra.Command, args []string) exitCode {
...
-		stats = append(stats, results[path]...)
-	}
-
-	cmdutil.Print(formatStats(stats))
+	cmdutil.Print(formatStats(paths, results))
 	return exitCode{0}
 }
+
+var dispatchers = []struct {
+	exec  func(client.Client, string) (<-chan apitypes.ExecPacket, error)
```
ekinanp: Nit: `execPS`, `parseOutput`? Slightly clearer.

plugin/entryAttributes.go
```diff
@@ -55,6 +85,8 @@ type EntryAttributes struct {
 	mtime   time.Time
 	ctime   time.Time
 	crtime  time.Time
+	os      OS
+	hasOS   bool
```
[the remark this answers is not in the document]
MikaelSmith: Some things will not have any OS information because they're not a computer. I think it makes sense to reflect that directly, rather than having an empty struct.

volume/items.go
```diff
@@ -0,0 +1,119 @@
...
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/puppetlabs/wash/plugin"
+)
+
+// ChildItemsCmd returns the PowerShell command required to stat all the files in a directory up to maxdepth.
+func ChildItemsCmd(path string, maxdepth int) []string {
```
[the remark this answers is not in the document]
MikaelSmith: Yeah, I thought about something like that but hadn't decided on names. These work.

plugin/entryAttributes.go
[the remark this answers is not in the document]
ekinanp: Also makes sense.

MikaelSmith:
Updated, added tests for volume.FS.

volume/stat.go
```diff
@@ -1,94 +1,51 @@
...
-	"github.com/puppetlabs/wash/munge"
 	"github.com/puppetlabs/wash/plugin"
 )
 
-// StatCmd returns the command required to stat all the files in a directory up to maxdepth.
-func StatCmd(path string, maxdepth int) []string {
-	// List uses "" to mean root. Translate for executing on the target.
-	if path == RootPath {
-		path = "/"
+func addAttributesForPath(dirmap DirMap, attr plugin.EntryAttributes, base, fullpath string, maxdepth int) {
```
ekinanp: Thanks, this helped a bit. I think this could be `addEntryToDirMap`; `attr` => `entryAttr`, `fullpath` => `fullEntryPath` (or `absEntryPath` for consistency?), then `relative` => `relativeEntryPath`

ekinanp (approved):
Comment's not a blocker, but I think addressing it would help w/ clarity a bit.

volume/stat.go
MikaelSmith: But none of the things it deals with are an Entry. Just the path representing one.

MikaelSmith:
Going to add some external plugin docs in a separate PR.

volume/stat.go
ekinanp: Yeah, it's subjective. If I read that, I'd think of it as "addEntryToDirMap" => "where's the entry ...?" => "Oh the path is the entry. Ok this makes sense now".

ekinanp (approved):

PR #722, opened 2020-02-07
Handle buffering incomplete lines in Stream output

### LINKED ISSUE #721: Non-line-buffered output to 'tail -f' results in extra lines
With the fix for https://github.com/puppetlabs/wash-ruby/issues/32, external plugins using that gem will no longer line-buffer output. That can result in getting writes that look like
```
write("a string of text")
write("\r\n")
```
The tail command was written assuming all output would be line-buffered, but that's not very flexible. It should handle aggregating lines when they're broken up.

MikaelSmith:
Fixes #721.

### DIFF
--- a/cmd/tail.go
+++ b/cmd/tail.go
@@ -1,11 +1,11 @@
 package cmd
 
 import (
+	"bytes"
 	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
-	"strings"
 	"time"
 
 	"github.com/Benchkram/errz"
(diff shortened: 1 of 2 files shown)


cmd/tail_test.go
```diff
@@ -0,0 +1,121 @@
...
+	n, err = lw.Write([]byte(msg[split:]))
+	assert.NoError(t, err)
+	assert.Equal(t, len(msg)-split, n)
+	n, err = lw.Write([]byte{'\r', '\n'})
+	assert.NoError(t, err)
+	assert.Equal(t, 2, n)
+	mark = time.Now()
+	lw.Finish()
+	ln, ok = <-out
+	assert.True(t, ok)
```
ekinanp: There's a lot of copy-paste assertions here. Could these be more higher-level, e.g. `assertLine` or something? Not a PR blocker, but if it's easy to do then it would make the tests readable.

cmd/tail.go
```diff
@@ -35,15 +35,41 @@ type line struct {
...
+	}
+
+	text := w.buf.Next(i)
+	// Consume \r, \n, or \r\n. Note that the Buffer takes care of re-using space when we catch up.
+	crOrLf, err := w.buf.ReadByte()
+	if err != nil {
+		return len(b), err
+	}
+
+	// If the last character was LF, check whether the character before was CR and if so remove it.
```
ekinanp: Why do we need to remove CR? Comment would help. Might also help to place this code closer to the "Consume \r" comment since this seems to be handling the `\r\n` case.

cmd/tail.go
```diff
@@ -35,15 +35,41 @@ type line struct {
...
 	name string
+	buf  bytes.Buffer
 	out  chan line
 }
 
-func (w lineWriter) Write(b []byte) (int, error) {
-	s := strings.TrimSuffix(string(b), "\n")
-	w.out <- line{Line: tail.Line{Text: s, Time: time.Now()}, source: w.name}
+func (w *lineWriter) Write(b []byte) (int, error) {
+	// Aggregate lines, then submit all completed lines to the output channel.
```
ekinanp: I'd probably add a node here so something like

> Note: Only completed lines are written to the output channel. For incomplete lines, we just return the # of written bytes

cmd/tail.go
```diff
@@ -35,15 +35,41 @@ type line struct {
...
+	w.buf.Write(b)
+	i := bytes.LastIndexAny(w.buf.Bytes(), "\r\n")
+	if i == -1 {
+		// Always return that we wrote all the bytes.
```
ekinanp: Comment like "Incomplete line, so just return" would help

cmd/tail.go
```diff
@@ -35,15 +35,41 @@ type line struct {
...
+		return len(b), nil
+	}
```
ekinanp: Similarly, comment like "Complete line, so do some cleanup" would also help here.

ekinanp (commented):
Few comments to make the code clearer -- it took me a bit to figure out what was going on here.

Also, the test comment is not at all a blocker.

cmd/tail.go
MikaelSmith: Can't really put it closer, but I'll update the comment.

ekinanp (approved):

PR #732, opened 2020-02-25
Tests for Query + improved error messages + architecture cleanup

ekinanp:


api/rql/internal/predicate/expression/not.go
```diff
@@ -44,32 +44,16 @@ func (n *not) Unmarshal(input interface{}) error {
...
-	a := n.p.(*atom)
-	result := a.p.(rql.Primary).EntryInDomain(e)
-	if ep, ok := a.p.(rql.EntryPredicate); ok {
-		result = result && !ep.EvalEntry(e)
-	}
-	return result
+	return !n.p.(rql.EntryPredicate).EvalEntry(e)
 }
 
 func (n *not) EvalEntrySchema(s *rql.EntrySchema) bool {
```
ekinanp: This is still a bit of an issue b/c something like `["NOT", ["path", <string_p>]]` would cause the SchemaP to return false. Again, we could document that negation only makes sense for the BooleanPredicate, and to not negate anything else. Could also enforce this in the AST by defining PEs as being made up of only AND/OR, and NPEs (negatable PE) as being made up of AND/OR/NOT. Then the RQL could be something like

```
Query := PE Primary
...
NamePrimary => ["name", NPE StringPredicate]
...
```

Not sure which approach is best. Keeping everything as PEs results in simpler code, but some (possibly) annoying edge cases/debugging. SO maybe NPEs are a good idea.

MikaelSmith: Annoying edge cases/debugging are also code, so probably debatable whether it's simpler. NPE sounds good for clarity.

Will `["NOT", ["path", <string_p>]]` now cause an error?

api/rql/internal/primary/cname_test.go
```diff
@@ -77,7 +67,7 @@ func (s *CNameTestSuite) TestExpression_AtomAndNot() {
 	e.CName = "foo"
 	s.EEFTC(expr, e)
 
-	s.EESTTC(expr, schema)
+	//s.EESTTC(expr, schema)
```
MikaelSmith: Why is this commented out?

api/rql/internal/primary/name_test.go
```diff
@@ -77,7 +67,7 @@ func (s *NameTestSuite) TestExpression_AtomAndNot() {
 	e.Name = "foo"
 	s.EEFTC(expr, e)
 
-	s.EESTTC(expr, schema)
+	//s.EESTTC(expr, schema)
```
MikaelSmith: Commented out?

api/rql/internal/primary/path_test.go
```diff
@@ -72,14 +64,12 @@ func (s *PathTestSuite) TestExpression_AtomAndNot() {
...
 	s.MUM(expr, []interface{}{"NOT", []interface{}{"path", []interface{}{"glob", "foo"}}})
-	e.Path = ""
-	s.EEFTC(expr, e)
 	e.Path = "bar"
 	s.EETTC(expr, e)
 	e.Path = "foo"
 	s.EEFTC(expr, e)
 
-	s.EESTTC(expr, schema)
+	//s.EESTTC(expr, schema)
```
MikaelSmith: Commented?

api/rql/internal/primary/timeAttr_test.go
```diff
@@ -82,5 +72,5 @@ func (s *TimeAttrTestSuite) TestExpression_AtomAndNot() {
 	s.setAttr(&e, s.TM(500))
 	s.EEFTC(expr, e)
 
-	s.EESTTC(expr, schema)
+	//s.EESTTC(expr, schema)
```
MikaelSmith: Commented?

MikaelSmith (changes requested):

api/rql/internal/predicate/expression/not.go
ekinanp: Yes it will. But `["path", [NOT, <string_p>]]` wont. And that kind of makes sense, you are querying the "path" field and that has a predicate assigned. So NPE seems like the right direction.

ekinanp: I'm also wondering whether the `true/false` primaries are necessary.. Can't really think of a good reason to include them. I kind of wonder if they were there b/c they made it easy to test the operators back in the old days, where writing unit tests was a lot harder.

ekinanp: `true`/`false` definitely make sense as value predicates though.

ekinanp: So another thing that's nice about this is negation matches user intent. So something like `["NOT", ["object", [["key", "foo"], p]]` reads "return true if NOT an object w/ key foo whose value satisfies p", while `["object", [["key", "foo"], ["NOT, p]]]` reads "return true if obj w/ key foo whose value satisfies NOT(p)".

Similarly, `["NOT", ["number, p]]` vs. `["number", [NOT, p]]` -- former reads "return true if _not_ number" satisfying p, other is "return true if number satisying NOT p".

`wash find` will probably still need its custom negation semantics, but that's b/c we're assuming user intent is always to trickle-down the NOT. Different front ends can change their parsing rules to accommodate different user intent, but the RQL's user intent is still generic.

MikaelSmith: I think reducing the number of ways to express the same thing makes sense.

ekinanp: Meh, may also be good enough to keep `wash find` symmetric with RQL once that refactor's done.

api/rql/internal/primary/timeAttr_test.go
ekinanp: Ah yeah, I wanted to hold off deleting this until NPE/PE was resolved. Since NPE's are OK, I'll remove this commented stuff.

ekinanp: Updated.

MikaelSmith (approved):

PR #733, opened 2020-02-25
Change MFA prompt, include wps in docs examples

ekinanp:


plugin/aws/profile.go
```diff
@@ -29,7 +29,7 @@ func newProfile(ctx context.Context, name string) (*profile, error) {
 
 	// profile-specific stdin prompt
 	tokenProvider := func() (string, error) {
-		return plugin.Prompt(fmt.Sprintf("Assume ROLE MFA token code for %v", name))
+		return plugin.Prompt(fmt.Sprintf("Enter ROLE MFA token code for %v", name))
```
MikaelSmith: This is clearer that it's a request. Might be clearer what you're enteringif if we said `Enter MFA token code for role %v`.

MikaelSmith (approved):

plugin/aws/profile.go
ekinanp: Makes sense. Updated.

plugin/aws/profile.go
```diff
@@ -29,7 +29,7 @@ func newProfile(ctx context.Context, name string) (*profile, error) {
 
 	// profile-specific stdin prompt
 	tokenProvider := func() (string, error) {
-		return plugin.Prompt(fmt.Sprintf("Assume ROLE MFA token code for %v", name))
+		return plugin.Prompt(fmt.Sprintf("Enter MFA token code for role %v", name))
```
MikaelSmith: Hmm, actually the name is of the profile. Maybe we need to mention role.

ekinanp: Updated again

MikaelSmith (approved):

plugin/aws/profile.go
MikaelSmith: Ugh, I meant to say "don't need to", but what's here seems fine.

ekinanp: No worries. Updated again.

MikaelSmith (approved):

PR #737, opened 2020-02-27
Refactor sidebar and docs

MikaelSmith:
Extract header and sidebar for modularity and to make them easier to read. Move more of docs navigation to the sidebar.


MikaelSmith:
This you definitely need to preview with Jekyll.

docs/docs.md
```diff
@@ -0,0 +1,3 @@
+---
+title: Docs
+---
```
MikaelSmith: Not sure what content we should put here. Most basic would be a list reflecting the links in the sidebar.

ekinanp: Maybe? Do we need to put something here?

ekinanp:
Kind of leaning towards flattening out "Concepts".

MikaelSmith:
They all felt related to me. Have an idea how you'd organize that?

ekinanp:
I was thinking just put it top-level. So under `Docs` you'd have a link to `CName`, link to `Attributes`, link to `Actions`.

MikaelSmith:
Maybe Concepts needs a preamble. I still feel like actions/cname/attributes are best understood as a set of concepts rather than distinct topics.

ekinanp:
I guess. I was kind of thinking of it as a user saying "Ok.. What were the supported actions?" => "Ah, I can see that as "Actions" under Docs", etc and that being the common case. But I'm not against what you have either, so up to you.

MikaelSmith:
I lean conservative for now, and we can change it later once we get more feedback.

ekinanp (approved):

PR #739, opened 2020-02-27
Include core plugin entry sgraph in external plugin entry sgraph

ekinanp:


ekinanp:
I'm going to add tests for this PR later today, just wanted to put soln. up. Main thing is we have no choice but to include the raw type IDs of core plugin entries since that's what their `List` methods will return. Hopefully a raw type ID doesn't conflict with anything in an external plugin. If that's an issue, we could probably just prefix + suffix that with `__` in https://github.com/puppetlabs/wash/blob/master/plugin/entrySchema.go#L369

ekinanp:
Updated with tests. This PR is ready for review.

MikaelSmith (approved):

PR #743, opened 2020-03-02
Add the meta primary

ekinanp:


ekinanp:
So making Object/Array predicate accept an `NPE ValuePredicate` gives the RQL some generality that `wash find` doesn't have (e.g. it paves the way for people to construct predicates on value _types_), but it also makes metadata schema stuff hard. Having Object/Array predicate accept `PE ValuePredicate` means we can mirror the metadata schema stuff in `find` here (by importing that key sequence stuff). But we also lose that type-checking generality.

Metadata schemas were built before the `kind` primary; with `kind`, metadata schemas aren't really necessary when it comes to optimizing find. I guess they are useful for making something like `find -m .tags[?] .key termination_date -exists` expressive if the CWD is the `instances` directory so meh (other way is to write `find -k '*instance' -m .tags[?] .key termination_date -exists` but that can be a bit verbose).

I like keeping the RQL general, but maybe that generality isn't necessary. E.g. when would you ever care about a metadata value's type? I guess if you were encoding type-values as strings (like we do for `termination_date` tag?). Even then, why not just query a specific value?

So seems like `PE ValuePredicate` seems like a good idea.

ekinanp:
Actually ignore my earlier comment. I think there's a way I could do the schemaP while still preserving the current grammar.

ekinanp:
I'm going to tack on the metadata schema stuff here.

api/rql/internal/predicate/boolean.go
```diff
@@ -16,11 +19,21 @@ type boolean struct {
...
 
 func (p *boolean) Marshal() interface{} {
-	return p.val
+	return []interface{}{"boolean", p.val}
 }
 
 func (p *boolean) Unmarshal(input interface{}) error {
-	val, ok := input.(bool)
+	if !matcher.Array(matcher.Value("boolean"))(input) {
+		return errz.MatchErrorf("must be formatted as ['boolean', <value>]")
```
MikaelSmith: I'm not really sure why the type name is needed, as JSON has boolean types.

api/rql/ast/metaPrimaryRealWorld_test.go
```diff
@@ -0,0 +1,170 @@
...
+	s.TTC("architecture",
+		s.A("string",
+			s.A("AND",
+				s.A("=", "x86_64"),
+				s.A("glob", "x86*"),
+			),
+		),
+	)
+
+	s.TTC("amiLaunchIndex", s.A("OR", nil, s.A("NOT", nil)))
```
MikaelSmith: So, nil or not nil? That should always pass no matter what's in the file, right?

api/rql/internal/predicate/array.go
```diff
@@ -0,0 +1,124 @@
...
+	default:
+		// Should never hit this code-path
+		panic(fmt.Sprintf("Unknown selector %T", p.selector))
+	}
+	return []interface{}{marshalledSelector, p.p.Marshal()}
+}
+
+func (p *arrayElement) Unmarshal(input interface{}) error {
+	array, ok := input.([]interface{})
+	formatErrMsg := "element predicate: must be formatted as [<element_selector>, NPE ValuePredicate]"
```
MikaelSmith: Where would they find a list of valid element selectors?

api/rql/internal/predicate/array.go
```diff
@@ -0,0 +1,124 @@
...
+		if firstElem != "some" && firstElem != "all" {
+			return errz.MatchErrorf(formatErrMsg)
+		}
+		if firstElem == "some" {
+			p.selector = some
+		} else {
+			p.selector = all
+		}
+	} else {
+		firstElem, ok := array[0].(float64)
```
MikaelSmith: Why float64 if we want an integer? Is that because it was unmarshaled into `interface{}` so it did it as an untyped number?

api/rql/internal/primary/meta/testdata/README.md
```diff
@@ -0,0 +1,21 @@
+`before_munging.json` was generated by running the following type through the [jsonschema](https://github.com/alecthomas/jsonschema) library (specifically ekinanp's fork):
```
MikaelSmith: Are you working on upstreaming these?

api/rql/internal/primary/meta/schema.go
```diff
@@ -0,0 +1,99 @@
...
+	loader gojsonschema.JSONLoader
+}
+
+type SchemaPredicate = func(Schema) bool
+
+func newSchema(s *plugin.JSONSchema) Schema {
+	// Given a value schema like ".key1.key2 5", the schema predicate's job is to check
+	// that m['key1']['key2'] == primitive_value. To do this correctly, we need to munge s
+	// a bit before returning our schema instance. That's what this code does.
+	var mungeProperties func(map[string]*jsonschema.Type) map[string]*jsonschema.Type
```
MikaelSmith: Not technically needed because `mungeProperties` isn't recursive.

api/rql/internal/primary/meta/testdata/README.md
MikaelSmith: Looks like schema_test.go mentions the issue created to start that process.

api/rql/internal/primary/meta/valueSchema.go
```diff
@@ -0,0 +1,101 @@
+package meta
+
+import (
+	"fmt"
+	"strings"
+)
+
+/*
+ValueSchema represents a metadata value's schema. It is used to generate
```
MikaelSmith: Looks like all the Schema and ValueSchema stuff in meta is not yet used.

api/rql/internal/predicate/array.go
ekinanp: Oh, `json.Unmarshal` unmarshals all numbers to float64 so that's why it is here.

api/rql/internal/predicate/array.go
ekinanp: The documentation. I do plan on putting out better error messages

api/rql/ast/metaPrimaryRealWorld_test.go
ekinanp: Yeah, this is really testing existence. I'll add a comment about that once the schema stuff's done.

api/rql/internal/primary/meta/valueSchema.go
ekinanp: Yeah, my annoying habit of pushing stuff up before it is fully finished. I'll comment when the schema stuff's done.

api/rql/internal/predicate/boolean.go
ekinanp: Keep symmetric w/ the other types (only if said type has more than one possible value). We have "string", "number", "object", "array". "boolean" has two values true/false, so it made sense to me to phrase it that way. `nil` has only one value, so that's why it's labeled as such.

I guess I can go back to `true`/`false` if that's preferred too.

MikaelSmith: I think I'd like to see all of those as a set of examples before deciding. All of those have types in JSON, so I don't remember why we need to call out their types.

ekinanp:
Still working on metadata schemas. I pushed up an update that slightly changed the metadata schema framework; the change came as a result of me trying to formalize all the semantics as can be seen in `meta/valuePredicate.go`.

api/rql/internal/primary/meta/schema.go
ekinanp: Yes, but opted to keep it because code looks a little prettier.

ekinanp:
Ok, this is ready for review. I split up all the schema work into their own commits.

I think I'm leaning towards changing `["boolean", <value>]` back to just `<value>` since JSON does have Boolean types. I'll wait once all the schema stuff's reviewed before I do that.

api/rql/internal/primary/meta/testdata/README.md
ekinanp: Yeah, haven't gotten around to it.

MikaelSmith (changes requested):

api/rql/internal/primary/meta/valuePredicate.go
```diff
@@ -0,0 +1,98 @@
...
+	  And {"foo": [5]} is a satisfying value for the leaf-AST's predicate. In fact, all
+	  satisfying values will have the general form {"foo": [nil]} (primitive types are
+	  normalized to "nil"). We can write this as ".foo[] nil".
+
+	* ".foo[] *" for the "NOT" node. To see why, notice that the leaf AST is
+			["object",
+				[["key", "foo"],
+				["array",
+					["some",
+						["NOT", ["number", ["=", 5]]]]]]
```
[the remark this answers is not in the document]
ekinanp: Ah thanks, bad copy-paste

api/rql/internal/primary/meta/valuePredicate.go
```diff
@@ -0,0 +1,98 @@
...
+Given a value predicate AST, assume that all negatable value predicate
+expressions are in reduced form (meaning all "NOT" operators are of the
+form ["NOT", ValuePredicate]). Consider a leaf in this tree (a ValuePredicate
+node or a "NOT" node). Then the leaf's satisfying value schema (SVS) is the
+schema of all satisfying values of the predicate generated by the leaf-AST [the
+AST consisting of all nodes up to this leaf ignoring any intermediate "AND"/"OR"
+operators].
+
+As an example, consider the following AST:
+	["object",
```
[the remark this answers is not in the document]
ekinanp: It's an internal package so won't be documented. This is meant for maintainers.

ekinanp:
Updated.

MikaelSmith (approved):

PR #745, opened 2020-03-02
Specify canonical user for gcp instance exec

### LINKED ISSUE #725: gcp compute exec uses user from ssh config file
My SSH config file has a `Host *` line that sets the default user to root. GCP compute exec uses the user parsed from the GCE SSH files (see https://github.com/puppetlabs/wash/blob/master/plugin/gcp/computeInst.go#L113) and sets that user as the FallbackUser. Thus, the SSH transport will use the `root` user when SSH'ing into the instance (b/c it reads from ~/.ssh/config via https://github.com/puppetlabs/wash/blob/master/transport/ssh.go#L66-L69) instead of the user parsed from my GCE SSH files. Thus, Exec fails because I'm ssh'ing as the wrong user.

I'm filing this issue because exec works fine when I do `gcloud ssh`, meaning GCP probably doesn't read the ~/.ssh/config file for extra info. That means we probably shouldn't either.

WORKAROUND: Just make sure you don't have any `Host` matchers that can conflict with a GCP compute instance's host name. So for me that means removing the `Host *` matcher or changing it to something else like `foo`.

MikaelSmith:
Previously we supplied the user registered with GCP metadata as a `FallbackUser`, which allowed a user's SSH config to change the user. However this behavior is inconsistent with `gcloud ssh` and resulted in default user entries (`Host *`) in SSH config breaking this functionality.

Switch to using the user from GCP metadata as the canonical user for login. If we later need to support changing that user, it should be a new feature of the `wexec` command/API.

Fixes #725.

### DIFF
--- a/plugin/gcp/computeInst.go
+++ b/plugin/gcp/computeInst.go
@@ -134,7 +134,7 @@ func (c *computeInstance) Exec(ctx context.Context, cmd string, args []string,
 
 	identity := transport.Identity{
 		Host:         hostname,
-		FallbackUser: user,
+		User:         user,
 		IdentityFile: conf.privateKey,
 		KnownHosts:   conf.knownHosts,
 		HostKeyAlias: hostKeyAlias(c.instance),


ekinanp (approved):
User will always be non-empty right?

MikaelSmith:
If it's empty then it will behave as before. And I think it will always be non-empty.

PR #748, opened 2020-03-05
Improve reporting on non-zero exit for fs List

### LINKED ISSUE #742: Failure in volume.FS List is ignored
If `find` in `volume.FS`'s `List` fails, the error is logged but the `ls` call still succeeds (exits 0) and reports the directory as empty. It should fail and return the error.

MikaelSmith:
Some non-zero exits are normal with `find` (see examples in code). The messages that caused this can appear on stdout or stderr depending on whether a tty is used or if it's a `stat` executed by `find`. Ignore normal failures, but if we get an unknown one fail `VolumeList` and return the error.

Fixes #742.

### DIFF
--- a/volume/fs.go
+++ b/volume/fs.go
@@ -50,12 +50,12 @@ func (d *FS) List(ctx context.Context) ([]plugin.Entry, error) {
 
 type nonZeroError struct {
 	cmdline  []string
-	output   string
+	stderr   string
 	exitcode int
 }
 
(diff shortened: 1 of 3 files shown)


ekinanp (approved):

MikaelSmith:
Changed it slightly because sometimes `stat:` and `find:` messages appear on stderr, and that caused the docker tests to fail.

ekinanp (approved):

PR #751, opened 2020-03-06
Implement the find endpoint

ekinanp:


api/rql/find.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/puppetlabs/wash/plugin"
+)
+
+// Find returns all children of the start entry that satisfy the given query
+// (including the start entry itself). Note that all returned entries' paths
+// will start from "", which represents the start path. For example, if "childOne"
+// and "childTwo" are the cnames of the start entry's children, then their "Path"
+// fields will be set to "childOne" and "childTwo" (where the start entry's path of
+// "" is automatically prefixed).
+func Find(ctx context.Context, start plugin.Entry, query Query, options Options) ([]Entry, error) {
```
ekinanp: Thinking the RQL should exclude the start path given that `Path` and `Kind` don't match it. Then the spec is "this returns all descendants of path that satisfy the given query"

api/rql/prune.go
```diff
@@ -0,0 +1,135 @@
+package rql
```
ekinanp: See https://github.com/puppetlabs/wash/tree/master/cmd/internal/find/types, specifically the `entrySchema` files as much of this was copied from there. Only addition was modifying the schema paths to match what Kind primary expects.

api/find.go
```diff
@@ -0,0 +1,95 @@
...
+	maxDepth, hasMaxDepth, errResp := getIntParam(r.URL, "maxdepth")
+	if errResp != nil {
+		return errResp
+	}
+	fullMeta, errResp := getBoolParam(r.URL, "fullmeta")
+	if errResp != nil {
+		return errResp
+	}
+	var rawQuery interface{}
+	if err := json.NewDecoder(r.Body).Decode(&rawQuery); err != nil {
```
ekinanp: Hm, starting to think the Boolean primaries should be added back. Then empty query can just set the rawQuery to `true`.

api/find.go
```diff
@@ -0,0 +1,95 @@
...
+//       500: errorResp
+var findHandler = handler{fn: func(w http.ResponseWriter, r *http.Request) *errorResponse {
+	ctx := r.Context()
+	entry, path, errResp := getEntryFromRequest(r)
+	if errResp != nil {
+		return errResp
+	}
+
+	/*
+		// TODO: Include this?
```
ekinanp: This depends on whether we want to include the start path or not. If no, then we should probably put this check here.

api/rql/find.go
ekinanp: Also one annoying this is we can't enforce order of returned values. We do know that returned values will always be parent then children, but there's no ordering in which parent is traversed. E.g., given something like `.`, `./foo`, `./foo/1`, `./bar`, we can always assume that `./foo` will come before `./foo/1` in the return results, but we cannot ordering of `./bar` and `./foo`.

This is b/c of `plugin.EntryMap`, which stores things in a `map` object and `map` has no order. Maybe this is OK?

ekinanp:
Docs will probably be done separately. Should they go on Website as something like `Documentation::Resource Query Language` ?

api/rql/find.go
MikaelSmith: We usually sort them before traversing when it matters.

ekinanp: Ok, updated to sort properly.

What about excluding the start path?

api/find.go
ekinanp: Added it back.

api/find.go
ekinanp: Removed it. RQL already handles this for us.

api/rql/find.go
MikaelSmith: Not sure I understand the impact yet.

MikaelSmith: Yeah, I'd exclude start path. It feels clearer, and consistent with what you'd expect (how `find` currently works).

api/types/rqlQuery.go
```diff
@@ -0,0 +1,3 @@
+package apitypes
```
ekinanp: Oh I guess this file isn't being used anywhere. Let me remove it.

api/rql/find.go
MikaelSmith: Oh, I was wrong about how `find` works. But it would be difficult to implement the way `find` works in the API, so it feels better to exclude start path in the API and if we want our `find` command to work consistent with the BSD implementation then we can handle that in the command.

ekinanp: Updated to exclude the start path.

api/types/rqlQuery.go
ekinanp: Done.

api/find.go
```diff
@@ -0,0 +1,97 @@
...
+//       404: errorResp
+//       500: errorResp
+var findHandler = handler{fn: func(w http.ResponseWriter, r *http.Request) *errorResponse {
+	ctx := r.Context()
+	entry, path, errResp := getEntryFromRequest(r)
+	if errResp != nil {
+		return errResp
+	}
+
+	if !plugin.ListAction().IsSupportedOn(entry) {
```
ekinanp: Yeah, had to add this back. Now that we're excluding start path, having `find` return an empty array for a file was kind of weird.

api/find.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/puppetlabs/wash/activity"
+	"github.com/puppetlabs/wash/api/rql"
+	"github.com/puppetlabs/wash/api/rql/ast"
+	apitypes "github.com/puppetlabs/wash/api/types"
+	"github.com/puppetlabs/wash/plugin"
+)
+
+// swagger:route GET /fs/find find listEntries
+//
+// Recursively descends the given path returning it and its children.
```
MikaelSmith: This is where in-code API docs would go. But I think how to use this API should be its own section on our documentation site.

api/find.go
MikaelSmith: This might make it challenging to implement something like `find foo -meta .key value`. I feel like I'm still working to understand the trade-offs.

api/find.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	var rawQuery interface{}
+	if err := json.NewDecoder(r.Body).Decode(&rawQuery); err != nil {
+		if err != io.EOF {
+			return badRequestResponse(fmt.Sprintf("could not decode the RQL query: %v", err))
+		}
+		rawQuery = true
+	}
+	query := ast.Query()
+	if err := query.Unmarshal(rawQuery); err != nil {
```
MikaelSmith: That doesn't feel like the normal definition of Unmarshal. More like a deep type cast.

MikaelSmith: If this were unmarshaling json, you could still do the previous test with a `json.RawMessage`.

api/find.go
ekinanp: For something like that you can just use the parent and restrict the cname to something specific.

api/find.go
ekinanp: Yeah, I'll work on that after this.

api/find.go
ekinanp: Not sure what you mean.

MikaelSmith: Everything around ASTNode's `Unmarshal` seems kind of complicated. Why first unmarshal to an `interface{}` then pass that to a function to handle it rather than unmarshal directly in that function? If we need to do partial unmarshaling, that's what `json.RawMessage` is meant to handle.

plugin/entrySchema.go
```diff
@@ -31,6 +31,9 @@ func schema(e Entry) (*EntrySchema, error) {
 		if err != nil {
 			return nil, err
 		}
+		if graph == nil {
```
MikaelSmith: What's this handling?

MikaelSmith (approved):
Looks ok, just have a few questions.

plugin/entrySchema.go
ekinanp: Mocking out schema graphs in the tests. Needed to add that for the `walker` tests.

api/find.go
ekinanp: I like thinking of the input as Go types than some standardized thing (so map[string]interface{} instead of as a JSON object). That makes it easer to support YAML ASTs or some other structured data representation.

I'm open to changing it, just not sure what a better representation would look like.

Also, there isn't any partial unmarshalling going on here so don't think `json.RawMessage` is necessary.

PR #752, opened 2020-03-08
Refactor tests to be more representative of the real world

ekinanp:
Real-world RQL use involves unmarshaling the AST from JSON then calling
the appropriate Eval methods. The values passed into value predicates
are also unmarshaled JSON values. The unit tests should simulate these
real-world scenarios. Thus, this commit modifies the helpers in
asttest.Suite to ensure that JSON unmarshaled Go types are passed as
input both for the ast and for value predicates.

### DIFF
--- a/api/rql/ast/asttest/suite.go
+++ b/api/rql/ast/asttest/suite.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"github.com/puppetlabs/wash/api/rql"
-	"github.com/puppetlabs/wash/api/rql/internal"
 	"github.com/puppetlabs/wash/api/rql/internal/errz"
 	"github.com/puppetlabs/wash/plugin"
 	"github.com/shopspring/decimal"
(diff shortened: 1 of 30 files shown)


api/rql/ast/asttest/suite.go
```diff
@@ -6,17 +6,23 @@ import (
...
 	"github.com/stretchr/testify/suite"
 )
 
-// Suite represents a type that tests RQL AST nodes
+// Suite represents a type that tests RQL AST nodes. N represents the AST node
+// constructor that returns an empty AST node object. It can be overridden by
+// each test. DefaultN represents the default node constructor that should be
+// set when the suite class is created.
+//
+// TODO: Add some more comments once a tested version of this is working.
```
MikaelSmith: I'm not sure what more needs to be done to address this TODO.

MikaelSmith (changes requested):

api/rql/ast/asttest/suite.go
ekinanp: Oh, yeah that's no longer needed. Thanks, updated.

MikaelSmith (approved):

PR #753, opened 2020-03-09
Merge rql branch into master

ekinanp:


plugin/entrySchema.go
```diff
@@ -31,6 +31,9 @@ func schema(e Entry) (*EntrySchema, error) {
 		if err != nil {
 			return nil, err
 		}
+		if graph == nil {
```
MikaelSmith: When addressing some of the TODOs, we should add a comment to this that it's primarily to simplify testing.

MikaelSmith (approved):

PR #754, opened 2020-03-09
Add initial RQL docs to the website

ekinanp:
This will be fleshed out with information about entry schema
optimization.

### DIFF
--- a/docs/_docs/rql.md
+++ b/docs/_docs/rql.md
@@ -1,0 +1,470 @@
+---
+title: Resource Query Language
+---
+
+* [Background](#background)
+* [AST Grammar](#ast-grammar)
+* [Primaries](#primaries)
+  * [action](#action)
+  * [boolean](#boolean)
+  * [name](#name)
+  * [cname](#cname)
+  * [path](#path)
+  * [kind](#kind)
+  * [atime](#atime)
+  * [crtime](#crtime)
+  * [ctime](#ctime)
+  * [mtime](#mtime)
+  * [size](#size)
+  * [meta](#meta)
+* [Detailed Meta Primary Overview](#detailed-meta-primary-overview)
+
+## Background
+
+The resource query language (RQL) lets you query Wash resources. You can submit your queries to the API's `find` endpoint. Here's an example query
+
+```
+Eniss-MacBook-Pro:wash enisinan$ curl -X POST --unix-socket /tmp/WASH_SOCKET --header "Content-Type: application/json" --data '["kind", ["glob", "*ec2*instance"]]' 'http://localhost:/fs/find?path=/tmp/WASH_MOUNT/aws/wash' 2>/dev/null | jq
+[
+  {
+    "type_id": "aws::github.com/puppetlabs/wash/plugin/aws/ec2Instance",
+    "path": "/tmp/WASH_MOUNT/aws/wash/resources/ec2/instances/i-04621c13583930e6c",
+...
+```
+
+> Note that I started my own Wash server instance with `WASH_SOCKET="/tmp/WASH_SOCKET" ./wash server /tmp/WASH_MOUNT`
+
+This query returns all entries under the `aws/wash` entry whose `kind` matches the glob `*ec2*instance`. Informally, this query returns all AWS EC2 instances under the `wash` profile.
+
+You can view the [API docs]({{'/docs/api' | relative_url}}) for more details on the `find` endpoint, including its query parameters (not to be confused with an RQL query, which is specified in the request body).
+
+## AST Grammar
+
+This section documents the RQL's AST grammar. For convenience, let `PE <PredicateType>` denote the following grammar (where `PE` => `PredicateExpression`).
+
+```
+PE <PredicateType> :=
+  [BinaryOp, PE <PredicateType>, PE <PredicateType>] |
+  <PredicateType>
+
+BinaryOp := "AND" | "OR"
+```
+
+where `<PredicateType>` has its own AST grammar. Informally, `PE <PredicateType>` translates to `a predicate expression of <PredicateType>s`. For example, `PE Primary` translates to `a predicate expression of primaries`.
+
+Similarly, let `NPE <PredicateType>` denote the following grammar (where `NPE` => `NegatablePredicateExpression`):
+
+```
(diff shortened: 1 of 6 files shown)


docs/_docs/rql.md
```diff
@@ -0,0 +1,470 @@
...
+  * [size](#size)
+  * [meta](#meta)
+* [Detailed Meta Primary Overview](#detailed-meta-primary-overview)
+
+## Background
+
+The resource query language (RQL) lets you query Wash resources. You can submit your queries to the API's `find` endpoint. Here's an example query
+
+```
+Eniss-MacBook-Pro:wash enisinan$ curl -X POST --unix-socket /tmp/WASH_SOCKET --header "Content-Type: application/json" --data '["kind", ["glob", "*ec2*instance"]]' 'http://localhost:/fs/find?path=/tmp/WASH_MOUNT/aws/wash' 2>/dev/null | jq
```
MikaelSmith: I'd simplify the prompt to
```suggestion
$ curl -X POST --unix-socket /tmp/WASH_SOCKET --header "Content-Type: application/json" --data '["kind", ["glob", "*ec2*instance"]]' 'http://localhost/fs/find?path=/tmp/WASH_MOUNT/aws/wash' 2>/dev/null | jq
```

Also, what were you getting on stderr?

docs/_docs/rql.md
```diff
@@ -0,0 +1,470 @@
...
+BinaryOp := "AND" | "OR"
+```
+
+where `<PredicateType>` has its own AST grammar. Informally, `PE <PredicateType>` translates to `a predicate expression of <PredicateType>s`. For example, `PE Primary` translates to `a predicate expression of primaries`.
+
+Similarly, let `NPE <PredicateType>` denote the following grammar (where `NPE` => `NegatablePredicateExpression`):
+
+```
+NPE <PredicateType> :=
+  [UnaryOp,  NPE <PredicateType>]                       |
```
MikaelSmith: Since it's already specialized with "negatable", maybe drop the generic Unary and just inline `'NOT'`.

docs/_docs/rql.md
```diff
@@ -0,0 +1,470 @@
...
+  Path    |
+  Kind    |
+  Atime   |
+  Crtime  |
+  Ctime   |
+  Mtime   |
+  Size    |
+  Meta
+
+Action  := [“action”, NPE ActionPredicate]
```
MikaelSmith: I'd lean towards flattening this. I think the mapping is pretty obvious (e.g. `Name` -> `["name", ...]`, and no one wants to read a long grammar definition.

docs/_docs/rql.md
```diff
@@ -0,0 +1,470 @@
...
+  [“=”,      <string>]
+
+TimePredicate := [ComparisonOp, TimeValue]
+```
+
+This grammar tells us several things:
+
+* An RQL query is a predicate expression of primaries. A primary is a predicate on a specific field of an entry. Examples of primaries include `name`, `cname`, and `ctime`. See the [primaries](#primaries) section for all of the available primaries.
+
+* Primaries take predicates, specifically predicate expressions.
```
MikaelSmith: I think I'd put a few small examples here before getting into specific primaries.

MikaelSmith: You want to take what they just read and map it into some concrete examples that help to show the simple patterns it represents.

docs/_docs/rql.md
```diff
@@ -0,0 +1,470 @@
...
+        ["AND",
+          ["object",
+            [["key", "key"),
+              ["string", ["=", "termination_date"]]]]],
+          ["object",
+            [["key", "value"),
+              ["time", ["<", "2017-08-07T13:55:25.680464+00:00"]]]]]]]]]]]
+```
+
+Returns true if `m['tags']` has at least one object o s.t. `o['key'] == termination_date` AND `o['value'] < 8/07/2017 ...` (i.e. `o['value']` has expired). In the real world, this example could be combined with the `kind` primary to filter out all EC2 instances whose `termination_date` tag expired. The full request to `find` would look something like
```
MikaelSmith:
```suggestion
Returns true if `m['tags']` has at least one object `o` s.t. `o['key'] == termination_date` AND `o['value'] < 8/07/2017 ...` (i.e. `o['value']` has expired). In the real world, this example could be combined with the `kind` primary to filter out all EC2 instances whose `termination_date` tag expired. The full request to `find` would look something like
```

MikaelSmith (changes requested):

docs/_docs/rql.md
ekinanp: Ok, can do that.

stderr I was getting progress status.

docs/_docs/rql.md
ekinanp: Ah, I see what you mean. Yeah, I'll do that.

ekinanp:
Updated. This should now be complete.

docs/_docs/rql.md
```diff
@@ -0,0 +1,658 @@
...
+                    └── fs
+                        ├── [dir]
+                        │   ├── [dir]
+                        │   └── [file]
+                        └── [file]
+```
+
+We will refer to this hierarchy when talking about the AWS plugin. EC2 instances are the only execable entries in this hierarchy. 
+
+The RQL will optimize its search to only recurse into entries that are execable or have an execable descendant. For the AWS plugin, this means that the RQL will _not_ recurse into S3 buckets (the node labeled `[bucket]`) or an EC2 instance's filesystem (the node labeled `fs`) -- those entries do not have any execable descendants. It will, however, recurse into AWS profiles (`[profile]`) and the resources directory (`resources`) since those entries have execable descendants (EC2 instances [`[instance]`]).
```
MikaelSmith: Not sure what the double brackets in
```
EC2 instances [`[instance]`]
```
is supposed to imply. I think you could drop the outer brackets.

MikaelSmith (approved):

docs/_docs/rql.md
ekinanp: Eh, just wanted to make it explicit. I agree though they aren't that necessary. Updated.

MikaelSmith (approved):

PR #755, opened 2020-03-11
Add swagger docs for the find endpoint

ekinanp:


docs/_docs/api.json
```diff
@@ -1174,6 +1246,7 @@
         }
       }
     },
+    "findResponse": {},
```
MikaelSmith: It doesn't look like findResponse got documented. I would've expected it to match entryList.

MikaelSmith (changes requested):

docs/_docs/api.json
ekinanp: Huh, I guess `type findResponse entryList` didn't work. Thanks, updated.

MikaelSmith (approved):

PR #760, opened 2020-03-26
Replace HOME with tilde in shell prompt

### LINKED ISSUE #757: User home directory path doesn't use tilde
If I `cd ~`, I'd expect the prompt path to use `~` rather than the absolute path. I see
```
wash /Users/michaelsmith ❯ cd Documents
wash /Users/michaelsmith/Documents ❯
```
I'd expect to see
```
wash ~ ❯ cd Documents
wash ~/Documents ❯
```

MikaelSmith:
Replace use of `realpath` with string substitution. If not in a Wash directory tree, but we are in the HOME directory tree, replace `$HOME` with `~`.

Fixes #757.

### DIFF
--- a/cmd/internal/shell/core.go
+++ b/cmd/internal/shell/core.go
@@ -52,10 +52,20 @@ func preparePrompt(cyan, green, reset, assign string) string {
 	return `
 function prompter() {
 	local prompt_path
-	if [ -x "$(command -v realpath)" ]; then
-		prompt_path=$(realpath --relative-base=$W "$(pwd)")
-	else
-		prompt_path=$(basename "$(pwd)")
+
+	if [ -v W ]; then
+		# If the current directory is W, replace with '.'. Else if it's a subdir, show just subpath.
+		if [ "${prompt_path}" = "${W}" ]; then
+			prompt_path=.
+		else
+			prompt_path=${PWD/#$W\//}
+		fi
+	fi
+
+	if [ -v HOME ]; then
+		# If prompt_path is still an absolute path (because it hasn't been changed) and is a subdir
+		# of HOME, replace HOME with '~'.
+		prompt_path=${prompt_path/#$HOME/\~}
 	fi
 	` + assign + `="` + cyan + `wash ${prompt_path}` + green + ` ❯` + reset + ` "
 }


cmd/internal/shell/core.go
```diff
@@ -49,13 +49,23 @@ func Get() Shell {
...
+	// If prompt path is a subdirectory of $HOME, then replace with '~'.
 	return `
 function prompter() {
 	local prompt_path
-	if [ -x "$(command -v realpath)" ]; then
-		prompt_path=$(realpath --relative-base=$W "$(pwd)")
-	else
-		prompt_path=$(basename "$(pwd)")
+
+	if [ -v W ]; then
```
ekinanp: What does `-v` do?

cmd/internal/shell/core.go
```diff
@@ -49,13 +49,23 @@ func Get() Shell {
...
+		prompt_path=${PWD/#$W\//}
+		if [ "${prompt_path}" = "${W}" ]; then
+			prompt_path=.
+		fi
+	fi
+
+	if [ -v HOME ]; then
+		prompt_path=${prompt_path/#$HOME/\~}
```
ekinanp: When will it ever be a subdirectory of `HOME` and `W`? If never, then this would be clearer if it were an `elseif`

cmd/internal/shell/core.go
```diff
@@ -49,13 +49,23 @@ func Get() Shell {
...
 	return `
 function prompter() {
 	local prompt_path
-	if [ -x "$(command -v realpath)" ]; then
-		prompt_path=$(realpath --relative-base=$W "$(pwd)")
-	else
-		prompt_path=$(basename "$(pwd)")
+
+	if [ -v W ]; then
+		prompt_path=${PWD/#$W\//}
```
ekinanp: So if PWD were `$W/`, would prompt_path be `''` ? Is that an issue? If no, comment explaining that would help.

cmd/internal/shell/core.go
MikaelSmith: Tests that the variable is set.

cmd/internal/shell/core.go
MikaelSmith: Yeah, I may be trying to make it shorter at the cost of readability.

cmd/internal/shell/core.go
MikaelSmith: PWD should never end with a `/`.

cmd/internal/shell/core.go
MikaelSmith: It's a little challenging to make the right logic. That's why I put the comment at the top. I'm relying on string substitution to strip the leading `/` so later substitution won't match.

cmd/internal/shell/core.go
ekinanp: Ok, thanks

cmd/internal/shell/core.go
ekinanp: I think it might be simpler if the comment were written in terms of prompt_path (instead of PWD _and_ prompt_path). Something like

```
// Here's how to calculate prompt_path.
//
//. prompt_path = PWD
//  If prompt_path is subdir of W, then
//     prompt_path == <part_after_W>
//.    if prompt_path == <empty> then prompt_path = "."
//. else if prompt_path is subdir of HOME, then
//.   prompt_path = ~/<part_after_HOME>
//
// return prompt_path
```

ekinanp (approved):
Had a [comment](https://github.com/puppetlabs/wash/pull/760#discussion_r398945959) on possibly rewording comment + calculation, but it is not a blocker.

cmd/internal/shell/core.go
MikaelSmith: Ok, I can update that.

ekinanp: Shouldn't the implementation reflect that?

MikaelSmith: I'll work on a new description. What you wrote doesn't precisely match what's happening.

ekinanp (approved):

cmd/internal/shell/core.go
ekinanp: Ok.

PR #763, opened 2020-03-30
Make kubernetes containers a directory

### LINKED ISSUE #203: Make kubernetes containers a directory
Kubernetes containers should follow a similar format to docker containers:
```
log
metadata.json
fs/
```
If we can't access the log due to permissions issues, we can omit the `log` file.

The implementation should look similar to https://github.com/puppetlabs/wash/blob/master/plugin/docker/container.go, where we list 'log' and 'metadata' entries for the pod, implement metadata for the pod, and then add pod_metadata and pod_log implementations to implement the files using references back to the pod. Also use the volume helpers to expose the filesystem.

MikaelSmith:
Change the Kubernetes container entry to a directory mirroring the Docker container structure: log, metadata, filesystem.

Also fixes using a tty with the container's exec method (which was previously unused).

Resolves #203.

### DIFF
--- a/plugin/kubernetes/container-log.go
+++ b/plugin/kubernetes/container-log.go
@@ -1,0 +1,64 @@
+package kubernetes
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/puppetlabs/wash/activity"
+	"github.com/puppetlabs/wash/plugin"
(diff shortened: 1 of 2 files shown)


MikaelSmith:
This is going to have some overlap with refactoring in #762, but it's functionally independent.

ekinanp (approved):
LGTM

PR #771, opened 2020-04-14
Implement Writable for the Volume plugin

### LINKED ISSUE #675: Implement Writable for Volumes
We've added the ability to write object stores. Another obvious use case is to change files on volumes, such as modifying configuration in a VM or container.

MikaelSmith:
Volumes now support writing content to files. Resolves #675.

Depends on #773 to ensure writes are immediately visible.

### DIFF
--- a/plugin/docker/volume.go
+++ b/plugin/docker/volume.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 	"io"
 	"io/ioutil"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -75,10 +77,9 @@ func (v *volume) createContainer(ctx context.Context, cmd []string) (string, fun
 	// Use tty to avoid messing with the extra log formatting.
 	cfg := docontainer.Config{Image: "busybox", Cmd: cmd, Tty: true}
 	mounts := []mount.Mount{{
-		Type:     mount.TypeVolume,
-		Source:   v.Name(),
-		Target:   mountpoint,
-		ReadOnly: true,
+		Type:   mount.TypeVolume,
+		Source: v.Name(),
+		Target: mountpoint,
 	}}
 	hostcfg := docontainer.HostConfig{Mounts: mounts}
 	netcfg := network.NetworkingConfig{}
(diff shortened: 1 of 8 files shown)


volume/fs.go
```diff
@@ -202,6 +202,44 @@ func (d *FS) VolumeStream(ctx context.Context, path string) (io.ReadCloser, erro
 	return r, nil
 }
 
+// VolumeWrite satisfies the Interface required by Write to write content to a file.
+func (d *FS) VolumeWrite(ctx context.Context, path string, b []byte) error {
+	activity.Record(ctx, "Writing %v bytes to %v on %v", len(b), path, plugin.ID(d.executor))
```
ekinanp, line 207: Should we just log this at the `plugin.Write` level?

plugin/docker/volume.go
```diff
@@ -110,57 +110,82 @@ func (v *volume) createContainer(ctx context.Context, cmd []string) (string, err
...
-	if err != nil {
-		return nil, err
-	}
-	defer func() {
-		err := v.client.ContainerRemove(ctx, cid, types.ContainerRemoveOptions{})
+type containerDeleter = func()
+type doInContainerFn = func(containerDeleter) error
+
+// Executes the callback while handling start and cleanup of an existing container.
+func (v *volume) doInTemporaryContainer(ctx context.Context, cid string, fn doInContainerFn) error {
```
ekinanp, line 117: Could `createContainer`'s signature be extended to `(cid, deleteFunc, error)`, and make the post-condition be that it returns a _running_ container? If so, then I don't think we need `doInTemporaryContainer`.

volume/fs.go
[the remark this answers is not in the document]
MikaelSmith: It says pretty much the same thing in FUSE logging. This splits up path and executor id, but that's probably not important. So I think we could omit it.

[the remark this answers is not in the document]
ekinanp: Ok, that's fine.

plugin/docker/volume.go
MikaelSmith: Yeah I think so. At one point `doInTemporaryContainer` handled some of the `defer` calls, but that wasn't flexible enough.

MikaelSmith:
I think this is ready again.

plugin/docker/volume.go
```diff
@@ -202,56 +206,67 @@ func (v *volume) VolumeRead(ctx context.Context, path string) ([]byte, error) {
...
+
+func (v *volume) VolumeWrite(ctx context.Context, path string, b []byte, mode os.FileMode) error {
+	// Create a container that mounts a volume and waits. Use it to upload a file.
+	cid, cleanup, err := v.createContainer(ctx, []string{"sleep", "60"})
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+
+	// Create a tar of the file contents and upload it.
```
ekinanp, line 244: Why are we tar'ing and copying? Some clarification would help. Seems weird that this is different from the Kubernetes PVC implementation.

ekinanp: Other than the clarification, PR looks good.

MikaelSmith: Added. I suppose I could've done it with execs like everything else. That would avoid needing to pass in the permissions.

ekinanp: How hard would it be to do that? The inconsistency between this and Kube PVC is kind of weird.

ekinanp (approved):

plugin/docker/volume.go
MikaelSmith: I'll explore cleaning this up similar to Kubernetes.

I didn't add the same logic to identify an existing container because
- Docker scheduling is a lot faster, so it's not much overhead compared to starting a new process
- Docker doesn't have anything like ReadWriteOnce, so we can always get read/write access

Issue #769, opened 2020-04-13
Add ability to use exec on directories


sbrow:
## Use Case
When using the docker plugin, in order to run commands in a particular directory, I have to do this:

```bash
wexec docker/containers/container_name sh -c 'cd /var/ && ls'
``` 

When it would be, IMO, much more convenient and idiomatic to do this:

```bash
wexec docker/containers/container_name/fs/var ls
```

## Describe the Solution You Would Like
Add the "exec" action to directories in the docker fs folder, where exec is equivilant to `docker exec -it -w $directory_name $container_name $cmd`.

For example: 
```bash
docker exec -it -w /shared/httpd/ devilbox_php_1 pwd
```
would be equivilant to:
```bash
wexec docker/containers/devilbox_php_1/fs/shared/httpd pwd
```

Thanks! I really love the project.

MikaelSmith:
That's a great idea, I like it. I'm going to mull it over for a bit, but I don't see any way that contradicts how exec might evolve in the future.

One thing it does break is a habit I have of equating executable things with directly accessible resources, but that's a kludge I use to limit queries anyway. I think a better solution would be to have a separate property (maybe an `attribute`) that reflects whether the entry is directly addressable or requires indirection to access.

sbrow:
Thanks for the blazing fast response! 😄 If you need any help implementing this on the docker plugin,  just point me in the right direction and I can start a PR.

MikaelSmith:
I definitely wouldn't get to it immediately, so that would be great.

I'd suggest starting by taking a look at https://github.com/puppetlabs/wash/blob/master/CORE_PLUGIN_DEVELOPMENT.md. The Docker plugin includes its view of the filesystem at https://github.com/puppetlabs/wash/blob/master/plugin/docker/container.go#L92-L94; all plugins actually use a shared implementation to view filesystems: https://pkg.go.dev/github.com/puppetlabs/wash/volume?tab=doc via NewFS.

The `volume.FS` type implements `volume.Interface`. Calling `List` on `volume.FS` lists the "root" of the filesystem as an array of `volume.dir` types. I think these are what we want to make executable; when you call `wexec docker/containers/container_name/fs/var ls`, you're running `ls` in the `/var` directory; the same wouldn't make sense for a file (aka `volume.File`).

[...]

PR #772, opened 2020-04-15
Implement exec on Directories

sbrow:
As discussed in issue #769. This is still a work in progress.

# Todos:
- [ ] Write tests
- [ ] Implement in plugins other than docker.
- [ ] Add option to enable in `volume.NewFS`?
- [ ] Update `entrySchema` on `execDir`.


MikaelSmith:
That looks pretty straightforward, although it looks like some test mocking needs to be tweaked.

PR #773, opened 2020-04-15
Clear parent List cache for prefetched

### LINKED ISSUE #770: Clearing cache on a volume doesn't clear the source
Setup is to run a Docker container and use Wash to navigate to some point in its filesystem. Depending on where you are, running `wclear .` doesn't actually clear cached data. `whistory` output is the clearest way to demonstrate that.

```
wash docker/containers/strange_fermat/fs/usr/local > wclear .
wash docker/containers/strange_fermat/fs/usr/local > whistory 1
Apr 14 16:24:36.627 API: DELETE /cache?path=%2FUsers%2Fmichaelsmith%2FLibrary%2FCaches%2Fwash%2Fmnt167843624%2Fdocker%2Fcontainers%2Fstrange_fermat%2Ffs%2Fusr%2Flocal
Apr 14 16:24:36.627 API: Cache DELETE /docker/containers/strange_fermat/fs/usr/local []
Apr 14 16:24:36.627 API: DELETE /cache?path=%2FUsers%2Fmichaelsmith%2FLibrary%2FCaches%2Fwash%2Fmnt167843624%2Fdocker%2Fcontainers%2Fstrange_fermat%2Ffs%2Fusr%2Flocal complete
```
The 2nd line of output shows nothing was deleted from the cache.

```
wash docker/containers/strange_fermat/fs/usr/local > wclear ..
wash docker/containers/strange_fermat/fs/usr/local > whistory 2
Apr 14 16:24:48.079 API: DELETE /cache?path=%2FUsers%2Fmichaelsmith%2FLibrary%2FCaches%2Fwash%2Fmnt167843624%2Fdocker%2Fcontainers%2Fstrange_fermat%2Ffs%2Fusr
Apr 14 16:24:48.079 API: Cache DELETE /docker/containers/strange_fermat/fs/usr [List::/docker/containers/strange_fermat/fs]
Apr 14 16:24:48.079 API: DELETE /cache?path=%2FUsers%2Fmichaelsmith%2FLibrary%2FCaches%2Fwash%2Fmnt167843624%2Fdocker%2Fcontainers%2Fstrange_fermat%2Ffs%2Fusr complete
```
Running the same thing against the parent directory, the `fs` entry is deleted from the cache.

This appears to happen because `wclear` clears any cache items related to the entry and its parent. However the `fs/usr/local` entry itself is derived from the `fs` entry and there are no actual cache entries that contain it or its parent.

MikaelSmith:
If an entry is Prefetched, it's parent may not have a cached List entry but it may have originated from an earlier ancestor. If there's an immediate parent that listed the entry still clear it, but if not search for an ancestor that prefetched child entries and clear that instead.

Fixes #770.

### DIFF
--- a/plugin/cache.go
+++ b/plugin/cache.go
@@ -93,17 +93,53 @@ func allOpKeysIncludingChildrenRegex(path string) *regexp.Regexp {
 //
 // TODO: If path == "/", we could optimize this by calling cache.Flush(). Not important
 // right now, but may be worth considering in the future.
-func ClearCacheFor(path string, clearParentList bool) []string {
+func ClearCacheFor(path string, clearAncestorList bool) []string {
 	rx := allOpKeysIncludingChildrenRegex(path)
 	deleted := cache.Delete(rx)
 
-	if clearParentList {
-		parentID, _ := splitID(path)
-		listOpName := defaultOpCodeToNameMap[ListOp]
-		deleted = append(deleted, cache.Delete(opKeyRegex(listOpName, parentID))...)
+	if clearAncestorList {
+		// If we can find the source ancestor, clear it and everything below it from cache.
+		if sourceAncestorID := getSourceAncestorPathFromCache(path); sourceAncestorID != "" {
+			listOpName := defaultOpCodeToNameMap[ListOp]
+			deleted = append(deleted, cache.Delete(opKeyRegex(listOpName, sourceAncestorID))...)
+		}
 	}
 
 	return deleted
+}
+
+// Get the path for the ancestor that prefetched an entry at path. If none are found in the cache,
+// returns an empty string. This may be overly aggressive in some cases where it finds an ancestor
+// but it's not the immediate source ancestor; this seems like an acceptable compromise to make
+// sure it works with incomplete information.
+func getSourceAncestorPathFromCache(path string) string {
+	// To determine the source, we need to find the first parent where it's child was not prefetched.
+	// This can be challenging because some items may not be in the cache.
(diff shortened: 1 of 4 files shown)


plugin/cache.go
```diff
@@ -98,14 +98,43 @@ func ClearCacheFor(path string, clearParentList bool) []string {
 	deleted := cache.Delete(rx)
 
 	if clearParentList {
+		// If an immediate parent exists, clear that. Otherwise search for the first available ancestor
```
ekinanp, line 101: If we change `clearParentList` => `clearAncestorList`, then could this be simplified to
```
ancestor = parent
if entry.Prefetched() {
  ancestor = getAncestor...
}
// Clearing logic here
```

MikaelSmith: We don't have an entry, and I want to avoid getting one because that requires doing a List call that might not be cached.

ekinanp: Ah I gotcha, I missed the fact that we didn't have an entry. But couldn't we still capture the basic idea in the current code? It would just be:

```
sourceAncestorID := getSourceAncestorPathFromCache(path)
// Clear the sourceAncestor's list op from the cache
...
```

(I'm using sourceAncestor because that's how we word things in other places, like the[ fuse code](https://github.com/puppetlabs/wash/blob/master/fuse/core.go#L119))

Also, I do think `clearParentList` => `clearSourceAncestorList` (since the issue shows that we may not clear parent's list for _all_ entries).

MikaelSmith: I'll try. I was having trouble making it generic.

ekinanp (approved):

PR #775, opened 2020-04-17
Fix panic when waiting for pod times out

### LINKED ISSUE #774: Error running container to inspect K8s pvc results in panic
```
2020/04/17 11:14:14 http: panic serving : interface conversion: interface {} is nil, not []uint8
goroutine 1269 [running]:
net/http.(*conn).serve.func1(0xc013d6fa40)
	/usr/local/Cellar/go/1.14.2_1/libexec/src/net/http/server.go:1772 +0x139
panic(0x5f94e80, 0xc0164d9ec0)
	/usr/local/Cellar/go/1.14.2_1/libexec/src/runtime/panic.go:975 +0x3e3
github.com/puppetlabs/wash/plugin/kubernetes.(*pvc).exec(0xc013b1a5a0, 0x68abbe0, 0xc0160dcf00, 0xc013ae7020, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/michaelsmith/puppetlabs/wash/plugin/kubernetes/pvc.go:179 +0x11c
github.com/puppetlabs/wash/plugin/kubernetes.(*pvc).VolumeList(0xc013b1a5a0, 0x68abbe0, 0xc0160dcf00, 0x0, 0x0, 0x37e11d600, 0x37e11d600, 0x0)
	/Users/michaelsmith/puppetlabs/wash/plugin/kubernetes/pvc.go:186 +0xd8
...
```

MikaelSmith:
Always check for an error before trying to cast the result, as casting a nil causes a panic. Also ensure the pod is deleted if there's an error waiting for it to be available. `kubectl get events` may help in identifying why the pod didn't start.

Fixes #774.

### DIFF
--- a/plugin/kubernetes/pvc.go
+++ b/plugin/kubernetes/pvc.go
@@ -136,13 +136,15 @@ func (v *pvc) inContainer(ctx context.Context, fn containerCb) (interface{}, err
 		if err != nil {
 			return nil, err
 		}
+		cleanup = func() {
+			// Use a background context to ensure deletion happens even if context was cancelled.
+			activity.Record(ctx, "Deleted temporary pod %v: %v", tempPod, tempPod.delete(context.Background()))
+		}
 		if err := tempPod.waitOnCreation(ctx); err != nil {
+			cleanup()
 			return nil, err
 		}
 		execContainer.pod = tempPod.pod
-		cleanup = func() {
-			activity.Record(ctx, "Deleted temporary pod %v: %v", tempPod, tempPod.delete(ctx))
-		}
 	} else {
 		mount := v.getMountInfo(mountingPod, volumeName)
 		execContainer.pod = mount.pod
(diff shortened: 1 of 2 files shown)


plugin/kubernetes/pvc.go
```diff
@@ -135,13 +135,15 @@ func (v *pvc) inContainer(ctx context.Context, fn containerCb) (interface{}, err
 		if err != nil {
 			return nil, err
 		}
+		cleanup = func() {
+			// Use a background context to ensure deletion happens even if context was cancelled.
+			activity.Record(ctx, "Deleted temporary pod %v: %v", tempPod, tempPod.delete(context.Background()))
```
ekinanp, line 140: Why background? Why not a timeout context?

MikaelSmith: Hmm. I guess. I'd hope the Delete implementation would have a somewhat bounded runtime, especially if it also supports context cancellation.

MikaelSmith: Same comment would also apply to https://github.com/puppetlabs/wash/blob/master/plugin/docker/volume.go#L122-L129.

ekinanp: Ah yeah, I missed that one. I guess the Q. here is "Does this code introduce a theoretically-possible edge-case of a dangling go-routine?" If so, may as well fix it since adding a timeout context is easy. If not, then no need.

Either way not a blocker. Rest of the PR looks good to me.

ekinanp (approved):

plugin/kubernetes/pvc.go
MikaelSmith: The timeout context is no guarantee either, it'll only halt the goroutine if the code that's blocking also checks the context for cancelation.

ekinanp (approved):

PR #778, opened 2020-06-17
Fix s3 downloads

### LINKED ISSUE #777: S3 objects larger than 4KB error when reading
Trying to read an object in S3 with `cat` results in
```
cat: README.md: Input/output error
```
Nothing relevant appears in `whistory` for that command either.

MikaelSmith:
Returning errors for FUSE operations get translated into system error codes and specific details are dropped. Ensure we log errors before joining them. Make them warnings so the cause of failure is clear to users.

The S3 bytes range is inclusive, so a request for byte range 0-8 will return 9 bytes. This was not obvious. Update so that a request for N bytes only returns N bytes.

Fixes #777.

### DIFF
--- a/plugin/aws/s3Object.go
+++ b/plugin/aws/s3Object.go
@@ -73,10 +73,16 @@ func (o *s3Object) Metadata(ctx context.Context) (plugin.JSONObject, error) {
 }
 
 func (o *s3Object) Read(ctx context.Context, size int64, offset int64) ([]byte, error) {
+	// Because bytes request is inclusive, short-circuit requests for 0 bytes.
+	if size <= 0 || offset < 0 {
+		return []byte{}, nil
+	}
+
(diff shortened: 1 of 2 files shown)


plugin/aws/s3Object.go
```diff
@@ -73,10 +73,16 @@ func (o *s3Object) Metadata(ctx context.Context) (plugin.JSONObject, error) {
 }
 
 func (o *s3Object) Read(ctx context.Context, size int64, offset int64) ([]byte, error) {
+	// Because bytes request is inclusive, short-circuit requests for 0 bytes.
+	if size == 0 {
+		return []byte{}, nil
+	}
+
+	// Bytes range is inclusive, so for N bytes get byte range 0-(N-1).
```
MikaelSmith, line 81: This was surprising.

PR #782, opened 2020-07-15
Update default branch to main

MikaelSmith:
Switches away from `master`, which has an oppressive history. See https://tools.ietf.org/id/draft-knodel-terminology-00.html.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,9 @@ name: Lint and Test
 
 on: 
   push:
-    branches: [master]
+    branches: [main]
   pull_request:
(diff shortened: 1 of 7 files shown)


MikaelSmith:
For now I plan to keep master around and in-sync for `/docs`. Hopefully GitHub changes that soon so you can use the `/docs` directory of the default branch.

ekinanp (approved):

ekinanp:
Would be nice to state the reason behind this change, maybe in the commit message? Like why are we switching to a `main` branch instead of `master`?

ekinanp (approved):

PR #784, opened 2021-02-25
Load all available k8s auth plugins instead of GCP only

skymeyer:
When using `wash` with kubernetes clusters a warning is returned:

```
WARN loading context freedom failed: no Auth Provider found for name "oidc"
```

This PR loads all available auth plugins instead of the one for `gcp` only (https://github.com/kubernetes/client-go/blob/master/plugin/pkg/client/auth/plugins.go), but will eat up some additional dependencies doing so.


MikaelSmith (approved):
Thanks! This seems like a simple improvement.

MikaelSmith:
I'll try to get a release of it out later today.

